Issue #8, opened 2025-01-08
Expose fingerprint and metavars fields


mwillfox:
Expose the `fingerprint` and `metavars` fields in the result output (both JSON and SARIF). These fields are necessary to track unique vulnerabilities and the original algorithm to generate the `fingerprint` field was originally published under an open source license.

dimitris-m:
Completed in #16

PR #16, opened 2025-01-15
Expose fingerprint and metavars fields

Kirill89:
I used [this](https://github.com/semgrep/semgrep/commit/9f4f2d64fd706c248b67da31f51657e61f931528) commit as a reference to solve https://github.com/opengrep/opengrep/issues/8.
Metavars are back now.

Please note, it's a lot of changes in this PR, but most of them are just snapshot files. Please pay attention to `.ml`, `.mli`, and  `.py` files changes.


Kirill89:
> Hi @Kirill89, thanks for contributing to Opengrep.
> 
> How did you test your changes?

```bash
make core
make core-test
cd cli
PATH=<ABS_PATH>/opengrep/bin:$PATH make ci-test
```

Kirill89:
@dimitris-m I also verified manually using `./bin/semgrep scan -f my_rule.yml my_code_path --json`

dimitris-m:
I have confirmed that these tests pass, going to check the diff in a bit more detail, aiming to review this by tomorrow. 

@willem-delbare let's wait for some final checks I need to do, we should be able to merge this soon.

src/rpc/RPC.ml
```diff
@@ -19,7 +19,7 @@ let handle_call (caps : < Cap.exec ; Cap.tmp >) :
       let modified_file_count, fixed_lines = RPC_return.autofix dryrun edits in
       Ok (`RetApplyFixes { modified_file_count; fixed_lines })
   | `CallSarifFormat
-      ( ctx,
+      ( _,
         {
           hide_nudge;
           engine_label;
```
dimitris-m, line 25: One question is, if the context `ctx` is no longer needed in this call, why have the parameter? 

Seems ok to leave in in the sense that it does not break something by ignoring it, and from a quick look I see that it's _some_ work to remove the parameter completely from this call.

If we don't fix that now we should open an issue to address this later. I guess this is a good compromise.
On the other hand we plan to eventually drop the python part, so... another argument against doing this work now.

dimitris-m (approved):
Looks good, I'll merge this in. 

Left some comments but they are non-blockers right now. We should probably create an issue as mentioned in my review.

I ran the tests and I see the same failure(s) as before, modulo stuff that fails because now in `main` we removed tracing and we still have tests that depend on it. So there's no regression introduced here, as far as I can tell.

Thanks for your contribution to Opengrep!

PR #38, opened 2025-01-24
Fix installation instructions

dimitris-m:
- Go straight to the .whl until the PyPi issue is solved


dimitris-m:
Closing as we focus on self-container binaries for now.

PR #48, opened 2025-01-27
Making it so that the URLs dont get normalised

jesse-merhi:
Same as https://github.com/semgrep/semgrep/pull/10798 

Currently, there exists an issue in the Language Server that when a URL is used for configuration it is normalised and as such
// is normalised to / which results in errors.

As can be seen here: https://github.com/semgrep/semgrep-vscode/issues/181

### DIFF
--- a/libs/commons/Uri_.ml
+++ b/libs/commons/Uri_.ml
@@ -26,3 +26,8 @@
 let of_string_opt (str : string) : Uri.t option =
   let uri = Uri.of_string str in
   if Uri.equal uri Uri.empty then None else Some uri
+
+
+let url_regex = Pcre2_.regexp "^https?://"
+let is_url config_path = 
+  Pcre2_.pmatch_noerr ~rex:url_regex config_path
--- a/libs/commons/Uri_.mli
+++ b/libs/commons/Uri_.mli
@@ -2,3 +2,9 @@
  * empty uri in case of error, we return None here.
  *)
 val of_string_opt : string -> Uri.t option
+
+(* Checks if the string starts with 'http://' or 'https://'
+ * Returns true only for valid HTTP(S) URL prefixes.
+ * Note: This only validates the scheme prefix, not the full URL structure.
+ *)
+val is_url : string -> bool
--- a/src/osemgrep/language_server/server/Session.ml
+++ b/src/osemgrep/language_server/server/Session.ml
@@ -279,12 +279,15 @@ let fetch_rules session =
   in
   let home = !Semgrep_envvars.v.user_home_dir in
   let rules_source =
-    session.user_settings.configuration |> List_.map Fpath.v
-    |> List_.map Fpath.normalize
-    |> List_.map (fun f ->
-           let p = Fpath.rem_prefix (Fpath.v "~/") f in
-           Option.bind p (fun f -> Some (home // f)) |> Option.value ~default:f)
-    |> List_.map Fpath.to_string
+    session.user_settings.configuration
+    |> List_.map (fun config_path ->
+      if Uri_.is_url config_path then
+        config_path
+      else
+        let f = Fpath.v config_path |> Fpath.normalize in
+        let p = Fpath.rem_prefix (Fpath.v "~/") f in
+        Option.value ~default:f (Option.map (fun f -> home // f) p)
+        |> Fpath.to_string)
   in
   let rules_source =
     if rules_source = [] && ci_rules = None then (


dimitris-m (commented):
@jesse-merhi thanks for contributing to Opengrep!

Before I can merge your PR, please consider some points. 

1. If you want me to maintain these commits in a merge commit: 

- The first commit `fixing opengrep` is not very informative. Why not rename it to something specific like `check if path is a url before path normalisation` or something of that sort?
- I would squash the second commit into the first, since the first one does not compile on its own. Moreover, the second commit also does not have a very informative name. Too generic.

Alternatively, I can merge your PR with a squash commit and you don't need to do anything.

2. The use of `Str` is not really recommended, especially as we are already using OCaml 5 and we will refactor the whole setup to make use of it. It's not a good idea to let things creep in, but I'm happy to make the change myself at a later time (unless if you _really_ want to do it now! :))... So it would be good to at least add some kind of note above `is_url` in `Uri_.ml` or `Uri_.mli`, for example `TODO: Use Pcre2, this is not thread-safe`.

Once more, thanks for your contribution.

jesse-merhi:
@dimitris-m !!

Sorry 😅 let me squash them and add the extra comments. 

I have literally never written OCaml until I found this bug so I have no clue what is right and wrong at this point - you can tell because I didn't call the function correctly 😆

I would love further explanations regarding `Str` and I am happy to try to fix it too!

dimitris-m:
As a side-note, I'm a little surprised that Semgrep's CI checks did not catch the compilation error, or am I missing something?

dimitris-m:
> @dimitris-m !!
> 
> Sorry 😅 let me squash them and add the extra comments.
> 
> I have literally never written OCaml until I found this bug so I have no clue what is right and wrong at this point - you can tell because I didn't call the function correctly 😆
> 
> I would love further explanations regarding `Str` and I am happy to try to fix it too!

Ok, see file `libs/commons/tests/Unit_Pcre2_.ml`, and if you like to know how the wrapper is used, `libs/commons/Pcre2_.mli` for interface / `libs/commons/Pcre2_.ml` for implementation.

To play around go into the `lib/commons` directory and run `dune utop` to start the top-level. 
The you can play around with the `Pcre2_`, for example: 

```ocaml 
utop # let regex = Pcre2_.regexp "^(http|https)://";;
val regex : Pcre2_.t = {Pcre2_.pattern = "^(http|https)://"; regexp = <abstr>}
─( 00:44:04 )─< command 2 >───────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Pcre2_.pmatch ~rex:regex "aaaaaaaaaaaaaaaaa!";;
- : (bool, Pcre2_.error) result = Ok false
─( 00:48:42 )─< command 3 >───────────────────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Pcre2_.pmatch ~rex:regex "http://google.com";;
- : (bool, Pcre2_.error) result = Ok true
```

jesse-merhi:
Yeah I mean I cant even get semgrep to compile locally to be honest... the documentation is just a bit vague and seems to be out of date.

dimitris-m:
> Yeah I mean I cant even get semgrep to compile locally to be honest... the documentation is just a bit vague and seems to be out of date.
> 
> Hopefully this commit is better. Let me know if it needs updating still :)

We should try to write some more beginner (in the sense of OCaml) friendly instructions, it's not easy...

libs/commons/Uri_.ml
```diff
@@ -26,3 +26,13 @@
...
+
+
+let url_regex = Pcre2_.regexp "^https?://"
+let is_url config_path =
+  match Pcre2_.pmatch ~rex:url_regex config_path with
+  | Ok true -> true
+  | Ok false -> false 
+  | Error err ->
+      Log.warn (fun m -> m "Error Validation URL: %a" pp_error err);
+      false
```
dimitris-m: Here you can just have : 

```ocaml 
| Ok res -> res 
| Error err  -> (* your code *)
```

Or even better, just use:

```ocaml
Pcre2_.pmatch_noerr
```

that will do the logging for you, ie: 

```ocaml
let is_url config_path = Pcre2_.pmatch_noerr ~rex:url_regex config_pat
```

(no need for `match` in that case).

jesse-merhi:
Yeah it would be awesome, especially for me because I am on MacOS to have a definitive list of how to compile.

dimitris-m:
> Yeah it would be awesome, especially for me because I am on MacOS to have a definitive list of how to compile.

I assume you have an OCaml switch, preferably with compiler version 5.2.1.
Make sure it's active and run `eval $(open env)` if asked.

As preparation ensure you run `make install-deps-MACOS-for-semgrep-core`.

At the root of the project do: 

```bash 
pushd libs/ocaml-tree-sitter-core; ./configure; ./scripts/install-tree-sitter-lib; popd
``` 

then try: `make core`.

libs/commons/Uri_.ml
dimitris-m: In fact your version will probably not compile, but my proposed solution with `pmatch_noerr` will.

dimitris-m: Notice that in `Pcre2_.ml`, line 27, we define: `module Log = (val Logs.src_log src : Logs.LOG)`.

This is used in the body of `pmatch_noerr` within `Pcre2_.ml`, but it's not visible in `Uri_.ml` which is why compilation failed.

jesse-merhi:
So after running those commands I get:

/Applications/Xcode.app/Contents/Developer/usr/bin/make minimal-build
dune build _build/install/default/bin/opengrep-core
File "libs/profiling/ppx/dune", line 7, characters 3-9:
7 |    ppxlib

dimitris-m:
> way

Add me on LinkedIn if you want for now, then we (Opengrep) might create other channels for contributors, such as Slack.

felix-montanari:
For reference, I've managed to get it building on my mac after running these commands:

```
opam switch create 5.2.1
opam switch 5.2.1
git submodule update --init --recursive
make install-deps
make install-deps-MACOS-for-semgrep-core
pushd libs/ocaml-tree-sitter-core; ./configure; ./scripts/install-tree-sitter-lib; popd
make install-opam-deps
make core
```

dimitris-m (approved):
lgtm, thanks for the fix :)

PR #69, opened 2025-02-04
Removing Semgrep refs from cli outputs

HenriqueOCabral:
CLI output ideas to cleanup some semgrep refs

### DIFF
--- a/cli/src/semgrep/output.py
+++ b/cli/src/semgrep/output.py
@@ -290,7 +290,7 @@ class OutputHandler:
             errors[path].sort()
             error_msg = f"Warning: {num_errs} timeout error(s) in {path} when running the following rules: [{separator.join(errors[path])}]"
             if num_errs == self.settings.timeout_threshold:
-                error_msg += f"\nSemgrep stopped running rules on {path} after {num_errs} timeout error(s). See `--timeout-threshold` for more info."
+                error_msg += f"\nOpengrep stopped running rules on {path} after {num_errs} timeout error(s). See `--timeout-threshold` for more info."
             print_threshold_hint = print_threshold_hint or (
                 num_errs > 5 and not self.settings.timeout_threshold
             )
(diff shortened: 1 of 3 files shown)


dimitris-m (commented):
Assuming you have a local installation -- for example after running `make core; make install` -- please run `make ci-test` from within the `cli/` directory. 

Are any tests failing with this change?

cli/src/semgrep/output.py
```diff
@@ -486,7 +486,7 @@ def output(
                 and state.metrics.is_using_registry
                 and (not auth.is_logged_in_weak())
             ):
-                suggestion_line = "\n(need more rules? `semgrep login` for additional free Semgrep Registry rules)\n"
+                suggestion_line = "\n(Want to share your rules? Create a PR on: https://github.com/opengrep/opengrep-rules) and contribute with the Opengrep community\n"
```
dimitris-m: Should we do that given the licensing question marks of the rules repo? Why not have a clean repo in that case, without possible issues? @willem-delbare 

That is, I propose we create a new, empty repo for such PRs.

cli/src/semgrep/output.py
```diff
@@ -496,9 +496,9 @@ def output(
                 ):
                     missed_count_line = f"💎 Missed out on {unit_str(missed_rule_count, 'pro rule')} since you aren't logged in!"
                     learn_more_url = with_color(
-                        Colors.cyan, "https://sg.run/rules", underline=True
+                        Colors.cyan, "https://github.com/opengrep", underline=True
                     )
-                    learn_more_line = f"⚡ Supercharge Semgrep OSS when you create a free account at {learn_more_url}."
+                    learn_more_line = f"⚡ Contribute with Opengrep OSS at {learn_more_url}."
```
dimitris-m: "Contribute to Opengrep at"

cli/src/semgrep/output.py
```diff
@@ -496,9 +496,9 @@ def output(
                 ):
                     missed_count_line = f"💎 Missed out on {unit_str(missed_rule_count, 'pro rule')} since you aren't logged in!"
                     learn_more_url = with_color(
-                        Colors.cyan, "https://sg.run/rules", underline=True
+                        Colors.cyan, "https://github.com/opengrep", underline=True
```
dimitris-m: We should probably put the full repo url to opengrep until other things are settled, ie, https://github.com/opengrep/opengrep.

dimitris-m (commented):
Good catch. 

Make some light recommendations, plus we need to check tests.

cli/src/semgrep/output.py
dimitris-m: For now we can remove this suggestion completely.

HenriqueOCabral: Now it is empty string. There is an if that only shows this message if findings were 0, we should keep just in case we want to use it in the near future:)

dimitris-m (approved):
I ran the python tests manually, LGTM!

Thanks!

PR #80, opened 2025-02-11
fix: FromAsCasing in Dockerfile

aviadhahami:
Update Dockerfile's casing to be uppercase so docker build won't crash on newer
docker engine versions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@
 # semgrep-core. This is done in a multi-stage build so that the final COPY
 # happens in a single layer.
 
-FROM busybox:stable as semgrep-core-files
+FROM busybox:stable AS semgrep-core-files
 WORKDIR /src/semgrep
 
 # copy over the entire semgrep repository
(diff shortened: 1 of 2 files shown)


Dockerfile
dimitris-m, line 1: Hi, sorry we forgot to remove this file, in fact this Dockerfile will not be used any more. See our Linux CI workflows.

If we were to adapt this file, our current minimum OCaml version is 5.3.0 (because we need the statistical profiler that was only re-introduced in that version, it was removed in OCaml 5.0.0).

Also this file has a few semgrep-specific stuff.

Perhaps a simpler Dockerfile based on what we do in `.github/workflows/build-test-core-x86.yml` would work better. Specifically, using the `ocamlpro/ocaml:5.3.0` image.

aviadhahami: @dimitris-m oh that's cool

So if I modify this Dockerfile to be up-to-date based on the CI config it should be fine?

aviadhahami:
@dimitris-m may I suggest we create a CI for image creation that will push to GHCR for easier use?

dimitris-m:
> @dimitris-m may I suggest we create a CI for image creation that will push to GHCR for easier use?

Right now we produce binaries that can be easily embedded into containers.

So I guess it can be convenient but it's not at the top of the list at this moment, because of other immediate priorities.

Having said that, an alpine based image following our Linux workflow should be very easy to deliver, for x86 and aarch64.

Dockerfile
dimitris-m: I think this could be useful and it would be a welcome contribution.

PR #83, opened 2025-02-14
Fix windows: use subprocess and wait, not execvp

dimitris-m:
### Changes

- The behaviour of `execvp` on Windows differs from the one in Unix systems, and it creates issues when we invoke Python to OCaml to Python ... 
- This PR ensures that on windows, we use subprocesses and we wait for completion of these subprocesses, which means the CLI behaves like on Unix, without strange spawning and killing of processes.

### DIFF
--- a/cli/src/semgrep/console_scripts/entrypoint.py
+++ b/cli/src/semgrep/console_scripts/entrypoint.py
@@ -194,8 +194,9 @@ def main():
 
     if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):  # PyInstaller
         os.environ["_OPENGREP_BINARY"] = sys.executable
-        if IS_WINDOWS: # Because `execvp` on windows does not replace the current process.
-            os.environ["PYINSTALLER_RESET_ENVIRONMENT"] = "1"
+        # NOTE: Not needed because we now use `subprocess.run` and we wait for completion.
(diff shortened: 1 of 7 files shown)


cli/src/semgrep/console_scripts/entrypoint.py
```diff
@@ -178,20 +179,24 @@ def exec_osemgrep():
...
 # Needed for similar reasons as in pysemgrep, but only for the legacy
 # flag to work
 def main():
 
     if getattr(sys, 'frozen', False) and hasattr(sys, '_MEIPASS'):  # PyInstaller
         os.environ["_OPENGREP_BINARY"] = sys.executable
-        if IS_WINDOWS: # Because `execvp` on windows does not replace the current process.
-            os.environ["PYINSTALLER_RESET_ENVIRONMENT"] = "1"
+        # NOTE: Not needed because we now use `subprocess.run` and we wait for completion.
+        # if IS_WINDOWS: # Because `execvp` on windows does not replace the current process.
```
willem-delbare, line 198: ?

dimitris-m: Comment to be removed on next PR, indeed it is no longer needed because we now do not allow the process to be killed which is what happens on Windows with execvp.

PR #100, opened 2025-02-22
Updated Installation script and README for installation instructions

deepakvijayaraj-official:
Updated Installation script and README for installation instructions
install.sh script that can be used for automating installation of latest version of opengrep or a specific version using install.sh -v option.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,221 @@
+#!/bin/bash
+# Script for installing opengrep on Linux and MacOS
+set -e
+
+# Color output
+YELLOW='\033[0;33m'
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+NC='\033[0m' # No Color
+
+# Global variable for temporary directory
+TEMP_DIR=""
+
+# Cleanup function
+cleanup() {
+    local exit_code=$?
+    printf "%b" "\n${YELLOW}Cleaning up...${NC}\n"
+    
+    # Remove temporary directory if it exists
+    if [ -n "$TEMP_DIR" ] && [ -d "$TEMP_DIR" ]; then
+        rm -rf "$TEMP_DIR"
+        printf "%b" "${YELLOW}Removed temporary directory: $TEMP_DIR${NC}\n"
+    fi
+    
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -0,0 +1,221 @@
...
+# Function to download and install the package
+install_package() {
+    local system=$1
+    local version=$2
+    local base_url="https://github.com/opengrep/opengrep/releases/download/${version}"
+    local binary_name="opengrep"
+    
+    # Create temporary directory and store in global variable
+    TEMP_DIR=$(mktemp -d)
+    printf "%b" "${YELLOW}Created temporary directory: $TEMP_DIR${NC}\n"
```
dimitris-m, line 140: Why not download directly to the target installation directory?

install.sh
```diff
@@ -0,0 +1,221 @@
...
+    
+    cd "$TEMP_DIR" || exit 1
+    printf "%b" "${YELLOW}Downloading OpenGrep ${version} for $system...${NC}\n"
```
dimitris-m, line 143: Not "OpenGrep", but "Opengrep".

install.sh
```diff
@@ -0,0 +1,221 @@
...
+TEMP_DIR=""
+
+# Cleanup function
+cleanup() {
+    local exit_code=$?
+    printf "%b" "\n${YELLOW}Cleaning up...${NC}\n"
+    
+    # Remove temporary directory if it exists
+    if [ -n "$TEMP_DIR" ] && [ -d "$TEMP_DIR" ]; then
+        rm -rf "$TEMP_DIR"
```
dimitris-m, line 21: In the event a bug is introduced into this script, this can be very dangerous. 

As mentioned in my other comment, please consider downloading directly into the install directory.

dimitris-m (changes requested):
I left some comments with some initial changes we should do, will return to this later for a more comprehensive review.

Here's a good install script, if you need inspiration: https://install.duckdb.org

dimitris-m:
Closing this as there is no response by author.

PR #123, opened 2025-03-06
Remove parmap, use domainslib

### LINKED ISSUE #18: Remove Parmap, at least for Windows
### Parmap is not compatible with Windows

This means we are forced to pass `-j 1` and everything runs sequentially. Therefore the Windows version is expected to be significantly degraded in terms of performance.

### Suggested solution

The less invasive approach is to replace the parmap wrapper in `libs/parallelism/Parmap_.mli`, and related parts, with something that works on Windows. This should take into account that we are moving immediately to OCaml 5, so we have some options. 

To minimise the impact, we could initially try to use Dune's _virtual libraries_ and have a new module _only_ for Windows. 

Note that our Windows build is already using OCaml 5.2.1.

dimitris-m:
### Changes 

⚠️ It is necessary to update submodules with `git pull --recurse-submodules; git submodule update --init --recursive` ⚠️

It is recommended to review this PR by looking into each commit separately, as they are relatively self-contained.

_**The following invariant needs to hold (and, hopefully, it does): parsing or analysing functions are never interleaved in a domain. They run from start to end before another xtarget can be processed. This is what makes our use of `Domain.DLS.key` safe.**_

- Transitions from parmap to domainslib
- Ensures that at least those shared data structures that are used by parallel tasks are thread-safe. Specifically, many hash tables are now replaced by lock-free versions from `Kcas_data`. There is more work to do but it's not blocking us for now. 
- Shared counters are now Atomic. Note that we use _a lot_ of increments to these counters, so ideally we should have one counter per parse and it could be a simple int reference, which would make these operations x2 faster. But it requires a lot of changes so is best left to be considered later. See `libs/commons/Gensym.ml`, following though we can see that the interesting functions are `empty_id_info` in `AST_generic.ml` and `SId.mk` which is used in `AST_to_IL.ml` and `Naming_AST.ml`. But the real issues is `empty_id_info` which is used in many places and it's not obvious how to add the extra counter parameter. In any case, this is probably not a bottleneck so will probably never change.
- Ensures that `lazy` are not forced in parallel, since this is not thread safe.
- Updates all the language submodules to point to our forks, which have been patched to be safe for use in domains. Specifically, the tree-sitter parsers are now domain-local.
- We now use our fork of memprof-limits also for global memory, since the existing solution is fragile when used with domains. But we lose some fancy features.
- The timeout functionality which was already ported to use our memprof-limits is now improved so that the program does not wait for timeout threads for more than a small threshold.
- Some cli printing functions are protected with a mutex.
- Logging is protected by a reentrant mutex, because we got deadlocks with OCaml's standard mutex. Maybe we can use the same mutex for the above function, remains as TODO.
- Classes were created for various objects, to make the code cleaner; this was not necessary but I ended up doing it while debugging some nasty concurrency bug (arising from a shared reference...).
- Fixed tests. Removed one test that checks the `--help` command, because it behaves differently on OSX vs Linux, although it can probably be fixed.
- Fixed the positioning of the `--experimental` flag, when it takes place automatically. It must not go before `scan` or `test` etc.
- Added notes, TODOs, and by necessity also FIXMEs since this work is already very extensive and there will still be issues. I already know that there is a chance of crash when printing reports, but it happened once after (150 | 2000 | 5000) sequential runs, so it will be addressed next. 
- Added notes in `Globals.ml`, expanding on what was already there.
- Some of the non tree-sitter lexers / parsers were not thread-safe: Scala, Js (secondary), Json, PHP (primary). This is now fixed.

Note that although the python tests are not executed in CI, I fixed them and it can be checked that `make ci-test` is now green, modulo 1 test that failed before on OSX.

Closes #18.


cli/src/semgrep/terminal.py
```diff
@@ -81,7 +81,7 @@ def configure(
         env = Env()
         # env.user_log_file dir must exist
         env.user_log_file.parent.mkdir(parents=True, exist_ok=True)
-        file_handler = logging.FileHandler(env.user_log_file, "w")
+        file_handler = logging.FileHandler(env.user_log_file, "w", encoding="utf-8")
```
willem-delbare, line 86: what's the impact of this , can we add a comment?

libs/commons/UCommon.ml
```diff
@@ -37,6 +37,7 @@ let pr2 s =
   UStdlib.prerr_string "\n";
   flush UStdlib.stderr
 
+(* FIXME: Get rid of that, not thread-safe. *)
```
willem-delbare: can we create github issues instead of FIXME?

libs/ojsonnet/Eval_jsonnet_subst.ml
```diff
@@ -36,6 +36,8 @@ module H = Eval_jsonnet_common
 (* Standard library *)
 (*****************************************************************************)
 
+(* FIXME: Check all those [Lazy.t] for possible concurrent access. *)
```
willem-delbare: github issue if it needs to be addressed?

libs/commons/UCommon.ml
dimitris-m: Yes, once merged we will create them -- but I guess they will only be relevant when we merge? 
Then we can remove the FIXME one by one. Easier tracking.

dimitris-m: Or where you thinking to eliminate the FIXME before merge? We could do that too.

libs/commons/UTmp.ml
```diff
@@ -67,21 +72,37 @@ let register_temp_file_cleanup_hook f = Stack_.push f temp_file_cleanup_hooks
 
 let default_temp_file_prefix = USys.argv.(0) |> Filename.basename
 
+(* TODO: To avoid collusions, we could set the DLS key for the temp folder
```
willem-delbare, line 74: github issue?

libs/commons/UTmp.ml
```diff
@@ -38,15 +38,20 @@ let temp_files_created : (Fpath.t, unit) Hashtbl.t = Hashtbl.create 101
...
     temp_files_created
-    |> Hashtbl.iter (fun path () ->
+    |> Kcas_data.Hashtbl.iter (fun path () ->
            Log.info (fun m -> m "deleting: %s" !!path);
            USys.remove !!path);
-    Hashtbl.clear temp_files_created)
+    Kcas_data.Hashtbl.clear temp_files_created)
 
 (* hooks for with_temp_file() *)
+(* FIXME? This does not look like it's used in a single thread.
```
willem-delbare: github issue?

libs/process_limits/Memory_limit.ml
```diff
@@ -55,7 +57,53 @@ let default_heap_warning_mb = 400
...
+
 *)
+
+(* We should not pass the limit as parameter because it's global. *)
+let set_global_memory_limit_mb mem_limit_mb =
+  (* NOTE: We get 0 from somewhere... memprof-limits takes -1 to mean 'no limit.' *)
+  let mem_limit_mb = if Int.equal mem_limit_mb 0 then -1 else mem_limit_mb in
+  M.set_global_memory_limit (mem_limit_mb * 1024) 
+
+(* TODO: The [mem_limit_mb] is global, and should not be a parameter, but the
```
willem-delbare: github issue?

libs/process_limits/Memory_limit.ml
```diff
@@ -55,7 +57,53 @@ let default_heap_warning_mb = 400
...
+        Gc.compact ();
+        raise (ExceededMemoryLimit (spf "Exceeded memory limit of %d MiB" mem_limit_mb))
+    | exception exn ->
+        let e = Exception.catch exn in
+        Logs.err (fun m -> m "exn while in run_with_global_memory_limit");
+        (* Try to free up some space. Expensive operation. *)
+        Gc.compact ();
+        Exception.reraise e
+
+(* FIXME: This is no longer used, except for tests. Remove.
```
willem-delbare: github issue?

libs/process_limits/Time_limit.ml
```diff
@@ -48,11 +48,26 @@ let string_of_timeout_info { Exception.name; max_duration } =
...
-let set_timeout (_caps : < Cap.time_limit >) ~name max_duration f =
+(* XXX: The sleeping threads delays the whole program! Hence the incremental sleeping,
+ * and the approach of stopping the sleeping thread when the work is done. Ideally,
+ * the granularity parameter should be set also by CLI flag. *)
+let set_timeout (_caps : < Cap.time_limit >) ?(granularity_float_s=0.1) ~name max_duration f =
   let token = M.Token.create () in
-  let timeout () = Thread.delay max_duration; M.Token.set token in
-  ignore (Thread.create timeout ());
-  match M.limit_with_token ~token f with
+  (* TODO:
```
willem-delbare: github issue?

src/analyzing/Constant_propagation.ml
```diff
@@ -103,29 +103,37 @@ let rec lvars_in_lhs expr =
   | Container ((Tuple | Array), (_, es, _)) -> List.concat_map lvars_in_lhs es
   | __else__ -> []
 
+class ['self] no_cycles_in_sym_prop_visitor =
```
willem-delbare, line 106: is this covered in a test? assume so :)

src/main/Main.ml
```diff
@@ -83,12 +83,30 @@
...
+  Array.find_index
+    (fun arg -> String.starts_with ~prefix:"-" arg)
+    argv
+
+let position_for_experimental_flag argv =
+  match first_hyphen_in_argv argv with
+  | None -> Array.length argv - 1
+  | Some i -> i
+
+(* TODO: Add some expoectation tests for such functions. *)
```
willem-delbare: github issue

src/analyzing/Constant_propagation.ml
dimitris-m: we did not change the code, just create a class from an "immediate" object with the same methods.

libs/process_limits/Time_limit.ml
dimitris-m: This one is not worth I guess, just wanted the comment to be there, will remove the TODO.

libs/commons/UTmp.ml
dimitris-m: #129

libs/commons/UTmp.ml
dimitris-m: #128

libs/process_limits/Memory_limit.ml
dimitris-m: #126

libs/process_limits/Memory_limit.ml
dimitris-m: #132

libs/ojsonnet/Eval_jsonnet_subst.ml
dimitris-m: #127

libs/process_limits/Time_limit.ml
dimitris-m: Converted to simple comment, since the code works well already.

src/analyzing/Constant_propagation.ml
dimitris-m: We ok with this? We did not change the logic here, just refactored a bit the code.

src/main/Main.ml
dimitris-m: #131

libs/commons/Common.ml
```diff
@@ -146,18 +146,38 @@ let finalize f cleanup =
...
-    | Not_found ->
+    (* NOTE: In a parallel scenario, we can of course run [f ()] more than once,
+     * so it's better to make sure that any [f] passed has only tolerable side
+     * effects (like reading a file we know should be there) and always returns
+     * the same result or raises the same exception. We could propably use a
+     * [Kcas] transaction if we wished to ensure that replacement of existing
+     * values does not happen, but I don't see the benefit for our purposes. *)
+    match Kcas_data.Hashtbl.find_opt h k with
+    | Some v -> v
+    | None ->
```
mbacarella, line 177: What do you think about perhaps saving future-us a lot of potential suffering by using a write lock in the None case? This way we ensure we only run `f ()` once? Just seems hard to require everyone to remember to make sure f is purely functional/has tolerable side effects.

libs/commons/Logs_.ml
```diff
@@ -14,6 +14,8 @@
  *)
 open Common
 
+module TLS = Thread_local_storage
```
mbacarella, line 17: nit: throwing around "TLS" might be confusing, like implying SSL TLS. could we stick to calling it DLS?

libs/commons/UTmp.ml
```diff
@@ -38,15 +38,19 @@ let temp_files_created : (Fpath.t, unit) Hashtbl.t = Hashtbl.create 101
 
 let save_temp_files = ref false
 
+(* XXX: This is called outside of the domainslib loop, so it should be ok. *)
 let erase_temp_files () =
   if not !save_temp_files then (
     temp_files_created
-    |> Hashtbl.iter (fun path () ->
+    |> Kcas_data.Hashtbl.iter (fun path () ->
```
mbacarella, line 45: if it's called outside of the domainslib loop can we just leave it as the simple Hashtbl? :thinking:

libs/commons2/common2.ml
```diff
@@ -845,10 +848,12 @@ let save_excursion_and_enable reference f =
 let memoized ?(use_cache = true) h k f =
   if not use_cache then f ()
```
mbacarella: can we just use reference the Commons.memoized? this one here in Commons2 seems like a copy/paste

mbacarella: (not sure why it was copy/pasted to begin with, but since it requires thinking about threads now, may as well just have one of them)

libs/parallelism/Domainslib_.ml
```diff
@@ -0,0 +1,65 @@
...
+   * since more than one task can run on the same thread. *)
+  assert (Int.equal chunksize 1);
+  (* It can be detrimental to performance if we go above the CPU count, so we
+   * place an upper bound. TODO: Add a log when this happens? *)
+  let num_domains = min num_domains (get_cpu_count ()) in
+  let pool = T.setup_pool ~num_domains:(num_domains - 1) () in
+  let xs_array = Array.of_list xs in
+  let res_array = Array.make (Array.length xs_array) None in
+  let f' x = wrap_result f ~exception_handler x in
+  Common.protect ~finally:(fun () -> T.teardown_pool pool) (fun () ->
```
mbacarella, line 52: curious: can this protect handler ever catch an exception? wrap_result catches all exceptions and turns them into Result.Errors

libs/parallelism/dune
```diff
@@ -5,6 +5,7 @@
    TCB
    parmap
```
mbacarella: also delete `parmap`?

libs/commons/Common.ml
```diff
@@ -146,18 +146,38 @@ let finalize f cleanup =
     res)
   else protect f ~finally:cleanup
 
+(* This is everywhere, and it's not thread-safe when the ref is non-local.
+ * It's best to force this to use DLS to be safe. We already don't interleave tasks
+ * on each domain at any time. But sometimes it's used for local variables and we
+ * should not pay the cost in these cases (see Visit_rule), so we need 2 versions
+ * of this function, and we must decide on each call site which one to use. *)
 let save_excursion reference newv f =
+  let module DLS = Domain.DLS in
```
mbacarella, line 155: aside: this whole (original) save_excursion pattern strikes me as really weird. it's probably nothing.

src/core_scan/Core_scan.ml
```diff
@@ -20,6 +20,7 @@ module ESet = Core_error.ErrorSet
 module MR = Mini_rule
 module R = Rule
 module Out = Semgrep_output_v1_j
+module TLS = Thread_local_storage
```
mbacarella, line 23: `module DLS = ...` ? TLS is kind of confusing

src/engine/Match_rules.ml
```diff
@@ -20,6 +20,7 @@ module RP = Core_result
 module E = Core_error
 module OutJ = Semgrep_output_v1_t
 module Log = Log_engine.Log
+module TLS= Thread_local_storage
```
maciejpirog: whitespace

dimitris-m: Fixed, thanks

libs/process_limits/Time_limit.mli
```diff
@@ -14,18 +14,27 @@ val string_of_timeout_info : timeout_info -> string
    Launch the specified computation and abort if it takes longer than
    specified (in seconds).
 
-   This uses a global timer. An Invalid_argument exception will be raised
+   This uses a local timer. An Invalid_argument exception will be raised
    if the timer is already running.
 
-   tl;dr nesting will fail
+   The [granularity_float_s] parameter is optional and specifies the amount
+   of time to sleep before checking if the memprof-limits token is set,
```
maciejpirog: Misleading comment (I suppose): In fact the thread sleeps and checks if `duration` is below `0.0`, in which case it **sets** the token, not checks if it is set

dimitris-m: Good point. Let me know if it's better now.

src/core_scan/Parmap_targets.ml
```diff
@@ -14,13 +14,6 @@
  *)
```
mbacarella: how come this file persists even though we've eliminated parmap?

mbacarella: rename?

src/osemgrep/core_runner/Core_runner.ml
```diff
@@ -98,7 +98,11 @@ let default_conf : conf =
      * TODO: figure out a solution for Windows multi-processing (OCaml 5 in
      * the worst case)
      *)
-    num_jobs = min 16 (if Sys.unix then Parmap_.get_cpu_count () else 1);
+    (* TODO: What if we are executing on a very large machine? The upper bound
```
mbacarella, line 101: created an issue: https://github.com/opengrep/opengrep/issues/139

dune-project
```diff
@@ -557,6 +557,12 @@ For more information see https://opengrep.dev.
     ; packages that would avoid this effect.
     git-unix
     (memprof-limits (= "dev")) ; NOTE: Just to invalidate CI cache for now.
+    thread-local-storage ; NOTE: We should not mess with memprof-limits, but in
+                         ; fact there is a TLS module in there also; however, it's
+                         ; not exposed in the public interface.
+    domainslib
+    kcas_data ; lock-free data structures, also makes life easier for refactoring.
+    batteries ; for reentrant mutex, used in logging.
```
mbacarella: batteries is huge. we should just cherry pick the re-entrant mutex out of it. follow-up issue here: 
https://github.com/opengrep/opengrep/issues/140

src/osemgrep/language_server/server/Session.ml
```diff
@@ -32,6 +32,7 @@ module OutJ = Semgrep_output_v1_j
 (* Refs *)
 (*****************************************************************************)
 
+(* XXX: Why is that a reference, it is not set anywhere. *)
 let scan_config_parser_ref = ref OutJ.scan_config_of_string
```
mbacarella: `(* XXX: Why is that a reference, it is not set anywhere. *)`

Correct, it doesn't need to be a ref. Fixed it.

libs/commons2/common2.ml
mbacarella: got rid of it. just uses Common.memoized now

libs/commons/Common.ml
dimitris-m: In the current usage it seems ok to run > 1 times, so I wouldn't do anything about this now.

libs/commons/Logs_.ml
dimitris-m: Hmm this would create confusion with the actual Domain.DLS module. 

In fact we may not need TLS at all but I don't want to remove it now.

libs/commons/UTmp.ml
dimitris-m: But is it?

libs/parallelism/Domainslib_.ml
dimitris-m: Yeah some of this code came from the original [Parmap_] module and may be redundant.

src/osemgrep/language_server/server/Session.ml
dimitris-m: Perfect

src/core_scan/Parmap_targets.ml
dimitris-m: Good idea, but maybe for a next PR, this is already huuuuge

dimitris-m: We could create issue for it, and delay the change?

libs/parallelism/dune
dimitris-m: done

src/core_scan/Parmap_targets.ml
dimitris-m: Ok I'll do it now

dimitris-m: Done

libs/commons/UTmp.ml
dimitris-m: This is not a hot path and it feels safer, normally I only changed hashtables that were used in parallel, leaving others alone.

dune-project
mbacarella: did this

maciejpirog (approved):

PR #143, opened 2025-03-11
fix: tools/languages_dumper builds again

mbacarella:
tools/languages_dumper broke because
* `parser_c.menhir` is now gone
* the `Flag_parsing.sgrep_mode` flag changed to a DLS key

fixed both

Not sure this tool is actually used, but it was stopping `dune build @check -w` in the top-level of the tree from succeeding


dimitris-m (approved):
Good catch. Indeed it's not used, that's why the project was building with a type error.

PR #168, opened 2025-03-20
Use opengrep's fork of semgrep-interfaces

maciejpirog:
The `semgrep-interfaces` repo consists the definition of the schema of the output. We want to make custom changes to it, which make it diverge from the semgrep's version, hence we point to our fork instead of the original semgrep's repo.


dimitris-m (approved):

dimitris-m:
Approved because the building phase passed so it's ok.

Issue #103, opened 2025-02-25
Expose Tree-sitter AST Context in Results


EdenBlumenstykk:
**Current state:**

Currently, even though OpenGrep already parses code using tree-sitter to create Abstract Syntax Trees (AST), valuable contextual information like containing class names and function names isn't exposed in the results. To get this information, we currently run our own tree-sitter parsing on the same files again.

**Desired state:**

Expose the AST contextual information that OpenGrep already has from its tree-sitter parsing in the results. For example, when finding a vulnerability in this code:

```class UserService {
    func validateInput() {
        db.Query(userInput)  // vulnerability found here
    }
}
```

We would get something like:
```
{
...
  "message": "SQL injection detected",
  "metadata": {
    "containing_class": "UserService",
    "containing_function": "validateInput"
  }
...
}
```

PR #169, opened 2025-03-20
[WIP] Add support for reporting enclosure of matches

maciejpirog:
An "enclosure" (a tentative name) is the syntactic context in which the match resides (A function, class definition, etc.)

Issue: #103

This is a work-in-progress PR. Tasklist:
- [x] Implement logic
- [ ] Add flag to turn this feature on
- [ ] Add tests

At the moment, the enclosure information is given in the `extra` section of the result for a rule. We reveal:

- The kind of the enclosing construct (function, class, or module),
- its name,
- its location in the code.

For example, matching `x = 5` in

```csharp
class MyClass {
  public int MyFunction() {
    int x;
    x = 5;
    return x;
  }
}
```

returns the following enclosure information:

```json
[
  {
    "kind":"function",
    "name":"MyFunction",
    "start":{"line":2, "col":3, "offset":18},
    "end":{"line":6, "col":4, "offset":83}
  },
  {
    "kind":"class",
    "name":"MyClass",
    "start":{"line":1,"col":1,"offset":0},
    "end":{"line":7,"col":2,"offset":85}
  }
]
```

The list is sorted from the innermost to the outermost constructs.

### DIFF
--- a/libs/ast_generic/Enclosure.ml
+++ b/libs/ast_generic/Enclosure.ml
@@ -1,0 +1,51 @@
+module AST = AST_generic
+
+type delimiter_kind = Module | Class | Func
+[@@deriving show, eq]
+
+(* delimiter_info values are included in Core_match.t, which is why we
+ * don't include the relevant AST nodes here, not to block the GC from
+ * cleaning up the AST of targets that have been already processed.
+ *)
+type delimiter_info = {
+  kind : delimiter_kind;
+  name : string;
+  range : (Tok.location * Tok.location) option
+}
+[@@deriving show, eq]
+
+type t = delimiter_info list
+[@@deriving show, eq]
+
+let human_readable_entity_name (name : AST.entity_name) : string =
+  match name with
+  | AST.EN (Id ((i, _tok), _)) -> i
+  | AST.EN (IdQualified {name_last = ((i, _tok), _type_args); _}) -> i
+  | _ -> failwith "impossible"
+
+let delimiter_kind_of_definition_kind (kind : AST.definition_kind) : delimiter_kind =
+  match kind with
+  | AST.FuncDef _ -> Func
+  | AST.ClassDef _ -> Class
+  | AST.ModuleDef _ -> Module
+  | _ -> failwith "impossible"
+
+let delimiter_info_of_stmt (stmt : AST.stmt) : delimiter_info =
+  match stmt.s with
+  | AST.DefStmt (entity, kind) ->
+      { kind = delimiter_kind_of_definition_kind kind;
+        name = human_readable_entity_name entity.name;
+        range = AST_generic_helpers.range_of_any_opt (S stmt) }
+  | _ -> failwith "impossible"
+
+let is_stmt_named_delimiter (stmt : AST.stmt) : bool =
+  match stmt.s with
+  | AST.DefStmt ({name = EN _; _},
+      (AST.FuncDef _ | AST.ClassDef _ | AST.ModuleDef _)) -> true
+  | _ -> false
+
+let delimiter_kind_for_output (k : delimiter_kind) : string =
+  match k with
+  | Module -> "module"
+  | Class -> "class"
+  | Func -> "function"
--- a/src/matching/Match_patterns.ml
+++ b/src/matching/Match_patterns.ml
@@ -153,7 +153,7 @@ let (rule_id_of_mini_rule : Mini_rule.t -> Core_match.rule_id) =
 
 let match_rules_and_recurse
     ({ m_env; path; hook; matches (* Core_match.t list ref *); has_as_metavariable } : env)
-    rules matcher k any x =
+    rules matcher k any enclosure x =
   rules
   |> List.iter (fun (pattern, rule) ->
          let matches_with_env = matcher rule pattern x m_env in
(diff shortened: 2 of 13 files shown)


libs/ast_generic/Enclosure.ml
```diff
@@ -0,0 +1,51 @@
...
+[@@deriving show, eq]
+
+type t = delimiter_info list
+[@@deriving show, eq]
+
+let human_readable_entity_name (name : AST.entity_name) : string =
+  match name with
+  | AST.EN (Id ((i, _tok), _)) -> i
+  | AST.EN (IdQualified {name_last = ((i, _tok), _type_args); _}) -> i
+  | _ -> failwith "impossible"
```
dimitris-m: maybe `assert false` if this really should not happen?

maciejpirog: Thanks, I wasn't sure what kind of "impossible" to use

maciejpirog: Refactored to avoid this altogether

src/matching/Match_patterns.ml
```diff
@@ -399,60 +415,59 @@ class ['self] check_visitor range_filter m_env has_as_metavariable
...
+                        Stack_.push pm mp_env.matches;
+                        hook pm));
+
+      match enclosure, Enclosure.delimiter_info_of_stmt x with
+      | Some enclosure_stack, Some delim ->
+          Stack_.push delim enclosure_stack;
+          super#visit_stmt env x;
+          ignore (Stack_.pop enclosure_stack)
+      | _ ->
+          super#visit_stmt env x;
```
dimitris-m: I think we could avoid the call to `Enclosure.delimiter_info_of_stmt x` if enclosure is `None`, this would be more optimal, no?

src/matching/Unit_matcher.ml
```diff
@@ -15,6 +15,43 @@ let t = Testo.create
...
+    (Rule_options.default, []) [rule]
+    (Fpath.v "some/path", Origin.File (Fpath.v "some/origin"), lang, program)
+  |> pred
+  |> Alcotest.(check bool)
+       (spf "failed test for tracking enclosing context in matching: |%s|" file)
+       true
+                                                                                                                 
+
+    (* Parse_python.parse_program (Fpath.v "tests/parsing/python/tuple_expansion.py")
+    *)
```
dimitris-m: Maybe cleanup / remove this.

src/matching/Unit_matcher.ml
```diff
@@ -15,6 +15,43 @@ let t = Testo.create
...
+  in
+  let rule =
+    Mini_rule.({id=Rule_ID.of_string_exn "myrule1";
+                pattern;
+                inside=false;
+                message="msg1";
+                metadata=None;
+                severity=`Warning;
+                langs=[lang];
+                pattern_string="patstr1";
```
dimitris-m: Maybe replace "patstr1" with "test: no need for pattern string" as is done elsewhere?

src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -602,6 +604,12 @@ let o_gitlab_secrets_outputs =
 
 let o_junit_xml_outputs = make_o_format_outputs ~fancy:"JUnit XML" "junit-xml"
 
+let o_output_enclosing_context : bool Term.t =
+  H.negatable_flag [ "output-enclosing-context" ] ~neg_options:[ "no-output-enclosing-context" ]
+    ~default:Match_patterns.default_matching_conf.track_enclosing_context
+    ~doc:
+      {|Include information about the syntactic context of the matched fragmetns of code, such as the function or the class in which the match is defined.|}
```
dimitris-m: typo: fragmetns

dimitris-m: what's the point of having the negated flag?

tests/match-patterns-check/cpp/namespace.cpp
```diff
@@ -0,0 +1,8 @@
+namespace Ns1
+{
+  template <typename X>
+  class  Class1
+  {
+    void fun1() { int i; void fun2(){ i = 5; }}
+  }
+}
```
dimitris-m, line 8: Is this used?

dimitris-m (approved):
lgtm, left some comments that you may want to address.

src/matching/Match_patterns.ml
maciejpirog: Right, fixed.

src/matching/Unit_matcher.ml
maciejpirog: Oops! Removed

src/matching/Unit_matcher.ml
maciejpirog: Fixed

src/osemgrep/cli_scan/Scan_CLI.ml
maciejpirog: only because other flags are negatable. Sometimes if you're running from a script, it's more convenient to pass the other flag than not to pass the flag at all

tests/match-patterns-check/cpp/namespace.cpp
maciejpirog: Oops! Added test data, but forgot to add the test. It is added now.

src/osemgrep/cli_scan/Scan_CLI.ml
maciejpirog: typo fixed

PR #178, opened 2025-03-31
Warn if --output-enclosing-context without --json

maciejpirog:
At the moment, the enclosing context captured when run with `--output-enclosing-context` shows up only in the `--json` output. So, if the user specifies the former but not the latter, we emit a warning that the former will have no effect.

NB. Adding enclosing context in other outputs should be easy and can be added to opengrep on-demand


dimitris-m (approved):
lgtm -- we should also make sure the warning is emitted in `opengrep ci ...`

maciejpirog:
> lgtm -- we should also make sure the warning is emitted in `opengrep ci ...`

Fixed

dimitris-m (approved):

Issue #189, opened 2025-04-03
Better support for String Templates in Kotlin


sampion88:
**Is your feature request related to a problem? Please describe.**
When looking for patterns with tainted user input in template strings like
```kotlin
fun foo(taintedString: String): String {
     return bar("This is bad ${taintedString}")
}
fun foo(taintedString: String): String {
     return bar("This is bad $taintedString")
}
fun foo(taintedString: String): String {
     return bar("""This is bad 
            $taintedString""")
}
```
It is possible to determine if your tainted user input goes into the template string.
But what is not possible, is something like this:
```yaml
- focus-metavariable: $TAINTED
- pattern-either:
  - pattern: "$BAD$TAINTED"
  - pattern: "$BAD${$TAINTED}"
- metavariable-regex:
      metavariable: $BAD
      regex: This is bad
```

**Describe the solution you'd like**
Make it possible to check for text inside a template string using metavariable-regex, also finding the tainted parameter (like it already does).

**Describe alternatives you've considered**
There is a workaround, but it is far from perfect and a bit buggy.
```yaml
- focus-metavariable: $TAINTED
- pattern-either:
    - pattern: |
        "...${$TAINTED}..."
    - pattern: |
        """...${$TAINTED}..."""
    - pattern-regex: |
        ^.*"("")?This is bad.*
```
I have to add the beginning of the string in the pattern-regex. But I cannot add the ending, or it does not work anymore, which makes it unfit for the multi line strings.

PR #191, opened 2025-04-04
Fix string templates in Kotlin

maciejpirog:
Issue: #189 

In Kotlin, string templates are of the form
```kotlin
"My name is ${name}, my age is ${realAge - 10}!"
```

Additionally, if the expression in `${ }` is a single variable, we can skip the braces, e.g.

```kotlin
"My name is $name, my age is ${realAge - 10}!"
```

This causes obvious clashes with opengrep's syntax for patterns. The semantics **before the fix** was:

- `"$FOO"` matches the entire string if it is a pure string literal,
- `"${$FOO}"` matches the expression within the template in a string template that consists of a single template, e.g. `"$i"`,
- `"$FOO`" within a template pattern always matches an expression within the template.

This means we have a problem, because:
- **There is no way to match a string segment in a template.**

It is a huge problem, because it is exactly what we need to capture things like SQL injection. For example, we would like a rule that recognises the following as a possible place of injection:

```kotlin
"SELECT $fieldName FROM ClientTable"
```

To specify that it is an injection, we want to capture `SELECT` and `FROM` using a regex, but to do this, we need to capture them as metavariables, which is now impossible!

Unfortunately, with the current matching infrastructure within opengrep, it seems impossible to fix it without a slight **change of semantics** of patterns. The semantics **after the fix**:

- `"$FOO"` matches a string segment
- `"${FOO}`" matches an expression within the template

For example, the example above can be captured with the following rule:

```
    pattern-sinks:
      - pattern-either:
          - patterns:	
              - pattern-either:
                  - pattern: |
                      "$FOO${$PARAM}$GOO"
              - metavariable-regex:
                  metavariable: $FOO
                  regex: (\s|.)*SELECT(\s|.)*
              - metavariable-regex:
                  metavariable: $GOO
                  regex: (\s|.)*FROM(\s|.)*
              - focus-metavariable: $PARAM
```

If we use this rule on the example above, we get:
- `$FOO` matches the `SELECT ` segment
- `$PARAM` matches the `fieldName` variable
- `$GOO` matches the ` FROM clientTable` segment

**WARNING**: The user needs to remember that `.` in regex DOES NOT capture newlines, so if you want the regexes to work in multiline string templates, we need to do sth like `(\s|.)` in the patterns for string segments.


languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
```diff
@@ -1933,20 +1933,6 @@ and string_literal (env : env) (v1, v2, v3) : expr =
...
-         which allows the existing literal metavariable machinery to run.
-
-         This does not affect Semgrep's expressive power, because a string containing
-         an interpolated identifier `X` can also be expressed in a Semgrep pattern via
-         `"{X}"`.
-      *)
-      let s1, t1 = str env v1 (* "$" *) in
-      let s2, t2 = simple_identifier env v2 in
-      G.L (G.String (l, (s1 ^ s2, Tok.combine_toks l [ t1; t2; r ]), r)) |> G.e
   | _ ->
```
dimitris-m: Now the:

```ocaml
match v2 with 
| _ -> 
```

is redundant

maciejpirog: Good catch! Fixed.

tests/patterns/kotlin/string-template.kt
dimitris-m, line 1: This is the same as: `tests/patterns/kotlin/not_string_metavar.kt`. 

Am I missing something? 

And given that `tests/patterns/kotlin/not_string_metavar.sgrep` is changed, the new test is exactly the same as the one modified?

dimitris-m: We should ideally have a test that would fail before if it was added, but passes now.

languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
```diff
@@ -1932,33 +1932,19 @@ and statements (env : env) ((v1, v2, v3) : CST.statements) =
...
-            | `Interp x -> interpolation env x)
-          v2
-      in
-      G.interpolated (l, v2, r)
+  let v2 =
+    List_.map
+      (function
+        | `Str_content tok ->
+            Either_.Left3 (str env tok)
+        | `Interp (`DOLLAR_simple_id (v1, v2)) when in_pattern env ->
```
dimitris-m, line 1940: Why keep `when in_pattern env` here?
Is this not how programs should be parsed?

maciejpirog: Indeed, inside interpolated strings we parse `$FOO` in patterns and in programs differently. In programs, they are the inner expressions in an interpolated string. In patterns, they are patterns that match string segments.
The context is that in every other language, `$FOO` in pattern matching an interpolated string means a string segment. For example, the pattern
```
$"{...}$FOO{...}"
```
in C# matches the following value
```
$"{someVal}xyz{someVal}"
```
with `$FOO` matching `xyz`.
In Kotlin, we want the pattern
```
"${...}$FOO${...}"
```
to match the following value
```
"${someVal}xyz${someVal}"
```
with $FOO matching, again, `xyz`.
This is the clash of syntax that is problematic here: `$FOO` in a pattern can be interpreted as a metavar matching a string segment (the convention used in every other language in opengrep) or a single-variable inner expression (as is the syntax of Kotlin). The choice proposed in this PR is that we arbitrarily say it is a string pattern, while if you want to match an inner expression, you need to add braces in the pattern (`${$FOO}`), which is semantically equivalent to `$FOO` in Kotlin and is not ambiguous.

tests/patterns/kotlin/string-template.kt
maciejpirog: No idea how this happened! Fixed now! The tests in this PR:
- `not_string_metavar` was changed, because of the change of syntax introduced in this PR
- `string_template` is a new test that would fail before the change. It tests that we can capture string segments in Ktolin -- something that was not possible before and is fixed in this PR

languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
dimitris-m: Ok, got it!

dimitris-m (approved):
LGTM 🚀 

One note: I wonder why `either3` was used here, since `Middle3` does not seem to appear, and probably `either` would suffice. But not worth to change this.

PR #194, opened 2025-04-07
Fix: .semgrepignore not working on windows when path is not relative

### LINKED ISSUE #193: on windows, .semgrepignore is ignored when scanning absolute paths

dimitris-m:
### Changes
- Fix fnmatch [.semgrepignore] patterns on windows, where / should become \\
- Fix dependencies (esp. base version)
- Fix Makefile, we moved the .opam files to opam/ and this was not reflected in all places
- Cleanup rolling release text
- Bump ubuntu 20.04 since CI started to fail, now we use the oldest available version when needed
- Fix a couple of env variables that we are migrating to have `OPENGREP_` prefix

Closes #193 in 274f993dffaac510d14a49b2c22971abcb696112

### DIFF
--- a/.github/workflows/build-test-windows-x86.yml
+++ b/.github/workflows/build-test-windows-x86.yml
@@ -243,9 +243,9 @@ jobs:
           opam env
 
           # opam pin --no-action add semgrep.dev .
-          echo "Pinning local packages..."
-          for f in opam/*.opam;
-            do opam pin --no-action add "$(basename "$f" .opam).dev" .; done;
+          # echo "Pinning local packages..."
+          # for f in opam/*.opam;
+          #   do opam pin --no-action add "$(basename "$f" .opam).dev" .; done;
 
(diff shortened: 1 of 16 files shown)


willem-delbare (approved):

.github/workflows/build-test-windows-x86.yml
```diff
@@ -243,9 +243,9 @@ jobs:
           opam env
 
           # opam pin --no-action add semgrep.dev .
-          echo "Pinning local packages..."
-          for f in opam/*.opam;
-            do opam pin --no-action add "$(basename "$f" .opam).dev" .; done;
+          # echo "Pinning local packages..."
```
willem-delbare, line 246: ?

dimitris-m: This was not needed and it was slowing down the CI

PR #195, opened 2025-04-08
Revive matching perf tests

maciejpirog:
Benchamarks for performance of matching inherited from semgrep. They don't run matching per se, but run `opengrep-core` with appropriate rules and programs.

Hit `run-perf-suite 2> /dev/null` to see the numbers


dimitris-m (approved):

maciejpirog:
> Noting comments made offline wrt relative path to binary.

This is fixed now

PR #201, opened 2025-04-09
Add union types to PHP menhir parser

### LINKED ISSUE #198: Opengrep cannot handle PHP parameter definition ...|null
**Describe the bug**
Starting from PHP 8.0 one can use 
```php
function foo(int|null $bar)
```
for nullable parameters instead of 
```php
function foo(?int $bar)
```
Apparently this messes up detection of findings that are after this point in a file.

**To Reproduce**
For following rule
```yaml
rules:
- id: foo_bar
  languages:
    - php
  severity: ERROR
  message: message.
  mode: taint
  pattern-sources:
    - patterns:
      - pattern: function $FUNC(..., $ARG, ...) {...}
      - focus-metavariable: $ARG
  pattern-sinks:
    - pattern-either:
      - patterns:
        - pattern: |
            "$BAD...$EXPR..."
        - metavariable-regex:
            metavariable: $BAD
            regex: ^\s*(foo_bar)([\s\S]*)?$

```
This code works:
```php
<?php

class foobar
{

	public function foo(int $id)
	{
		doSomething($id);
	}

	public function bar(string $input)
	{
		
        // ruleid: foo_bar
		$q = "
			foo_bar {$input}
		";

	}

}
```
and this one does not:
```php
<?php

class foobar
{

	public function foo(int|null $id)
	{
		doSomething($id);
	}

	public function bar(string $input)
	{
		
        // ruleid: foo_bar
		$q = "
			foo_bar {$input}
		";

	}

}
```

**Expected behavior**
Accept the use of |null in the funtion parameters

**Screenshots**

- [x] P2: regular bug that should get fixed

**Environment**
Are you running off docker, an official binary, a local build?

**Use case**
What will fixing this bug enable for you?

maciejpirog:
Our Menhir parser for PHP was missing union types.

For some reason, union types in the tree-sitter parser are translated to tuples. We do the same thing for menhir now.

Closes #198


dimitris-m (approved):
lgtm

PR #208, opened 2025-04-14
Fix logging mutex

dimitris-m:
### Changes 

- Remove reentrant mutex from logs, use normal Mutex: the reentrant mutex is not only slower but also using it it did not allow us to catch other bugs.
- Fix logging mutex release.
- Make a small change in `libs/ojsonnet/Value_jsonnet.ml`, since we seem to be logging recursively here, and 9 tests failed.

### DIFF
--- a/libs/commons/Logs_.ml
+++ b/libs/commons/Logs_.ml
@@ -192,6 +192,9 @@ let mk_reporter ?(additional_reporters : Logs.reporter list = []) ~dst
         msgf (fun ?header ?(tags = default_tag_set) fmt ->
             let pp_w_time ~tags =
               let current = now () in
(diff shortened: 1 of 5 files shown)


libs/commons/Logs_.ml
```diff
@@ -184,17 +187,14 @@ let mk_reporter ?(additional_reporters : Logs.reporter list = []) ~dst
           ((fun _ppf _style -> ()), "", "")
     in
     let k _ =
-      over ();
-      k ()
+      let v = k () in
+      v
     in
```
dimitris-m: We now don't need to redefine `k` here, will simplify.

dimitris-m: Maybe not

maciejpirog (approved):

PR #215, opened 2025-04-16
Add --opengrepignore-file flag 

tom-paz:
# Add `--opengrepignore-file` flag for custom ignore file configuration

## Summary
This PR introduces a new CLI flag `--opengrepignore-file` that allows users to specify a custom file for ignoring files during scanning, rather than relying on the default `.semgrepignore` file.

## Motivation
Currently, Opengrep uses `.semgrepignore` files to determine which files to exclude from scanning. However, users may want to:
- Use a differently named ignore file that better fits their project conventions (e.g., `.myignore`)
- Share ignore configurations between different tools
- Maintain multiple ignore configurations for different scanning purposes
- Use a global ignore file outside the project directory

## Implementation Details
- Added a new `--opengrepignore-file`  flag that accepts a path to a custom ignore file
- Updated the `Semgrepignore.ml` implementation to use the provided custom file instead of defaulting to `.semgrepignore`
- Updated documentation to use more general terminology for ignore files
- Modified the configuration processing to ensure the custom ignore file path is properly passed to the targeting system

## Usage Examples
```bash
# Use a custom ignore file
opengrep scan --config p/javascript --opengrepignore-file=.myignore .
```

### DIFF
--- a/src/osemgrep/cli_ci/Ci_CLI.ml
+++ b/src/osemgrep/cli_ci/Ci_CLI.ml
@@ -263,6 +263,7 @@ let scan_subset_cmdline_term : Scan_CLI.conf Term.t =
         respect_gitignore = use_git;
         respect_semgrepignore_files = not ignore_semgrepignore_files;
         exclude_minified_files;
+        semgrepignore_filename = None;
       }
     in
     let rule_filtering_conf =
@@ -315,6 +316,7 @@ let scan_subset_cmdline_term : Scan_CLI.conf Term.t =
         allow_local_builds;
         ls = false;
         ls_format = Ls_subcommand.default_format;
+        semgrepignore_filename = None;
       }
   in
   (* Term defines 'const' but also the '$' operator *)
@@ -378,13 +380,13 @@ let cmdline_term : conf Term.t =
     }
   in
   Term.(
-    const combine $ scan_subset_cmdline_term $ o_audit_on 
+    const combine $ scan_subset_cmdline_term $ o_audit_on
     $ o_dry_run $ o_internal_ci_scan_results
     $ o_x_dump_n_rule_partitions $ o_x_dump_rule_partitions_dir
     $ o_x_merge_partial_results_dir $ o_x_merge_partial_results_output
     $ o_x_partial_config $ o_x_partial_output
     $ o_x_validate_partial_results_actual
-    $ o_x_validate_partial_results_expected $ o_subdir 
+    $ o_x_validate_partial_results_expected $ o_subdir
     $ o_suppress_errors $ Git_metadata.env $ Github_metadata.env)
 
 let doc = "the recommended way to run opengrep in CI"
--- a/src/osemgrep/cli_ci/Ci_CLI.ml
+++ b/src/osemgrep/cli_ci/Ci_CLI.ml
@@ -186,15 +186,16 @@ let scan_subset_cmdline_term : Scan_CLI.conf Term.t =
       json json_outputs junit_xml junit_xml_outputs matching_explanations
       max_chars_per_line max_lines_per_finding max_log_list_entries
       max_memory_mb max_target_bytes metrics num_jobs no_secrets_validation
-      nosem optimizations oss output output_enclosing_context pro pro_intrafile pro_lang
-      pro_path_sensitive rewrite_rule_ids sarif sarif_outputs
+      nosem optimizations oss output output_enclosing_context pro pro_intrafile
+      pro_lang pro_path_sensitive rewrite_rule_ids sarif sarif_outputs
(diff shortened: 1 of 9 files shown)


dimitris-m:
Thanks for the PR.

A first comment before I review your PR, are you aware of the `SEMGREP_R2C_INTERNAL_EXPLICIT_SEMGREPIGNORE` environment variable which allows to set this?

tom-paz:
> Thanks for the PR.
> 
> A first comment before I review your PR, are you aware of the `SEMGREP_R2C_INTERNAL_EXPLICIT_SEMGREPIGNORE` environment variable which allows to set this?

Defenitely did not know about this env var, thanks.
Should we try and add this anyway?

dimitris-m (changes requested):
Please could you remove all changes that: 

- are just formatting changes, as this makes it harder to quickly evaluate your changes;
- are not relevant to what you are trying to do.

src/osemgrep/cli_ci/Ci_CLI.ml
```diff
@@ -186,15 +186,16 @@ let scan_subset_cmdline_term : Scan_CLI.conf Term.t =
...
-      _timeout_interfileTODO timeout_threshold (* trace trace_endpoint *) use_git
-      version_check vim vim_outputs =
+      _timeout_interfileTODO timeout_threshold
+      (* trace trace_endpoint *) use_git version_check vim vim_outputs =
     if output_enclosing_context && not json then
       Logs.warn (fun m ->
           m
-            "The --output-enclosing-context option has no effect without --json.");
+            "The --output-enclosing-context option has no effect without \
+             --json.");
```
dimitris-m, line 198: example of a change that seems unnecessary and makes it harder to filter out what matters.

there are several.

src/targeting/Semgrepignore.ml
```diff
@@ -11,26 +14,26 @@
    - '!' pattern negations aren't supported
    - character range patterns aren't supported
    - ':include' directives are a semgrepignore addition
-   - '.semgrepignore' files placed anywhere in the tree are ignored (?)
+   - ignore files placed anywhere in the tree are ignored (?)
```
dimitris-m, line 17: This would confuse users...

src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -1393,7 +1409,7 @@ let cmdline_term caps ~allow_empty_config : conf Term.t =
         dataflow_traces;
         nosem;
         strict;
-        time_flag;
+        time_flag = time;
```
dimitris-m, line 1412: These changes are not really needed for what you want to do in this PR.

src/targeting/Semgrepignore.ml
```diff
@@ -129,24 +132,40 @@ let create ?(cli_patterns = []) ~default_semgrepignore_patterns
...
+    if exclusion_mechanism.use_semgrepignore_files then
+      [ custom_semgrepignore_files ]
     else []
   in
-  (*
-     Check if there is a top-level '.semgrepignore'. If not, use builtins.
 
-     We don't check for '.semgrepignore' down the tree, so if a user needs
-     to override the default semgrepignore rules, they need at least an
-     empty root '.semgrepignore' file.
```
dimitris-m, line 144: This seems like important context, why is it being deleted...?

src/targeting/Semgrepignore.ml
```diff
@@ -129,24 +132,40 @@ let create ?(cli_patterns = []) ~default_semgrepignore_patterns
...
+     If a custom semgrepignore file is provided, check if it exists.
+     Otherwise, check for the default .semgrepignore file.
   *)
   let root_semgrepignore_exists =
-    let root_dir = Ppath.to_fpath ~root:project_root Ppath.root in
-    let semgrepignore_path = Fpath.add_seg root_dir ".semgrepignore" in
-    Sys.file_exists (Fpath.to_string semgrepignore_path)
+    match semgrepignore_filename with
+    | Some filename ->
+        Sys.file_exists filename
```
dimitris-m, line 164: `Fpath.to_string` will probably normalise the path as required, this is not the same as the `None` case.

src/targeting/Semgrepignore.mli
```diff
@@ -1,8 +1,8 @@
 (*
-   Parse and interpret '.semgrepignore' files in addition to '.gitignore'
+   Parse and interpret ignore configuration files in addition to '.gitignore'
```
dimitris-m, line 2: Changing terminology will confuse a lot of users, like I mentioned before...

dimitris-m:
As a further note, this PR only edits the OCaml CLI, but the python part has no such option.

PR #216, opened 2025-04-16
Add Custom Ignore Pattern Support for Code Comments

tom-paz:
# Add Custom Ignore Pattern Support for Code Comments

## Summary

This PR adds support for custom ignore pattern comments in addition to the default `nosem` and `nosemgrep` comment prefixes. This allows users to use their own custom prefix to suppress findings with a comment.

## Implementation Details

- Added a new command-line flag `--opengrep-ignore-pattern=<pattern>` that lets users specify a custom ignore pattern
- Made the regex pattern generation dynamic so it evaluates the patterns each time they're needed
- Improved pattern matching using word boundary markers (`\b`) to ensure exact matches
- Added proper logging instead of direct console output

## Example Usage

```javascript
// Default patterns still work
// nosem
console.log("This will be ignored");

// nosemgrep
console.log("This will also be ignored");

// Custom pattern works when flag is specified
// noopengrep 
console.log("This will be ignored when --opengrep-ignore-pattern=noopengrep is used");
```

With this change, users can define organization-specific ignore patterns that align with their coding practices.

### DIFF
--- a/cli/src/semgrep/constants.py
+++ b/cli/src/semgrep/constants.py
@@ -65,22 +65,22 @@ class RuleScanSource(Enum):
 
 RULE_ID_RE_STR = r"(?:[:=][\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?"
 
-# Inline 'noqa' implementation modified from flake8:
-# https://github.com/PyCQA/flake8/blob/master/src/flake8/defaults.py
-# We're looking for items that look like this:
-# ' nosem'
-# ' nosemgrep: example-pattern-id'
-# ' nosem: pattern-id1,pattern-id2'
-# ' NOSEMGREP:pattern-id1,pattern-id2'
-#
-# * We do not want to capture the ': ' that follows 'nosem'
-# * We do not care about the casing of 'nosem'
-# * We want a comma-separated list of ids
-# * We want multi-language support, so we cannot strictly look for
-#   Python comments that begin with '# '
-# * nosem and nosemgrep should be interchangeable
-#
-NOSEM_INLINE_RE_STR = r" nosem(?:grep)?" + RULE_ID_RE_STR
+# Custom ignore pattern can be set using --opengrep-ignore-pattern option
+# When set, this pattern replaces the default 'nosem' and 'nosemgrep' patterns
+CUSTOM_IGNORE_PATTERN = None
+
+# Function to get the appropriate pattern choice based on custom ignore pattern
+def get_nosem_pattern_choices():
+    """Returns either the custom pattern or the default 'nosem(?:grep)?' pattern"""
+    if CUSTOM_IGNORE_PATTERN:
+        # When custom pattern is set, use only this pattern 
+        return CUSTOM_IGNORE_PATTERN
+    else:
+        # When no custom pattern, use the original pattern
+        return "nosem(?:grep)?"
+
+# Define regex patterns using the pattern choices
+NOSEM_INLINE_RE_STR = r" " + get_nosem_pattern_choices() + RULE_ID_RE_STR
 NOSEM_INLINE_RE = re.compile(NOSEM_INLINE_RE_STR, re.IGNORECASE)
 
 # As a hack adapted from semgrep-agent,
@@ -98,7 +98,7 @@ NOSEM_INLINE_COMMENT_RE = re.compile(rf"[:#/]+{NOSEM_INLINE_RE_STR}$", re.IGNORE
 #   # nosemgrep
 #   print('nosemgrep');
 NOSEM_PREVIOUS_LINE_RE = re.compile(
-    r"^[^a-zA-Z0-9]* nosem(?:grep)?" + RULE_ID_RE_STR,
+    r"^[^a-zA-Z0-9]* " + get_nosem_pattern_choices() + RULE_ID_RE_STR,
     re.IGNORECASE,
 )
 
--- a/cli/src/semgrep/constants.py
+++ b/cli/src/semgrep/constants.py
@@ -65,6 +65,9 @@ class RuleScanSource(Enum):
 
 RULE_ID_RE_STR = r"(?:[:=][\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?"
 
+# Custom ignore pattern can be set using --opengrep-ignore-pattern option
+CUSTOM_IGNORE_PATTERN = None
+
 # Inline 'noqa' implementation modified from flake8:
 # https://github.com/PyCQA/flake8/blob/master/src/flake8/defaults.py
 # We're looking for items that look like this:
@@ -72,6 +75,7 @@ RULE_ID_RE_STR = r"(?:[:=][\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?"
 # ' nosemgrep: example-pattern-id'
 # ' nosem: pattern-id1,pattern-id2'
 # ' NOSEMGREP:pattern-id1,pattern-id2'
+# ' noopengrep: example-pattern-id' (if --opengrep-ignore-pattern=noopengrep is set)
 #
 # * We do not want to capture the ': ' that follows 'nosem'
 # * We do not care about the casing of 'nosem'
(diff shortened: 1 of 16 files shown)


dimitris-m (commented):
Hi -- I recommend you to first open issues if you want to propose a feature or change, in order to be able to gauge interest _before_ we proceed to make any changes. 

What is the benefit of this feature? It will be harder to know which comments have meaning for opengrep.

tom-paz:
> Hi -- I recommend you to first open issues if you want to propose a feature or change, in order to be able to gauge interest _before_ we proceed to make any changes.
> 
> What is the benefit of this feature? It will be harder to know which comments have meaning for opengrep.

Hey, this feature is extremely useful for unified syntax for different formatters and validators. This will also help creating a unified prefix for certain projects/organizations that do not want to necessarily commit to a single scanner in the actual code.

dimitris-m:
> > Hi -- I recommend you to first open issues if you want to propose a feature or change, in order to be able to gauge interest _before_ we proceed to make any changes.
> > What is the benefit of this feature? It will be harder to know which comments have meaning for opengrep.
> 
> Hey, this feature is extremely useful for unified syntax for different formatters and validators. This will also help creating a unified prefix for certain projects/organizations that do not want to necessarily commit to a single scanner in the actual code.

That assumes that _all_ scanners can be customised, which is unlikely. 

And the price to pay is: hard to understand what should be ignored by opengrep. 

Do you have a concrete need for this? Which other scanner are you using that requires this?
In summary, is this PR solving a concrete problem for you?

I'm not so positive about this change but will mention it to the other members.

tom-paz:
> > > Hi -- I recommend you to first open issues if you want to propose a feature or change, in order to be able to gauge interest _before_ we proceed to make any changes.
> > > What is the benefit of this feature? It will be harder to know which comments have meaning for opengrep.
> > 
> > 
> > Hey, this feature is extremely useful for unified syntax for different formatters and validators. This will also help creating a unified prefix for certain projects/organizations that do not want to necessarily commit to a single scanner in the actual code.
> 
> That assumes that _all_ scanners can be customised, which is unlikely.
> 
> And the price to pay is: hard to understand what should be ignored by opengrep.
> 
> Do you have a concrete need for this? Which other scanner are you using that requires this? In summary, is this PR solving a concrete problem for you?
> 
> I'm not so positive about this change but will mention it to the other members.

tom-paz:
> > > Hi -- I recommend you to first open issues if you want to propose a feature or change, in order to be able to gauge interest _before_ we proceed to make any changes.
> > > What is the benefit of this feature? It will be harder to know which comments have meaning for opengrep.
> > 
> > 
> > Hey, this feature is extremely useful for unified syntax for different formatters and validators. This will also help creating a unified prefix for certain projects/organizations that do not want to necessarily commit to a single scanner in the actual code.
> 
> That assumes that _all_ scanners can be customised, which is unlikely.
> 
> And the price to pay is: hard to understand what should be ignored by opengrep.
> 
> Do you have a concrete need for this? Which other scanner are you using that requires this? In summary, is this PR solving a concrete problem for you?
> 
> I'm not so positive about this change but will mention it to the other members.

This is a very important feature for us, we use internal scanners for our customers as well that require ignoring certain code lines just like the `nosemgrep` comment does. This helps us and our customers put a single comment instead of replacing by need or when an additional scanner is added. 

This PR is defenitely solving a concrete problem for us .

dimitris-m:
> > > > Hi -- I recommend you to first open issues if you want to propose a feature or change, in order to be able to gauge interest _before_ we proceed to make any changes.
> > > > What is the benefit of this feature? It will be harder to know which comments have meaning for opengrep.
> > > 
> > > 
> > > Hey, this feature is extremely useful for unified syntax for different formatters and validators. This will also help creating a unified prefix for certain projects/organizations that do not want to necessarily commit to a single scanner in the actual code.
> > 
> > 
> > That assumes that _all_ scanners can be customised, which is unlikely.
> > And the price to pay is: hard to understand what should be ignored by opengrep.
> > Do you have a concrete need for this? Which other scanner are you using that requires this? In summary, is this PR solving a concrete problem for you?
> > I'm not so positive about this change but will mention it to the other members.
> 
> This is a very important feature for us, we use internal scanners for our customers as well that require ignoring certain code lines just like the `nosemgrep` comment does. This helps us and our customers put a single comment instead of replacing by need or when an additional scanner is added.
> 
> This PR is defenitely solving a concrete problem for us .
> 
> As to the price to pay: I dont see how that is an issue, for those who use `opengrep` to scan, they should control and know what they are using, besides, it would be nice to pass `noopengrep` by default if thats something you is interesting,

Ok, will be considered and reviewed in detail, thanks.

cli/src/semgrep/constants.py
```diff
@@ -65,22 +65,22 @@ class RuleScanSource(Enum):
...
-# ' nosem: pattern-id1,pattern-id2'
-# ' NOSEMGREP:pattern-id1,pattern-id2'
-#
-# * We do not want to capture the ': ' that follows 'nosem'
-# * We do not care about the casing of 'nosem'
-# * We want a comma-separated list of ids
-# * We want multi-language support, so we cannot strictly look for
-#   Python comments that begin with '# '
-# * nosem and nosemgrep should be interchangeable
-#
```
dimitris-m: Once more, here you are deleting important context.

dimitris-m: For this PR to have a chance of getting merged into main, you will need to be surgical about the changes you want to do and avoid deleting useful context / info...

tom-paz: 👍 Reverted the comments, what other context did I delete?

tom-paz: Hey, any chance we can get an estimation for this PR ?

src/configuring/Flag_semgrep.ml
```diff
@@ -43,6 +43,25 @@ let equivalence_mode = ref false
 
 let output_enclosing_context = ref false
 
+(* Custom pattern to replace or supplement the default 'nosem' or 'nosemgrep' prefixes
+ * for comments that should be ignored by opengrep. By default this is None, which means
+ * the system will use only the standard 'nosem' or 'nosemgrep' prefixes.
+ *)
+let custom_ignore_pattern : string option ref = ref None
```
maciejpirog: Can this flag be passed via a config datatype passed to reporting rather than via a global ref? We're in the process of removing all global state, and the `Flag_semgrep.ml` file is deemed to go altogether.

src/reporting/Nosemgrep.ml
```diff
@@ -138,6 +156,10 @@ let rule_match_nosem (pm : Core_match.t) : bool * Core_error.t list =
 
   let no_ids = List.for_all Option.is_none in
 
+  (* Get the regexes fresh each time to ensure we use the latest custom pattern *)
```
maciejpirog: What is the "latest" pattern? My understanding is that the ignore patterns do not change over time, as they are passed via the flag, so why would we expect two calls to `get_nosem_inline_re` or `get_nosem_previous_line_re` to return different results in two different points in time?

src/reporting/Nosemgrep.ml
```diff
@@ -52,8 +53,26 @@ let rule_id_re_str = {|(?:[:=][\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?|}
      Python comments that begin with '# '
    * nosem and nosemgrep should be interchangeable
 *)
-let nosem_inline_re_str = {| nosem(?:grep)?|} ^ rule_id_re_str
-let nosem_inline_re = Pcre2_.regexp nosem_inline_re_str ~flags:[ `CASELESS ]
+
+let get_nosem_pattern_choices () =
+  let patterns =
+    match !Flag_semgrep.custom_ignore_pattern with
```
maciejpirog: This match is already present in `Flag_semgrep.get_ignore_patterns`.

src/configuring/Flag_semgrep.ml
```diff
@@ -43,6 +43,25 @@ let equivalence_mode = ref false
...
+
+(* The list of patterns to use for ignoring lines. Always includes the standard patterns,
+ * plus any custom pattern that has been set.
+ *)
+let get_ignore_patterns () : string list =
```
maciejpirog: This function is not use anywhere.

maciejpirog (changes requested):
The functionality looks perfectly legitimate, and I definitely agree we shouldn't have any hard-coded constants like `nosem` or `nosemgrep` that we can't change via a flag. However:
- Let's not add a new global ref,
- I don't understand the need for `nosem_inline_re` and `nosem_previous_line_re` to be functions. Is there something non-trivial happening here?

src/reporting/Nosemgrep.ml
tom-paz: Correct, this was a misleading comment. remove it

maciejpirog (approved):
After the changes, the PR looks good. Good job, thanks!

PR #228, opened 2025-04-22
Fix INSTALL.md manual development instruction

aayachnes:


dimitris-m (changes requested):
Have you tested this?

From a quick look, there are a couple of issues to be fixed in the INSTALL.md file, if you want to improve it:

- I would replace `cd semgrep` (in your branch, `cd opengrep`) with `cd cli/src/semgrep` to be more clear about which directory this refers to. (PS: I can understand the confusion).
- Below in the same file there is something to fix: 

```
opengrep-dev() {
  PIPENV_PIPFILE=~/semgrep/cli/Pipfile pipenv run opengrep "$@"
}
```

should be 

```
opengrep-dev() {
  PIPENV_PIPFILE=~/opengrep/cli/Pipfile pipenv run opengrep "$@"
}
```

perhaps with a clarification that the assumption is that the repo is checked out at `~/opengrep`.

Plus:

The reference to "Semgrep" on line 37.

And the section on docker which is deprecated (we don't support the original dockerfile).

dimitris-m:
As a side note, my workflow is as follows: 

- go into `cli`
- run `pipenv shell`
- whenever I make changes, `make core` and then `make install` and everything works

PR #238, opened 2025-04-29
Allow multiple files with test command

### LINKED ISSUE #237: Ability to use multiple test files for a rule with `opengrep --test`
Currently, Opengrep matches test files on the filename (for example, my_rule.test.yaml will be tested on file my_rule.test.js). `opengrep --test` will not look for multiple test files nor variations (e.g. `.test1.js`). This limits the ability to test for file global sanitizers (e.g. an import as a sanitizer which would ignore any results in the entire file).

Support multiple test files for rules, e.g. `<ruleid>.test<1 ... n>.yaml`.

maciejpirog:
All correct previous commands work the same

New option is to specify multiple files:

```
opengrep test -c path/to/rules.yaml file1.py file2.py
```

Or (when run as a shell command):

```
opengrep test -c path/ro/rules.ysml file*.py
```

Closes #237

### DIFF
--- a/src/osemgrep/cli_test/Test_CLI.ml
+++ b/src/osemgrep/cli_test/Test_CLI.ml
@@ -44,7 +44,7 @@ type conf = {
  *)
 and target_kind =
   | Dir of Fpath.t * Rules_config.config_string option (* optional --config *)
-  | File of Fpath.t * Rules_config.config_string (* mandatory --config *)
+  | Files of Fpath.t list * Rules_config.config_string (* mandatory --config *)
 [@@deriving show]
 
 (*************************************************************************)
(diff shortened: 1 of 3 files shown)


src/osemgrep/cli_test/Test_CLI.ml
```diff
@@ -128,19 +128,20 @@ let target_kind_of_roots_and_config target_roots config =
...
+      else Files ([file], config)
   | [ file ], [] ->
       if Sys.is_directory !!file then Dir (file, None)
       else
         (* was raise Exception but cleaner abort I think *)
         Error.abort "--config is required when running a test on single file"
-  | _ :: _ :: _, _ ->
-      (* stricter: better error message '(directory or file)' *)
-      Error.abort "only one target (directory or file) allowed for tests"
+  | [], _ -> Error.abort "bug: no valid target roots"
```
dimitris-m: I would not say "bug" here as it's invalid input.

dimitris-m: maybe "at least one target required for tests"

dimitris-m (approved):
lgtm -- left some minor comments

src/osemgrep/cli_test/Test_CLI.ml
maciejpirog: fixed according to your suggestion

Issue #92, opened 2025-02-17
Inconsistent miss of matches in C#


nir-valtman:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behavior.
1. Add the following rule:
```yaml
rules:
  - id: identified-parameter-in-csharp-code
    languages:
      - csharp
    severity: ERROR
    message: Detected the parameter name(s) `$PARAM` inside `$FUNC`
    patterns:
      - pattern-either:
          - patterns:
              - pattern: |-
                  enum $FUNC {
                    ...,
                    $PARAM,
                    ...
                  }
              - focus-metavariable: $PARAM
          - patterns:
              - pattern-either:
                  - pattern: $X $FUNC(..., $T $PARAM, ...) { ... }
                  - pattern: $T $FUNC = (..., $PARAM, ...) => ...;
              - metavariable-regex:
                  metavariable: $T
                  regex: (?i)^(?!(boolean|int|float|double))
              - focus-metavariable: $PARAM
          - patterns:
              - pattern-either:
                  - pattern: $T $PARAM;
                  - pattern: $T $PARAM = $VAL;
                  - pattern: $T $PARAM = new $U(...);
              - pattern-inside: |-
                  class $FUNC {
                    ...
                  }
              - pattern-not-inside: |-
                  $X $FUNC2(...) {
                      ...
                  }
              - metavariable-regex:
                  metavariable: $T
                  regex: (?i)^(?!(boolean|int|float|double))
              - focus-metavariable: $PARAM
          - patterns:
              - pattern-either:
                  - pattern: |-
                      $T $FUNC(...) 
                      { 
                        ... 
                        return $PARAM;
                        ...  
                      }
                  - pattern: |-
                      $T $FUNC(...) 
                      { 
                        ... 
                        return (..., $PARAM, ...);
                        ...  
                      }
                  - pattern: |-
                      $T $FUNC(...) 
                      { 
                        ... 
                        return new $TYPE(..., $PARAM, ...);
                        ...  
                      }
              - pattern-not-inside: |-
                  class $CLASS {
                    ...
                    $VAR $PARAM;
                    ...
                  }
              - pattern-not-inside: |-
                  class $CLASS {
                    ...
                    $VAR $PARAM = $VAL;
                    ...
                  }
              - focus-metavariable: $PARAM
      - metavariable-regex:
          metavariable: $PARAM
          regex: (?i)(alice|bob|eve)

```
2. Add the following file:
```cs
[...]

PR #239, opened 2025-04-30
Fix: relax c sharp parser issue 92 regression

dimitris-m:
We could no longer parse some patterns and code that were OK before, after the last change.

Now the grammar is a bit simplified and the examples work, see: 
https://github.com/opengrep/semgrep-c-sharp/tree/dm/opengrep-issue-92-b

Related: Issue #92, PR #231 .


maciejpirog (approved):
lgtm

PR #247, opened 2025-05-08
Add script to compare running time of two executables

maciejpirog:
For now you need to edit the script directly to change the paraemters (commands, repos, number of repetitions of each test)

### DIFF
--- a/perf/opengrep-scripts/cmp-execs.py
+++ b/perf/opengrep-scripts/cmp-execs.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+import sys
+import time
+import csv
+from urllib.parse import urlparse
+from datetime import datetime
+import statistics
+import resource
+
+# SETUP
+
+# TODO: make some of these cli options?
+
+repeat_each_test_n_times = 3
+
+semgrep_command = "semgrep"
+
+opengrep_command = "opengrep"
+
+repositories = [ ("https://github.com/golang/go","93e3d5dc5f2af317c874fd61cbd354409ea9fd33")
+               , ("https://github.com/grafana/grafana", "afcb55156260fe1887c4731e6cc4c155cc8281a2")
+               , ("https://github.com/rust-lang/rust","0f73f0f3941e6be6b19721548fab4e2bf919a525")
+               , ("https://github.com/torvalds/linux","01f95500a162fca88cefab9ed64ceded5afabc12")
+               , ("https://github.com/tensorflow/tensorflow","71691c769e76655e5b2b869f258572d1e8febe64")
+               , ("https://github.com/nodejs/node","f275121b72916d0c0a8b017b7677ddd6e2573e46")
+               , ("https://github.com/kubernetes/kubernetes","2ac0bdf360cf2529a3675c7012d0bf415e1051f3")
+               ]
+
+num_cpus = str(max(1, os.cpu_count() - 1))
+
+def regular_cmd(cmd, repo):
+    return [cmd, "scan",
+            "-c", "rules",
+            repo,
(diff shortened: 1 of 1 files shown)


perf/opengrep-scripts/cmp-execs.py
```diff
@@ -0,0 +1,161 @@
...
+
+# SETUP
+
+# TODO: make some of these cli options?
+
+repeat_each_test_n_times = 5
+
+command1 = "semgrep"
+
+command2 = "opengrep"
```
dimitris-m: Why not rename these to `semgrep_command` and `opengrep_command`?

perf/opengrep-scripts/cmp-execs.py
```diff
@@ -0,0 +1,161 @@
...
+    sys.stdout.flush()
+    subprocess.run(cmd, cwd=cwd, check=True)
+
+def run_command_with_stdout(command):
+    result = subprocess.run(command, shell=True, text=True, capture_output=True)
+    return result.stdout
+
+def clone_specific_commit(repo_url, commit_hash, name):
+    if os.path.exists(name):
+        print(f"Repository '{name}' present.")
```
dimitris-m, line 76: But this should fail maybe? 
Or checkout the commit hash if possible (failing if there are uncommitted changes, for example)?

perf/opengrep-scripts/cmp-execs.py
```diff
@@ -0,0 +1,161 @@
...
+    run(["rm", "-f", ".pre-commit-config.yaml"], cwd="rules")
+
+    for r in repos:
+        clone_specific_commit(r["url"], r["sha"], "repos/" + r["name"])
+
+def single_run(cmd, repo):
+    time.sleep(10)
+    usage_start = resource.getrusage(resource.RUSAGE_CHILDREN)
+    run(regular_cmd(cmd, repo))
+    usage_end = resource.getrusage(resource.RUSAGE_CHILDREN)
```
dimitris-m, line 116: what does that do, maybe add some comment?

perf/opengrep-scripts/cmp-execs.py
```diff
@@ -0,0 +1,161 @@
...
+        writer.writeheader()
+        writer.writerows(with_stats)
+
+def go():
+    setup()
+    log_to_file(f"OPENGREP vs SEMGREP TEST {ts}\n\n")
+    print(f"which opengrep:\n{run_command_with_stdout('which opengrep')}")
+    print(f"which semgrep:\n{run_command_with_stdout('which semgrep')}")
+    log_to_file(f"which opengrep:\n{run_command_with_stdout('which opengrep')}")
+    log_to_file(f"which semgrep:\n{run_command_with_stdout('which semgrep')}")
```
dimitris-m: Would be good to log the respective versions too (eg 1.1.5 vs 1.100.0).

dimitris-m (approved):
Left some comments, otherwise lgtm

perf/opengrep-scripts/cmp-execs.py
maciejpirog: good idea! will fix!

perf/opengrep-scripts/cmp-execs.py
maciejpirog: but these are repositories cloned by the script, the user is not supposed to touch them

perf/opengrep-scripts/cmp-execs.py
maciejpirog: fixed

perf/opengrep-scripts/cmp-execs.py
maciejpirog: added some comments

Issue #259, opened 2025-05-19
opengrep-ignore-pattern flag doesnt work with the 'experimental' flag combination


tom-paz:
**Describe the bug**
The `--opengrep-ignore-pattern` flag only takes effect when the `--experimental` flag is *not* used. When running with `--experimental`, the custom ignore pattern is ignored and the default behavior is used instead.

**To Reproduce**
1. Run a scan with a custom ignore pattern:
   ```
   opengrep scan --opengrep-ignore-pattern=mycustompattern <other-args>
   ```
   Lines with `# mycustompattern` are correctly ignored.
2. Run the same scan but add the `--experimental` flag:
   ```
   opengrep scan --opengrep-ignore-pattern=mycustompattern --experimental <other-args>
   ```
   Lines with `# mycustompattern` are **not** ignored; the flag appears to have no effect.

**Expected behavior**
The `--opengrep-ignore-pattern` flag should apply regardless of whether the `--experimental` flag is used. Custom ignore patterns should be respected in both standard and experimental modes.

**Screenshots**
N/A

**What is the priority of the bug to you?**

- [x] P0: blocking your adoption of Opengrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
Local build (please specify if you are using docker, an official binary, or other environment details if different).

**Use case**
Fixing this bug will allow us to use custom ignore patterns in all scan modes, including experimental, ensuring consistent results and workflow regardless of the scan mode.

maciejpirog:
I could not reproduce:

```
(cli) 🐪 5.3.0 ➜  opengrep (main) ✗ which opengrep                                                                                                 
/home/maciej-aikido/.local/share/virtualenvs/cli-sLD-UEk6/bin/opengrep
(cli) 🐪 5.3.0 ➜  opengrep (main) ✗ opengrep --version
1.2.2
(cli) 🐪 5.3.0 ➜  opengrep (main) ✗ cat ../test/nosem/rules/rules.yaml 
rules:

 - id: xxxxx
   languages:
     - csharp
   message: x is equal to $VAR !!!
   pattern: |
     x = $VAR;
   severity: WARNING
(cli) 🐪 5.3.0 ➜  opengrep (main) ✗ cat ../test/nosem/code/code.cs    
public class C
{
    public void MyFunction() {
      string x;
      
      x = "ok";
      
      // NOSEM
      x = "nosem";
      
      // CUSTOM
[...]

PR #260, opened 2025-05-19
fix: --opengrep-ignore-pattern declared twice in python cli

maciejpirog:
Because of this bug, the python cli issues a warning, but does not alter behaviour (in particular, it does not affect #259).


dimitris-m (approved):
Reviewed offline.

Recommended to just delete this commented out part.

Also about this weird env var: `SEMGREP_OPENGREP_IGNORE_PATTERN` that somehow escaped our initial review: we should mention in the changelog that it's gone, hopefully nobody depends on it.

PR #263, opened 2025-05-20
Fixed incremental output with custom ignore flag 

tom-paz:
Added `--opengrep-ignore-pattern` and `--incremental-output` to the list of flags that require the `--experimental` option.
This ensures users get proper error messages when using these advanced features without the experimental flag.

Fixed an issue in the match processing pipeline by adding `Core_json_output.dedup_and_sort` to the processing chain.

This ensures correct incremental output when using custom ignore patterns by properly deduplicating and sorting matches

### DIFF
--- a/src/main/Main.ml
+++ b/src/main/Main.ml
@@ -113,7 +113,7 @@ let with_experimental_flag argv =
                    = [| "opengrep"; "--experimental"; "-c"; "rules"; "libs" |]) *)
 
 let flags_that_require_experimental : string list =
-  [ "--output-enclosing-context"; "--opengrep-ignore-pattern"; "--incremental-output"; "--files-with-matches" ]
+  [ "--output-enclosing-context"; "--opengrep-ignore-pattern"; "--incremental-output" ]
 
 let experimental_flags_error_msg : string =
   "The --experimental option required for the following flags: "
--- a/src/main/Main.ml
+++ b/src/main/Main.ml
@@ -96,12 +96,16 @@ let position_for_experimental_flag argv =
 
 (* TODO[Issue #131]: Add some expectation tests for such functions. *)
 let with_experimental_flag argv =
-  let len = position_for_experimental_flag argv in
-  Array.concat [
-    Array.sub argv 0 len;
-    [| "--experimental" |];
-    Array.sub argv len (Array.length argv - len);
-  ]
+  (* Only add the flag if it's not already present *)
+  if Array.mem "--experimental" argv then
+    argv
+  else
+    let len = position_for_experimental_flag argv in
+    Array.concat [
+      Array.sub argv 0 len;
+      [| "--experimental" |];
+      Array.sub argv len (Array.length argv - len);
+    ]
 
 (* let _ = assert (with_experimental_flag [| "opengrep"; "scan"; "--help" |]
                    = [| "opengrep"; "scan"; "--experimental"; "--help" |])
@@ -109,7 +113,7 @@ let with_experimental_flag argv =
                    = [| "opengrep"; "--experimental"; "-c"; "rules"; "libs" |]) *)
 
 let flags_that_require_experimental : string list =
-  [ "--output-enclosing-context" ]
+  [ "--output-enclosing-context"; "--opengrep-ignore-pattern"; "--incremental-output"; "--files-with-matches" ]
 
 let experimental_flags_error_msg : string =
   "The --experimental option required for the following flags: "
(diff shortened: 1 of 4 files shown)


tom-paz:
#259

src/reporting/Nosemgrep.ml
```diff
@@ -301,7 +301,12 @@ let produce_ignored ?(config=Engine_config.default) (matches : Core_result.proce
   in
   (matches, List_.flatten wide_errors)
 
-let filter_ignored ~keep_ignored (matches : OutJ.core_match list) =
+let filter_ignored ~keep_ignored ?(config=Engine_config.default) (matches : OutJ.core_match list) =
+  (* Note: config is not used directly here since filter_ignored acts on 
+   * already processed matches that have the is_ignored field set.
+   * The config parameter exists for consistency with the produce_ignored function.
+   *)
+  let [@warning "-27"] _unused = config in
```
dimitris-m: If it's not used, we can leave a comment about it, but it should not be a parameter.

src/main/Main.ml
```diff
@@ -96,20 +96,24 @@ let position_for_experimental_flag argv =
...
+    ]
 
 (* let _ = assert (with_experimental_flag [| "opengrep"; "scan"; "--help" |]
                    = [| "opengrep"; "scan"; "--experimental"; "--help" |])
    let _ = assert (with_experimental_flag [| "opengrep"; "-c"; "rules"; "libs" |]
                    = [| "opengrep"; "--experimental"; "-c"; "rules"; "libs" |]) *)
 
 let flags_that_require_experimental : string list =
-  [ "--output-enclosing-context" ]
+  [ "--output-enclosing-context"; "--opengrep-ignore-pattern"; "--incremental-output" ]
```
dimitris-m, line 116: @tom-paz you previously implemented this flag (`--opengrep-ignore-pattern`) on the python side too.

We cannot just require `--experimental` when the python code to implement this has been added, and in fact does not seem to work.

- If we were to only support this with `--experimental`, all the python code would have to be reverted. (Now is too late for that.)
- But since we do have the feature in the python part, we should not require `--experimental` but rather we should also fix the python part too.

dimitris-m (changes requested):
Please see comments above.

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -493,12 +493,12 @@ let trim_core_match_fix (r : Out.core_match) =
...
   { r with extra }
 
-let adjust_nosemgrep_and_autofix ~keep_ignored ~engine_config (res : Core_runner.result) :
+let adjust_nosemgrep_and_autofix ~keep_ignored (res : Core_runner.result) :
     Core_runner.result =
   let filtered_matches =
     res.core.results
     |> List_.map trim_core_match_fix
-    |> Nosemgrep.filter_ignored ~keep_ignored ~config:engine_config
+    |> Nosemgrep.filter_ignored ~keep_ignored
```
dimitris-m: But don't you need the engine config here?

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -229,10 +229,16 @@ let mk_file_match_hook (conf : Scan_CLI.conf) (rules : Rule.rules)
...
+    let processed_matches = 
       pms
-      (* OK, because we don't need the postprocessing to report the matches. *)
       |> List_.map Core_result.mk_processed_match
+    in
+    let processed_matches, _errors =
+      Nosemgrep.produce_ignored ~config:conf.core_runner_conf.engine_config processed_matches
+    in
+    let core_matches : Out.core_match list =
+      processed_matches
```
dimitris-m, line 241: I wonder why that is needed here, given that the standard `nosem` worked -- are we sure this is the right place?

dimitris-m: In particular we already have this kind of call in `src/core_scan/Pre_post_core_scan.ml`, module `Nosemgrep_processor`.

dimitris-m: Ok, not convinced about this.
We have this in 2 places, it should be in the post-processor module.

Investigating if this a config issue, ie, is the right config available to the post-processor?

dimitris-m: In fact since with `--experimental` this already works, this change here is not necessary!

The problem is when `--experimental` is not passed!

PR #267, opened 2025-05-22
Adding ignore pattern feature with incremental-output

tom-paz:
# Add support for excluding comments with `--incremental-output`

## Summary
This PR adds support for properly handling comment-based ignore patterns when using the `--incremental-output` flag. It introduces a new `--enable-ignore` flag to control whether ignored findings should be filtered out from incremental output.

## Problem
When using `--incremental-output`, matches with `nosem` or `nosemgrep` comments (or custom patterns via `--opengrep-ignore-pattern`) were still being displayed, unlike in the regular output mode where they are properly filtered.

## Solution
- Added a new `--enable-ignore` flag to work alongside `--incremental-output`
- Updated the default values in Scan_CLI.ml
- Modified the file match hook in Scan_subcommand.ml to correctly filter out ignored matches based on the new flag
- Ensured custom ignore patterns work alongside default patterns (nosem, nosemgrep)

## Usage
```bash
opengrep scan <path> --config <rules> --incremental-output --enable-ignore --experimental
```

When `--enable-ignore` is used, matches with ignore comments are filtered out of results.
Without this flag, all matches are shown (with is_ignored flag set correctly in JSON output).

## Testing
Verified the fix with multiple test cases, ensuring:
- Default patterns (nosem, nosemgrep) work correctly
- Custom patterns with `--opengrep-ignore-pattern` work correctly
- Comments both inline and on the immediate previous line are recognized

### DIFF
--- a/src/configuring/Engine_config.ml
+++ b/src/configuring/Engine_config.ml
@@ -14,6 +14,7 @@ let default = {
 
 (* Get the list of patterns to use for ignoring lines *)
 let get_ignore_patterns config : string list =
+  let default_patterns = ["nosem"; "nosemgrep"] in
   match config.custom_ignore_pattern with
-  | None -> ["nosem"; "nosemgrep"]
-  | Some pattern -> [pattern] 
+  | None -> default_patterns
+  | Some pattern -> pattern :: default_patterns
--- a/src/configuring/Engine_config.ml
+++ b/src/configuring/Engine_config.ml
@@ -12,9 +12,11 @@ let default = {
   custom_ignore_pattern = None;
 }
 
+(* Default patterns for ignoring findings *)
+let default_ignore_patterns = ["nosem"; "nosemgrep"]
+
 (* Get the list of patterns to use for ignoring lines *)
 let get_ignore_patterns config : string list =
-  let default_patterns = ["nosem"; "nosemgrep"] in
   match config.custom_ignore_pattern with
-  | None -> default_patterns
-  | Some pattern -> pattern :: default_patterns
+  | None -> default_ignore_patterns
+  | Some pattern -> [pattern]
--- a/src/osemgrep/cli_ci/Ci_CLI.ml
+++ b/src/osemgrep/cli_ci/Ci_CLI.ml
@@ -302,6 +302,7 @@ let scan_subset_cmdline_term : Scan_CLI.conf Term.t =
         output;
         output_conf;
         incremental_output;
+        enable_ignore = false;
         engine_type;
         rewrite_rule_ids;
         matching_conf;
--- a/src/osemgrep/cli_ci/Ci_CLI.ml
+++ b/src/osemgrep/cli_ci/Ci_CLI.ml
@@ -302,7 +302,7 @@ let scan_subset_cmdline_term : Scan_CLI.conf Term.t =
         output;
         output_conf;
         incremental_output;
-        enable_ignore = false;
+        apply_ignore_pattern = false;
         engine_type;
         rewrite_rule_ids;
         matching_conf;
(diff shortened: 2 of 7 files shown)


tom-paz:
#267

src/configuring/Engine_config.ml
```diff
@@ -14,6 +14,7 @@ let default = {
 
 (* Get the list of patterns to use for ignoring lines *)
 let get_ignore_patterns config : string list =
+  let default_patterns = ["nosem"; "nosemgrep"] in
   match config.custom_ignore_pattern with
-  | None -> ["nosem"; "nosemgrep"]
-  | Some pattern -> [pattern] 
+  | None -> default_patterns
+  | Some pattern -> pattern :: default_patterns
```
dimitris-m: Here you are changing the behaviour, by adding instead of replacing patterns. 

This does not belong in this PR.

src/osemgrep/cli_ci/Ci_CLI.ml
```diff
@@ -302,6 +302,7 @@ let scan_subset_cmdline_term : Scan_CLI.conf Term.t =
         output;
         output_conf;
         incremental_output;
+        enable_ignore = false;
```
dimitris-m: This is not a good variable name, since it's not clear what "enable_ignore" means.

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -229,20 +229,30 @@ let mk_file_match_hook (conf : Scan_CLI.conf) (rules : Rule.rules)
...
+    (* Process matches to set is_ignored flag based on nosem comments *)
+    let processed_matches, _errors =
       pms
-      (* OK, because we don't need the postprocessing to report the matches. *)
       |> List_.map Core_result.mk_processed_match
+      |> Nosemgrep.produce_ignored ~config:conf.core_runner_conf.engine_config
+    in
+    (* Convert to core matches *)
+    let core_matches : Out.core_match list =
+      processed_matches
```
dimitris-m: See my comment on the PR https://github.com/opengrep/opengrep/pull/263#discussion_r2098077493, this is not the place to add this code.

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -229,20 +229,30 @@ let mk_file_match_hook (conf : Scan_CLI.conf) (rules : Rule.rules)
...
     |> List_.map
          (Cli_json_output.cli_match_of_core_match
             ~fixed_lines:conf.output_conf.fixed_lines fixed_env hrules)
-    |> List_.exclude (fun (m : Out.cli_match) -> m.extra.is_ignored ||| false)
+    in
+    (* Only filter out ignored matches if --enable-ignore is set *)
+    if conf.enable_ignore then
+      matches |> List_.exclude (fun (m : Out.cli_match) -> m.extra.is_ignored ||| false)
+    else
+      matches
```
dimitris-m: But from what I understood, what you require is to filter out matches also (in addition to normal operation) when printing incrementally. 

Here something completely different happens, no?

dimitris-m (commented):
Couple of comments: 

- Have you tested this and does it work?
- Semantics of the custom ignore pattern should not be changed in this PR which has another purpose.
- Other comments as indicated.

src/configuring/Engine_config.ml
tom-paz: fixed

src/osemgrep/cli_scan/Scan_subcommand.ml
tom-paz: Removed the duplication to a utility function

src/osemgrep/cli_ci/Ci_CLI.ml
tom-paz: changed to `apply_ignore_pattern`

src/osemgrep/cli_scan/Scan_subcommand.ml
tom-paz: the ignore filtering logic in both regular and incremental output modes, depending on the flag. The filtering is handled in the file match hook, and the logic is shared and consistent for both output types.

dimitris-m:
Superseded by #274.

Issue #272, opened 2025-05-26
Wrong ranges of parenthesized expressions in differnet languages


maciejpirog:
In some languages, the parser kills the outer parens when calculated ranges of expressions. For example a match on

```
(2 + 2) * (3 + 4)
```

is reported as 

```
2 + 2) * (3 + 4
```

This needs to be fixed on a case-by-case basis. The original issue reported for java is #23

**Affected languages that we know of:**

| Lang | Fix |
| --- | --- |
| java | #258 |
| c# | |

PR #273, opened 2025-05-26
Fix parenthesized expressions in Rust

maciejpirog:
See #272

Note: This is the third fix for the problems listed in #272. In principle, the fix is similar in all these cases, but because the parsers are structured differently, at the moment there is no one "fix the paren expr" function that we can extract and put in some "helpers" module.


dimitris-m (approved):
lgtm

Issue #287, opened 2025-06-04
Add new flag for overriding the default `.semgrepignore` file, `--semgrepignore-filename`


tom-paz:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I need to use multiple different ignore file patterns across different projects or when working with teams that have established conventions for ignore files that don't use the standard `.semgrepignore` filename. Currently, Semgrep is hardcoded to only look for `.semgrepignore` files, which makes it difficult to:
- Use existing ignore files with different naming conventions (e.g., `.semgrep-ignore`, `semgrep.ignore`)
- Maintain consistency with other tools that use different ignore file patterns
- Work with legacy projects that already have established ignore file naming
- Use multiple ignore configurations for different scanning contexts

**Describe the solution you'd like**
I would like a `--semgrepignore-filename` flag that allows users to specify a custom filename for semgrepignore files instead of the hardcoded `.semgrepignore`. This flag should:
- Accept any valid filename (e.g., `--semgrepignore-filename .custom-ignore`)
- Work with both `semgrep scan` and `semgrep ci` commands
- Maintain all existing semgrepignore functionality (pattern matching, negation, etc.)
- Be gated behind the `--experimental` flag to indicate it's a new feature
- Use the custom filename for existence checks and file reading throughout the project

**Use case**
This feature will enable several important use cases:
- **Legacy project integration**: Working with existing codebases that use non-standard ignore file names
- **Team consistency**: Maintaining consistent naming conventions across tools and teams
- **Multi-context scanning**: Using different ignore configurations for different scanning contexts (e.g., `semgrep-security.ignore`, `semgrep-quality.ignore`)
- **Tool migration**: Easier migration from other static analysis tools that use different ignore file naming patterns
- **Monorepo management**: Different subdirectories could use context-specific ignore file names

**Additional context**
This feature should be implemented similar to how Git supports custom gitignore filenames in some contexts, or how many linters support configurable ignore file patterns. The feature should:
- Require the `--experimental` flag initially to allow for user feedback
- Maintain backward compatibility (no changes when flag is not used)
- Work consistently across both CLI and CI environments
- Follow existing semgrepignore syntax and behavior exactly
- Include proper documentation and help text explaining the functionality

PR #288, opened 2025-06-04
Add new semgrepignore filename flag 

tom-paz:
#287

### DIFF
--- a/src/osemgrep/cli_scan/Scan_CLI.ml
+++ b/src/osemgrep/cli_scan/Scan_CLI.ml
@@ -1355,7 +1355,7 @@ let cmdline_term caps ~allow_empty_config : conf Term.t =
             "!!! You're using one or more options starting with '--x-'. These \
              options are not part of the opengrep API. They will change or will \
              be removed without notice !!! ");
-    
+
     (* Create engine configuration *)
     let engine_config = {
       Engine_config.custom_ignore_pattern = opengrep_ignore_pattern;
--- a/src/osemgrep/cli_scan/Scan_CLI.ml
+++ b/src/osemgrep/cli_scan/Scan_CLI.ml
@@ -264,11 +264,10 @@ CHANGE OR DISAPPEAR WITHOUT NOTICE.
 
 let o_semgrepignore_filename : string option Term.t =
   let info =
-    Arg.info
+    Arg.info ~docv:"FILENAME"
       [ "semgrepignore-filename" ]
       ~doc:
-        {|Override the default '.semgrepignore' with the specified file path.
-REQUIRES --experimental|}
+        {|Use the file $(docv) instead of the default .semgrepignore to specify targets skipped during the scan. REQUIRES --experimental|}
   in
   Arg.value (Arg.opt Arg.(some string) None info)
 
--- a/src/osemgrep/cli_scan/Scan_CLI.ml
+++ b/src/osemgrep/cli_scan/Scan_CLI.ml
@@ -267,8 +267,7 @@ let o_semgrepignore_filename : string option Term.t =
     Arg.info
       [ "semgrepignore-filename" ]
       ~doc:
-        {|Override the default ignore file name instead of '.semgrepignore'.
-This file will be treated like the .semgrepignore just with a custom path.
+        {|Override the default '.semgrepignore' with the specified file path.
 REQUIRES --experimental|}
   in
   Arg.value (Arg.opt Arg.(some string) None info)
--- a/src/osemgrep/cli_scan/Scan_CLI.ml
+++ b/src/osemgrep/cli_scan/Scan_CLI.ml
@@ -262,6 +262,17 @@ CHANGE OR DISAPPEAR WITHOUT NOTICE.
   in
   Arg.value (Arg.flag info)
 
+let o_semgrepignore_filename : string option Term.t =
+  let info =
+    Arg.info
+      [ "semgrepignore-filename" ]
+      ~doc:
+        {|Override the default ignore file name instead of '.semgrepignore'.
+This file will be treated like the .semgrepignore just with a custom path.
+REQUIRES --experimental|}
+  in
+  Arg.value (Arg.opt Arg.(some string) None info)
+
 let o_scan_unknown_extensions : bool Term.t =
   let default = default.targeting_conf.always_select_explicit_targets in
   H.negatable_flag
@@ -1331,7 +1342,7 @@ let cmdline_term caps ~allow_empty_config : conf Term.t =
       metrics num_jobs no_secrets_validation nosem opengrep_ignore_pattern optimizations oss output output_enclosing_context
       pattern pro project_root pro_intrafile pro_lang pro_path_sensitive remote
       replacement rewrite_rule_ids sarif sarif_outputs scan_unknown_extensions
-      secrets severity show_supported_languages strict target_roots test
+      secrets semgrepignore_filename severity show_supported_languages strict target_roots test
       test_ignore_todo text text_outputs time_flag timeout
       _timeout_interfileTODO timeout_threshold (*  trace trace_endpoint *) use_git
       validate version version_check vim vim_outputs
@@ -1453,6 +1464,7 @@ let cmdline_term caps ~allow_empty_config : conf Term.t =
         explicit_targets;
         respect_gitignore;
         respect_semgrepignore_files = not x_ignore_semgrepignore_files;
+        semgrepignore_filename;
         exclude_minified_files;
       }
     in
@@ -1570,7 +1582,7 @@ let cmdline_term caps ~allow_empty_config : conf Term.t =
     $ o_output $ o_output_enclosing_context $ o_pattern $ o_pro $ o_project_root $ o_pro_intrafile
     $ o_pro_languages $ o_pro_path_sensitive $ o_remote $ o_replacement
     $ o_rewrite_rule_ids $ o_sarif $ o_sarif_outputs $ o_scan_unknown_extensions
-    $ o_secrets $ o_severity $ o_show_supported_languages $ o_strict
+    $ o_secrets $ o_semgrepignore_filename $ o_severity $ o_show_supported_languages $ o_strict
     $ o_target_roots $ o_test $ Test_CLI.o_test_ignore_todo $ o_text
     $ o_text_outputs $ o_time $ o_timeout $ o_timeout_interfile
     $ o_timeout_threshold $ (* o_trace $ o_trace_endpoint $ *) o_use_git $ o_validate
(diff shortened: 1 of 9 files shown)


src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -262,6 +262,17 @@ CHANGE OR DISAPPEAR WITHOUT NOTICE.
   in
   Arg.value (Arg.flag info)
 
+let o_semgrepignore_filename : string option Term.t =
+  let info =
+    Arg.info
+      [ "semgrepignore-filename" ]
+      ~doc:
+        {|Override the default ignore file name instead of '.semgrepignore'.
+This file will be treated like the .semgrepignore just with a custom path.
```
dimitris-m: I suggest you remove this sentence "This file ....".

By the way this sentence implies you want to change the path, not just the filename. (See other comments.)

src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -262,6 +262,17 @@ CHANGE OR DISAPPEAR WITHOUT NOTICE.
   in
   Arg.value (Arg.flag info)
 
+let o_semgrepignore_filename : string option Term.t =
+  let info =
+    Arg.info
+      [ "semgrepignore-filename" ]
+      ~doc:
+        {|Override the default ignore file name instead of '.semgrepignore'.
```
dimitris-m: _Override the default '.semgrepignore' with the specified file path._

src/targeting/Find_targets.ml
```diff
@@ -206,6 +206,8 @@ type conf = {
   max_target_bytes : int;
   respect_gitignore : bool;
   respect_semgrepignore_files : bool;
+  (* Custom semgrepignore filename to use instead of '.semgrepignore' *)
```
dimitris-m: Pls remove this comment. We don't have comments in the other parameters and it's clear anyway.

src/targeting/Semgrepignore.ml
```diff
@@ -106,7 +106,7 @@ let contents_of_builtin_semgrepignore = function
   | Empty -> ""
   | Semgrep_scan_legacy -> default_semgrepignore_for_semgrep_scan
 
-let create ?(cli_patterns = []) ~default_semgrepignore_patterns
+let create ?(cli_patterns = []) ?(semgrepignore_filename = None) ~default_semgrepignore_patterns
```
dimitris-m: This new parameter should be:

```ocaml
?(semgrepignore_filename = default_semgrepignore_filename)
```

where `default_semgrepignore_filename` should be defined near the top of the file, to be equal to `".semgrepignore"`.

As a side-note, if you wanted `None` for missing, you could just write `?semgrepignore_filename`.

src/targeting/Semgrepignore.ml
```diff
@@ -134,6 +134,16 @@ let create ?(cli_patterns = []) ~default_semgrepignore_patterns
...
   in
+  let semgrepignore_files =
+    match semgrepignore_filename with
+    | Some custom_filename ->
+        {
+          Gitignore.source_kind = "semgrepignore";
+          filename = custom_filename;
+          format = Gitignore.Legacy_semgrepignore;
+        }
+    | None -> semgrepignore_files
```
dimitris-m: This should be:

```ocaml
let semgrepignore_files = semgrepignore_filenames ~filename:semgrepignore_filename
```

and the definition of `semgrepignore_filenames` should be turned into a function that takes a filename (named) parameter.

I'm asking this change to avoid code duplication (you are copy pasting the definition) which is error prone.

So on line 98 you should have something like this definition: 

```ocaml
let semgrepignore_files ~filename : Gitignore.gitignore_filename =
  {
    source_kind = "semgrepignore";
    filename;
    format = Gitignore.Legacy_semgrepignore;
  }
```

src/targeting/Semgrepignore.mli
```diff
@@ -33,6 +33,7 @@ type exclusion_mechanism = {
 *)
 val create :
   ?cli_patterns:string list ->
+  ?semgrepignore_filename:string option ->
```
dimitris-m: the type should be `string`.

after you change it the compiler will help you adapt other places as required.

src/targeting/Semgrepignore.ml
```diff
@@ -150,7 +160,9 @@ let create ?(cli_patterns = []) ~default_semgrepignore_patterns
   *)
   let root_semgrepignore_exists =
     let root_dir = Ppath.to_fpath ~root:project_root Ppath.root in
-    let semgrepignore_path = Fpath.add_seg root_dir ".semgrepignore" in
+    let semgrepignore_path = Fpath.add_seg root_dir (match semgrepignore_filename with
+      | Some custom_filename -> custom_filename
+      | None -> ".semgrepignore") in
```
dimitris-m:
```ocaml
let semgrepignore_path = Fpath.add_seg root_dir semgrepignore_filename in 
...
```

We should not let the logic of choosing the filename spill all over the place.

src/targeting/Find_targets.ml
```diff
@@ -732,7 +735,9 @@ let setup_path_filters conf (project_roots : Project.roots) :
    *)
   let semgrepignore_filter =
     Semgrepignore.create ~cli_patterns:conf.exclude
-      ~default_semgrepignore_patterns:Semgrep_scan_legacy ~exclusion_mechanism
+      ~semgrepignore_filename:conf.semgrepignore_filename
```
dimitris-m:
```ocaml
?semgrepignore_filename:conf.semgrepignore_filename
```

src/targeting/Semgrepignore.ml
dimitris-m: By the way, what seems to happen here is that the _name_ of the file can change, but not the location, so I'm not sure that it does what the issue suggests.

dimitris-m (changes requested):
Left some suggested changes.

I'm not sure this does what #287 implies. 
Is the objective to override the name _or_ is it to override the complete _path_ to the file?

By the way, if you want the full path to be custom, we must rename the new parameter to avoid confusion with Semgrep's parameter which only changes the name.

Please could you add some tests in `src/targeting/Unit_semgrepignore.ml`, this will also clarify what the new parameter is supposed to do.

src/osemgrep/cli_scan/Scan_CLI.ml
dimitris-m: Update: I thought you want to update the path, not sure now.

src/targeting/Find_targets.ml
```diff
@@ -731,8 +733,11 @@ let setup_path_filters conf (project_roots : Project.roots) :
    * the .gitignore of all the parents of the scan_root.
    *)
   let semgrepignore_filter =
+    let semgrepignore_filename = Option.value conf.semgrepignore_filename ~default:".semgrepignore" in
```
dimitris-m: You already defined, in Semgrepignore.ml: 

```ocaml
let default_semgrepignore_filename = ".semgrepignore"
```

So the default value is defined in one place, it should not also be defined here, it's error prone.

You can just pass `?semgrepignore_filename:conf.semgrepignore_filename` in `Semgrepignore.create` below, instead of `~semgrepignore_filename`. (The above `let` should be removed).

src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -262,6 +262,16 @@ CHANGE OR DISAPPEAR WITHOUT NOTICE.
   in
   Arg.value (Arg.flag info)
 
+let o_semgrepignore_filename : string option Term.t =
+  let info =
+    Arg.info
+      [ "semgrepignore-filename" ]
+      ~doc:
+        {|Override the default '.semgrepignore' with the specified file path.
+REQUIRES --experimental|}
```
dimitris-m: This is a bit ambiguous: _Override the default '.semgrepignore' with the specified file path_.

Because by _path_ users will assume that it's an absolute path, and also the parameter itself implies that it's only a _filename_.

@maciejpirog what do you think? It seems the user can pass a relative path here..

maciejpirog: Also, the word "override" is very confusing here (are we overriding one file with a different file?). A more clear message would be sth like

```
--semgrepignore-filename=FILENAME
 Use the file FILENAME instead of the default .semgrepignore to specify targets skipped during the scan.
```

dimitris-m: I think, looking at Semgrep's feature, that we can do what @maciejpirog suggests with something like: 

```ocaml
let o_semgrepignore_filename : string option Term.t =
  let info =
    Arg.info ~docv:"FILENAME"
      [ "semgrepignore-filename" ]
      ~doc:
        {|Use the file $(docv) instead of the default .semgrepignore to specify targets skipped during the scan. REQUIRES --experimental|}
  in
  Arg.value (Arg.opt Arg.(some string) None info)
```

src/targeting/Unit_semgrepignore.ml
```diff
@@ -70,6 +70,54 @@ let test_filter ?excludes:cli_patterns (files : F.t list) selection () =
 (* The tests *)
 (*****************************************************************************)
 
+gi(*
+   Test helper for custom semgrepignore filename
+*)
```
dimitris-m: this `gi` seems like a misprint?

dimitris-m:
Once we address the comments from today, we can merge this in.

src/targeting/Semgrepignore.ml
```diff
@@ -150,7 +154,7 @@ let create ?(cli_patterns = []) ~default_semgrepignore_patterns
   *)
   let root_semgrepignore_exists =
     let root_dir = Ppath.to_fpath ~root:project_root Ppath.root in
-    let semgrepignore_path = Fpath.add_seg root_dir ".semgrepignore" in
+    let semgrepignore_path = Fpath.add_seg root_dir semgrepignore_filename in
```
dimitris-m, line 157: You should remove the tests where a path is passed, since this will not work.

Try this in utop: 

```ocaml
Fpath.add_seg (Fpath.v ".") "hello.txt";;
```

The above works. Now try: 

```ocaml
Fpath.add_seg (Fpath.v ".") "build/hello.txt" |> Fpath.to_string;;
```

The above fails.

So in fact only a filename can be passed.

I'm fine with this, in fact I prefer it. But if you want to propose a more flexible feature, let me know.

dimitris-m: This is why the tests fail -- it's important to run the tests locally before making a review request. 
You can use `make core-test` for this.

dimitris-m (approved):
I'll merge this now

Issue #291, opened 2025-06-10
Opengrep fails to scan on Windows due to case sensitivity in path comparison


tom-paz:
**Describe the bug**

Scan fails on Windows due to case sensitivity issues in path comparison. The scanner cannot generate relative paths when the current working directory and project root have different case in the drive letter (e.g., `c:\` vs `C:\`), causing the scan to fail with the error "cannot make path '.' relative to project root".

**To Reproduce**

1. Run Opengrep on a Windows system where the current working directory is returned with a lowercase drive letter (e.g., `c:\Users\User\Downloads\proj`)
2. The project root is detected with an uppercase drive letter (e.g., `C:\Users\User\Downloads\proj`)
3. Execute any scan command, for example:
   ```
   opengrep-cli scan --json --experimental --disable-version-check --config <config-path>
   ```

**Expected behavior**
The scan should proceed normally, recognizing that paths with different case are equivalent on Windows file systems (which are case-insensitive by default).

**Screenshots**
Error message:

```
[01.81][ERROR]: Error: cannot make path "." relative to project root "C:\\Users\\User\\Downloads\\proj".
cwd: c:\Users\User\Downloads\proj
realpath for .: { Rfpath.fpath = .;
rpath = (Rpath.Rpath C:\Users\User\Downloads\proj);
cwd = (Rpath.Rpath C:\Users\User\Downloads\proj) }
```

**What is the priority of the bug to you?**

- [x] P0: blocking your adoption of Opengrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
Windows system running official Opengrep binary

**Use case**
Fixing this bug will enable Windows users to run Opengrep scans without encountering path resolution failures due to case sensitivity mismatches. This is essential for Windows adoption since the current behavior makes Opengrep unusable on Windows systems where different system calls return paths with inconsistent case.

**Technical Details**
The issue occurs in the `Ppath.ml` module where path comparison functions (`Fpath.equal`, `Fpath.rem_prefix`) perform case-sensitive string comparisons even on Windows where file systems are case-insensitive. The bug manifests when:

- `Unix.getcwd()` returns a path with one case (e.g., uppercase drive letter)
- `Sys.getcwd()` returns the same path with different case (e.g., lowercase drive letter)
- Path prefix removal fails because `"C:\path"` ≠ `"c:\path"` in string comparison

**Proposed Solution**
Implement case-insensitive path comparison on Windows by:
[...]

PR #292, opened 2025-06-10
Fixing windows relative path issues

tom-paz:
#291

### DIFF
--- a/libs/glob/Match.ml
+++ b/libs/glob/Match.ml
@@ -131,12 +131,10 @@ let translate_root pat =
 (*****************************************************************************)
 
 (* Compile a pattern into an ocaml-re regexp for fast matching *)
-(* TODO: For windows, we should probably make lowecase since it's not
- * case sensitive. Then paths should also be transformed in the same
- * way. *)
 let compile ~source pat =
   let pcre = translate_root pat in
-  let re = Pcre2_.regexp pcre in
+  let flags = if Sys.os_type = "Win32" then [`CASELESS] else [] in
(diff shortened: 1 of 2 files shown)


libs/glob/Match.ml
```diff
@@ -131,12 +131,10 @@ let translate_root pat =
 (*****************************************************************************)
 
 (* Compile a pattern into an ocaml-re regexp for fast matching *)
-(* TODO: For windows, we should probably make lowecase since it's not
- * case sensitive. Then paths should also be transformed in the same
- * way. *)
 let compile ~source pat =
   let pcre = translate_root pat in
-  let re = Pcre2_.regexp pcre in
+  let flags = if Sys.os_type = "Win32" then [`CASELESS] else [] in
```
dimitris-m: Minor: `if Sys.win32` also works.

libs/paths/Ppath.ml
```diff
@@ -324,7 +357,7 @@ let in_project_unsafe_for_tests ~(phys_root : Fpath.t) (path : Fpath.t) =
            "cannot make path %S relative to project root %S.\n\
             cwd: %s\n\
             realpath for .: %s\n\
-            Sys.argv: %s" !!path !!phys_root (Sys.getcwd ())
+            Sys.argv: %s" !!path !!phys_root (Unix.getcwd ())
```
dimitris-m, line 361: I tested on a windows machine and both `Sys.getcwd ()` and `Unix.getcwd ()` return the same result (uppercase drive letter).

Issue #82, opened 2025-02-13
Binary install script


dimitris-m:
TL;DR: allow users to execute something like:

-  `curl -sfL https://raw.githubusercontent.com/opengrep/opengrep/main/install.sh | sh` on Linux, Mac 
- `Invoke-Expression "& { $(Invoke-RestMethod https://raw.githubusercontent.com/opengrep/opengrep/main/install.ps1) }"` on Windows

Resulting in the installation of the binary in the user's local bin directory, with the correct permissions

See https://www.reddit.com/r/opengrep/comments/1insz57/binary_install_script/.

deepakvijayaraj-official:
Hello @dimitris-m 

I am happy to work on this and help to get the script created.

At the moment, I have forked and created a script for this here in my repo.
https://github.com/deepakvijayaraj-official/opengrep/blob/main/install.sh

I have also updated the release workflow to add "latest" tag so that the script can work automatically.
https://github.com/deepakvijayaraj-official/opengrep/blob/rollingreleaseupdate-82/.github/workflows/rolling-release.yml

- name: Create or Update Rolling Release
         ..........
          draft: false  # Change if you want automatic publishing
          prerelease: false  # Change based on your release strategy
          make_latest: true
          # Make this conditional:
          make_latest: ${{ github.ref == 'refs/heads/main' }}  # Only make latest if from main branch
          body: |
            This is a rolling release from `main`.
            Note that the Windows version is not yet functional, but can be tested if the following parameters are passed to the `scan` command in addition to any other parameters: `-j 1 --timeout 0`.
         ..........

dimitris-m:
Hi, thanks for that. 

Why not open a draft PR from your branch?

I don't think we need to change the release workflow to include 'latest' in order for the script to work automatically, the script could: 

- get the actual latest version if no parameter is given, by parsing the available versions; 
- else it could take a parameter that would fetch a specific version, which is probably desirable for reproducibility.

It's a bit more work, but it's probably easy? 
I'm not suggesting to use anything like GH cli though, we should just use what can be reasonably expected to be installed on the user's system.

Bonus: 
The script could have an option to list available versions.

deepakvijayaraj-official:
Hi @dimitris-m ,
Thanks for your reply and the suggestions. 
I can definitely make it work something on those lines.
At the moment I dont seem to have access to the repository. Can you please help me get access to create a branch and raise a PR.
Thanks

dimitris-m:
[...]

PR #294, opened 2025-06-12
Binary install script for Mac OSX and Linux

Gable-github:
### Description:
Add a install shell script for easy binary installation of Opengrep.

### Includes:
- install.sh: Downloads and installs the correct prebuilt binary for Linux and macOS (x86_64 and arm64/aarch64), with automatic detection.

- Versioned installs in ~/.opengrep/cli/$VERSION and a symlink to latest.

- Attempts to symlink to ~/.local/bin/opengrep for user-wide CLI access (falls back to a hint for the user to do it manually)

- Handles argument parsing:

    `-v` install specific version
    
    `-l` list available versions
    
    `-h` help/usage

### Why?
Simplifies Opengrep setup for end users (no manual downloads or system-wide installs required), no need for sudo permissions for script to run successfully.

### Docs:
Updated README and install instructions for end users.

### Issue:
Resolves Enchancement Issue #82

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,3 +1,23 @@
+# Installing Opengrep (for users)
+
+Install the Opengrep CLI is with our install script:
+
+```sh
+curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install_opengrep.sh | bash
+```
+
+- This will install Opengrep to `~/.opengrep/cli/<version>` and set up a `latest` symlink.
+- To install a specific version, use:
+  ```sh
+  curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install_opengrep.sh | bash -s -- -v v1.0.0
+  ```
+- To list available versions:
+  ```sh
+  curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install_opengrep.sh | bash -s -- -l
+  ```
+
+---
+
 # Build instructions for developers
 
 ## Manual development
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,19 +1,17 @@
-# Installing Opengrep (for users)
-
-Install the Opengrep CLI is with our install script:
+# Install the Opengrep CLI with our install script:
 
 ```sh
-curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install_opengrep.sh | bash
+curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install.sh | bash
 ```
 
 - This will install Opengrep to `~/.opengrep/cli/<version>` and set up a `latest` symlink.
 - To install a specific version, use:
   ```sh
-  curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install_opengrep.sh | bash -s -- -v v1.0.0
+  curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install.sh | bash -s -- -v v1.0.0
   ```
 - To list available versions:
   ```sh
-  curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install_opengrep.sh | bash -s -- -l
+  curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install.sh | bash -s -- -l
   ```
 
 ---
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Apex · Bash · C · C++ · C# · Clojure · Dart · Dockerfile · Elixir · HTM
 
 ## Installation
 
-You can install Opengrep via our official install script.
+You can install Opengrep using our official install script.
 
 ### Quick Install (Recommended)
 ```sh
@@ -45,7 +45,7 @@ If you've cloned the repo and `install.sh` is in the root directory, you can run
 Which will install the latest version of Opengrep.
 
 You can also install manually:
-* Binaries available in the latest alpha [release](https://github.com/opengrep/opengrep/releases).
+* Binaries available in the [release page](https://github.com/opengrep/opengrep/releases).
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -29,8 +29,26 @@ Apex · Bash · C · C++ · C# · Clojure · Dart · Dockerfile · Elixir · HTM
 
 ## Installation
 
-Binaries available in the latest alpha [release](https://github.com/opengrep/opengrep/releases).
- 
+You can install Opengrep via our official install script.
+
+### Quick Install (Recommended)
+```sh
+curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/main/install.sh | bash
+```
+
+If you've cloned the repo and `install.sh` is in the root directory, you can run:
+
+```sh
+./install.sh
+```
+
+Which will install the latest version of Opengrep.
+
+You can also install manually:
+* Binaries available in the latest alpha [release](https://github.com/opengrep/opengrep/releases).
+
+---
+
 ## Getting started
 
 Create `rules/demo-rust-unwrap.yaml` with the following content:
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,191 @@
+#!/bin/bash -e
+# Opengrep installation script 
+
+print_usage() {
+    echo "Usage: $0 [-v version] [-l] [-h]"
+    echo "  -v    Specify version to install (optional, default: latest)"
+    echo "  -l    List available versions (latest 3)"
+    echo "  -h    Show this help message"
+}
+
+# Function to get available versions - already checked when running main
+get_available_versions() {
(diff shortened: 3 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,191 @@
+#!/bin/bash -e
```
JonZeolla:
```suggestion
#!/usr/bin/env bash -e
```

install.sh
```diff
@@ -0,0 +1,191 @@
...
+# Function to get available versions - already checked when running main
+get_available_versions() {
+    command -v curl >/dev/null 2>&1 || { echo >&2 "Required tool curl could not be found. Aborting."; exit 1; }
+    curl -s https://api.github.com/repos/opengrep/opengrep/releases | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
+}
+
+# Function to get latest release version
+get_latest_version() {
+    command -v curl >/dev/null 2>&1 || { echo >&2 "Required tool curl could not be found. Aborting."; exit 1; }
+    curl -s https://api.github.com/repos/opengrep/opengrep/releases | grep '"tag_name":' | head -1 | sed -E 's/.*"([^"]+)".*/\1/'
```
JonZeolla: This uses the GitHub API to [identify what the latest release is](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release) and get that. However it expects that jq is installed, so maybe using this with grep is more preferred? It also ensures that the user wouldn't be given pre-release or draft releases
```suggestion
    curl -s https://api.github.com/repos/opengrep/opengrep/releases/latest | jq -r '.tag_name'
```

JonZeolla: I confirmed this works:
```suggestion
    curl -s https://api.github.com/repos/opengrep/opengrep/releases/latest | grep '"tag_name":' | head -1 | sed -E 's/.*"([^"]+)".*/\1/'
```

README.md
```diff
@@ -29,8 +29,26 @@ Apex · Bash · C · C++ · C# · Clojure · Dart · Dockerfile · Elixir · HTM
...
+If you've cloned the repo and `install.sh` is in the root directory, you can run:
+
+```sh
+./install.sh
+```
+
+Which will install the latest version of Opengrep.
+
+You can also install manually:
+* Binaries available in the latest alpha [release](https://github.com/opengrep/opengrep/releases).
```
dimitris-m: The releases are not alpha any more.

dimitris-m: "Binaries available in the release page:"

INSTALL.md
```diff
@@ -1,3 +1,23 @@
+# Installing Opengrep (for users)
+
+Install the Opengrep CLI is with our install script:
```
dimitris-m: "Install the Opengrep CLI with our install script:"

README.md
```diff
@@ -29,8 +29,26 @@ Apex · Bash · C · C++ · C# · Clojure · Dart · Dockerfile · Elixir · HTM
 
 ## Installation
 
-Binaries available in the latest alpha [release](https://github.com/opengrep/opengrep/releases).
- 
+You can install Opengrep via our official install script.
```
dimitris-m: "You can install Opengrep using our official install script."

install.sh
```diff
@@ -0,0 +1,191 @@
...
+# Function to get available versions - already checked when running main
+get_available_versions() {
+    command -v curl >/dev/null 2>&1 || { echo >&2 "Required tool curl could not be found. Aborting."; exit 1; }
+    curl -s https://api.github.com/repos/opengrep/opengrep/releases | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
+}
+
+# Function to get latest release version
+get_latest_version() {
+    command -v curl >/dev/null 2>&1 || { echo >&2 "Required tool curl could not be found. Aborting."; exit 1; }
+    curl -s https://api.github.com/repos/opengrep/opengrep/releases/latest | grep '"tag_name":' | head -1 | sed -E 's/.*"([^"]+)".*/\1/'
```
dimitris-m: there's some code duplication here, why not use `get_available_versions` then take `head -1`?

install.sh
```diff
@@ -0,0 +1,191 @@
...
+    fi
+
+    if [ -z "${DIST}" ]; then
+        echo "Operating system '${OS}' / architecture '${ARCH}' is unsupported." 1>&2
+        exit 1
+    fi
+
+    URL="https://github.com/opengrep/opengrep/releases/download/${VERSION}/${DIST}"
+    echo
+    echo "*** Installing Opengrep version ${VERSION} for ${DIST} ***"
```
```
```

install.sh
```diff
@@ -0,0 +1,191 @@
...
+        if [ -z "$TEST" ]; then
+            echo "Failed to execute installed binary: ${INST}/opengrep." 1>&2
+            exit 1
+        fi
+
+        echo
+        echo "Successfully installed Opengrep binary to ${INST}/opengrep"
+        
+        ln -s "${INST}" "${LATEST}" || exit 1
+        echo "with a symlink from ${LATEST}/opengrep"
```
dimitris-m, line 111:
```
echo "Created symlink to ${LATEST}/opengrep"
```

install.sh
```diff
@@ -0,0 +1,191 @@
...
+    fi
+
+    LOCALBIN="${HOME}/.local/bin"
+    SYMLINK_CREATED=0
+
+    if [ -d "${LOCALBIN}" ] && [ -w "${LOCALBIN}" ] && [ ! -f "${LOCALBIN}/opengrep" ]; then
+        ln -s "${LATEST}/opengrep" "${LOCALBIN}/opengrep" && SYMLINK_CREATED=1
+        if [ $SYMLINK_CREATED -eq 1 ]; then
+            echo "Created a symlink: ${LOCALBIN}/opengrep → ${LATEST}/opengrep"
```
```
echo "Created symlink from ${LATEST}/opengrep to ${LOCALBIN}/opengrep"
```

install.sh
```diff
@@ -0,0 +1,191 @@
...
+        TEST=$("${INST}/opengrep" --version 2>/dev/null || true)
+        if [ -z "$TEST" ]; then
+            echo "Failed to execute installed binary: ${INST}/opengrep." 1>&2
+            exit 1
+        fi
+
+        echo
+        echo "Successfully installed Opengrep binary to ${INST}/opengrep"
+        
+        ln -s "${INST}" "${LATEST}" || exit 1
```
dimitris-m, line 110: This will fail on the second installation attempt, right? 

Please can you test and confirm that the script works with > 1 installations one after the other?
For example to install 2 different versions.

dimitris-m: You might need:

```
rm -f "${LATEST}" || exit 1
```

before creating the new symlink.

install.sh
```diff
@@ -0,0 +1,191 @@
...
+
+    if [ $SYMLINK_CREATED -eq 1 ]; then
+        echo
+        echo "To launch Opengrep now, type"
+        echo "opengrep"
+    else
+        echo
+        echo "To launch Opengrep now, type"
+        echo "${LATEST}/opengrep"
+    fi
```
dimitris-m: When can this happen?

dimitris-m (changes requested):
Left some comments, the most important is to ensure that installation 2 times in a row succeeds.

Also, for merging I will squash the commits to maintain a more or less linear looking history, once you address any issues.

install.sh
Gable-github: hi, thanks for the detailed review. I'll be updating the script such that the previous symlink to latest is (1) overriden and (2) only if the most recently installed version is a higher version number.

so latest points to the latest version released by the team instead of the most recent binary installed by the user. 

Or do you think i should follow the duckdb example that you mentioned before as attached?
image

Gable-github: i'll commit it in duckdb way first, as ive tested that and it works.

install.sh
Gable-github: Hi, the flawed logic went over my head. made the changes, it should work now. Thanks!!

Gable-github:
Hi @dimitris-m, thank you for the detailed review. I made the requested changes and cleaned up the logic. 

Note: 
- `latest` symlink at `/.local/bin` will point to the `latest` symlink in `/.opengrep/cli`
- `latest` symlink in `/.opengrep/cli` will update to point to the most recently installed version of opengrep binary by the user, following [the duckdb install script](https://install.duckdb.org/)
- cleaned up the logic for `echo Hint...` for user to manually update the PATH variable. This logic only fires off when the script dosent have permission to do it on their behalf. It was firing off erroneously in the previous commit because of this line `[ ! -f "${LOCALBIN}/opengrep" ]`. I have nested this logic.

install.sh
dimitris-m: I think duckdb way.

The default executable in the path should be the latest installed version.

Gable-github: Okay, it is implemented. Thanks for the guidance!

dimitris-m (approved):
Looks good, thanks! 

Merging now.

PR #296, opened 2025-06-12
Fix #295: PHP interpolated strings parsed as normal strings

### LINKED ISSUE #295: Some PHP interpolated strings parsed as normal strings when tree-sitter is used
**Describe the bug**
PHP tree-sitter parser does not parse correctly some interpolated strings.

**To Reproduce**

Try parsing this code (`core-repr.php`): 

```php
<?php

function f($arg) {
    return exec("hello $arg") ?? throw new RuntimeException('hello');
}
```

Because of `throw`, it falls back to the Tree-Sitter parser.

Inspect the parse tree with:

```bash
❯ opengrep-core -lang php -dump_ast code-repr.php
[00.09][INFO]: Executed as: opengrep-core -lang php -dump_ast code-repr.php
[00.09][INFO]: Version: 1.4.0
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("f", ()),
                {id_info_id=6; id_flags=Ref(2);
                 id_resolved_alternative=Ref([]); id_resolved=Ref(None);
                 id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[]; tparams=None; },
      FuncDef(
        {fkind=(Function, ());
         fparams=[Param(
                    {pname=Some(("$arg", ())); pdefault=None; ptype=None;
                     pattrs=[];
                     pinfo={id_info_id=2; id_flags=Ref(0);
                            id_resolved_alternative=Ref([]);
                            id_resolved=Ref(None); id_type=Ref(None);
                            id_svalue=Ref(None); };
                     })];
         frettype=None;
         fbody=FBStmt(
                 Block(
                   [Return((),
                      Some(Call(IdSpecial((Op(Nullish), ())),
                             [Arg(
                                Call(
                                  N(
                                    Id(("exec", ()),
                                      {id_info_id=3; id_flags=Ref(2);
                                       id_resolved_alternative=Ref([]);
                                       id_resolved=Ref(None);
                                       id_type=Ref(None);
                                       id_svalue=Ref(None); })),
                                  [Arg(L(String(("hello $arg", ()))))]));
                              Arg(
                                StmtExpr(
                                  Throw((),
                                    New((),
                                      {t_attrs=[];
                                       t=TyExpr(
                                           N(
                                             Id(("RuntimeException", ()),
                                               {id_info_id=4;
                                                id_flags=Ref(2);
                                                id_resolved_alternative=Ref(
                                                []); id_resolved=Ref(
                                                None); id_type=Ref(None);
                                                id_svalue=Ref(None); })));
                                       },
                                      {id_info_id=5; id_flags=Ref(0);
                                       id_resolved_alternative=Ref([]);
                                       id_resolved=Ref(None);
                                       id_type=Ref(None);
                                       id_svalue=Ref(None); },
                                      [Arg(L(String(("hello", ()))))]), ())))])),
                      ())]));
         })))])
```

Remove the throw and try again (`code-repr-no-throw.php`):

```php
<?php

function f($arg) {
    return exec("hello $arg");
}
```

```bash
❯ opengrep-core -lang php -dump_ast code-repr-no-throw.php
[00.09][INFO]: Executed as: opengrep-core -lang php -dump_ast code-repr-no-throw.php
[00.09][INFO]: Version: 1.4.0
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("f", ()),
                {id_info_id=5; id_flags=Ref(2);
                 id_resolved_alternative=Ref([]); id_resolved=Ref(None);
                 id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[]; tparams=None; },
      FuncDef(
        {fkind=(Function, ());
         fparams=[Param(
                    {pname=Some(("$arg", ())); pdefault=None; ptype=None;
                     pattrs=[];
                     pinfo={id_info_id=2; id_flags=Ref(0);
                            id_resolved_alternative=Ref([]);
                            id_resolved=Ref(None); id_type=Ref(None);
                            id_svalue=Ref(None); };
                     })];
         frettype=None;
         fbody=FBStmt(
                 Block(
                   [Return((),
                      Some(Call(
                             N(
                               Id(("exec", ()),
                                 {id_info_id=3; id_flags=Ref(2);
                                  id_resolved_alternative=Ref([]);
                                  id_resolved=Ref(None); id_type=Ref(
                                  None); id_svalue=Ref(None); })),
                             [Arg(
                                Call(
                                  IdSpecial(
                                    (ConcatString(InterpolatedConcat), ())),
                                  [Arg(L(String(("hello ", ()))));
                                   Arg(
                                     N(
                                       Id(("$arg", ()),
                                         {id_info_id=4; id_flags=Ref(
                                          0);
                                          id_resolved_alternative=Ref(
                                          []); id_resolved=Ref(None);
                                          id_type=Ref(None);
                                          id_svalue=Ref(None); })))]))])), ())]));
         })))])
```

The interpolated string is parsed correctly.

**Recommended solution**

- When Tree-Sitter is used, the conversion to generic AST should be corrected.
- Maybe update the Menhir parser to allow `throw` which apparently is a more recent PHP feature (compared to what the parser defines).

- [X] P1: important to fix or quite annoying

**Use case**

Tainting is affected obviously...

corneliuhoffman:
close #295 
We had the following issues:
1. The default parsed Menhir had not been updated for PHP 8.0 and so it fails to parse the following syntax
```php
return shell_exec($command)
            ?? throw new RuntimeException('Failed to get changed files');
 ```
 as a consequence the treesitter parser kick in. However we arrive at:
 2. there is a bug in the treesitter parser  in which the syntax
 ```php
 "convert $filename"
 ```
 is parsed literally as a string.
 
 So far I did
 1. introduced another token `??` in the menhir parser and updated the transformation as needed.
 2. changed a little the parsing of `throw` to make it an expression to be used in the ??
 3. manually  tested that the rules now find all the shell_exec and also they will find a ?? pattern
 
 TODO
 1. write an actual test for the change.
 2. continue the improvements to the parser so that it fits with:
 
Screenshot 2025-06-12 at 17 58 54

### CODE REVIEW

**Comment by corneliuhoffman** [comment]
issues remainig for the parser:
1. static return type:

```php
function f(): static {}
```
2. Match expressions:
```php
match($x) {
    1 => 'one',
    2 => 'two',
};
```
3. Enums:
```php
enum Status {
  case Active;
  case Inactive;
}
```
4. Null coallescing assignment
```php
$a ??= 'default';
```
5. Array spread:
```php
$parts = ['apple', 'pear'];
$fruits = ['banana', 'orange', ...$parts, 'watermelon'];
```
and non-numeric
```php
$arr = ["a" => 1, ...$b];
```
6. Readonly props or classes
```php
readonly class C { public int $i; }
class C { public readonly int $i; }
```
7. Dnf types, standalone types
```php
function f((A&B)|C $x) {}
```
8. propery hooks
```php
public string $credits { get { /*...*/ } }
```
9. Asymmetric visibility
```php
public private(set) int $i;
```
10. new without parentheses chaining
```php
new C($a)->meth();
```

corneliuhoffman:
Also to do, check the results against the treesitter parser.

languages/php/menhir/parser_php.mly
```diff
@@ -938,6 +938,9 @@ expr:
  | simple_expr T_AND_EQUAL    expr { AssignOp($1,(AssignOpArith And,$2),$3) }
  | simple_expr T_OR_EQUAL     expr { AssignOp($1,(AssignOpArith Or,$2),$3) }
  | simple_expr T_XOR_EQUAL    expr { AssignOp($1,(AssignOpArith Xor,$2),$3) }
+ | T_THROW expr_or_dots  { Throw($1,$2) }
+
+ | simple_expr T_NULL_COALLESCING  expr { AssignOp($1,(AssignOpArith Nullish,$2),$3) }
```
dimitris-m: I could be missing something, but why `AssignOp` ?

tests/rules/null-coallesce.yaml
```diff
@@ -0,0 +1,24 @@
...
+    $BB ?? $AA
+  severity: ERROR
+- id: embedded_word
+  message: "The web server executes a command on potential unescaped user input, which may lead to remote command execution."
+  languages:
+    - php
+  severity: ERROR
+  mode: taint
+  options:
+    taint_assume_safe_numbers: true
```
dimitris-m: I think this is not needed for this specific test example

tests/rules/null-coallesce.yaml
```diff
@@ -0,0 +1,24 @@
...
+  pattern-sinks:
+  - patterns:  
+    - pattern: $SOURCE
+  pattern-sources:
+  - patterns:
+    - focus-metavariable: $SOURCE
+    - pattern-either:
```
dimitris-m, line 17: Maybe this could be `pattern-inside` ?

languages/php/menhir/parser_php.mly
corneliuhoffman: it is a binary operator. Presumably I could do something fancier but followed the [tree-sitter](https://github.com/opengrep/opengrep/blob/80a6332c484b9e66bfc00d172c986da84f0523fb/languages/php/tree-sitter/Parse_php_tree_sitter.ml#L1272) and  used the "Nullish" operator  as it exists in the AST_generic, see also [the comment](https://github.com/opengrep/opengrep/blob/80a6332c484b9e66bfc00d172c986da84f0523fb/libs/ast_generic/AST_generic.ml#L1082) there.

tests/parsing/php/null-coallesce.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+
+function f($filename) {
+    // ruleid: AIK_exec-use
```
dimitris-m: this should be `// ruleid: test` I think

dimitris-m: sorry, this is not needed at all, since this is parsing only

tests/rules/null-coallesce.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+function f($filename) {
+    // ruleid: AIK_exec-use
```
dimitris-m: this should be `// ruleid: test`

tests/rules/null-coallesce.php
```diff
@@ -0,0 +1,12 @@
...
+    "convert " . $filename;
+}
+function f1($filename) {
+    // ruleid: AIK_exec-use
```
dimitris-m: this should be `// ruleid: test`

tests/rules/null-coallesce.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+function f($filename) {
+    // ruleid: AIK_exec-use
+    "convert " . $filename;
```
dimitris-m: I think the problematic case was using `"convert $filename"` so this is what we should test here

tests/rules/null-coallesce.yaml
```diff
@@ -0,0 +1,24 @@
+rules:
+- id: Throw
```
dimitris-m: `- id: test`

tests/rules/null-coallesce.yaml
corneliuhoffman: It did not really work with pattern inside, it is a simplified version based on the yaml we had before.

tests/rules/null-coallesce.yaml
```diff
@@ -0,0 +1,24 @@
...
+  languages:
+    - php
+  message: "throw"
+  pattern: |
+    $BB ?? $AA
+  severity: ERROR
+- id: embedded_word
```
dimitris-m: This rule `embedded_word` is not tested, the `id` much match the comment in the code being tested.

dimitris-m: In fact we should have 1 rule only, and duplicate the test for the case `??` above.

dimitris-m (commented):
I recommend to rebase the branch so that the tests are first, before the grammar changes.

Then we can checkout the commit with the test only, confirm that the test fails -- and if the tests pass at the head of the branch, we are good.

PS I'm doing this now on a local copy of the branch

PPS All good, but I added a function in the rule test, it looks like this in my test (not pushed to your branch):

```php
<?php

function f($filename) {
    // ruleid: test
    "convert " . $filename;
}

function g($filename) {
    // ruleid: test
    "convert $filename";
}

function f1($filename) {
    // ruleid: test
    return shell_exec($filename)
            ?? throw new RuntimeException('Failed to get filename');
}
```

Indeed without your menhir changes, line 10 above (on function `g`) fails to match.
But the first test in `f` still succeeds, so maybe we don't need it.

And the rule:

```yaml
rules:

- id: test_1
  languages:
    - php
  message: "throw"
  pattern: |
    $BB ?? $AA
  severity: ERROR

- id: test
  message: "The web server executes a command on potential unescaped user input, which may lead to remote command execution."
  languages:
    - php
  severity: ERROR
  mode: taint
  options:
    taint_assume_safe_numbers: true
  pattern-sinks:
  - patterns:  
    - pattern: $SOURCE
  pattern-sources:
  - patterns:
    - focus-metavariable: $SOURCE
    - pattern-either:
        - pattern: function $FUNC(..., $SOURCE, ...) { ... }
```

which means only `if: test` is relevant, the first rule should be removed or placed in separate test.

tests/rules/null-coallesce.php
corneliuhoffman: oups

corneliuhoffman:
rebased, changes etc

tests/parsing/php/null-coallesce.php
dimitris-m: I suggest we remove the comment

tests/rules/null-coallesce.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+function f($filename) {
+    // ruleid: embed_word
```
dimitris-m: this should be `// ruleid: test`

and in the rule we should use also the same `id: test`.

tests/rules/null-coallesce.php
```diff
@@ -0,0 +1,11 @@
...
+    "convert $filename";
+}
+function f1($filename) {
+    return shell_exec($filename)
```
dimitris-m: We should add `// ruleid: test` above this line for the match to be expected.

tests/rules/null-coallesce.yaml
dimitris-m: `id: test`

dimitris-m (commented):
some more comments

corneliuhoffman:
cleaned the test further

dimitris-m (approved):
This looks good, I'll merge

PR #300, opened 2025-06-18
Fix #291: cwd can begin with lowecase drive letter on windows

### LINKED ISSUE #291: Opengrep fails to scan on Windows due to case sensitivity in path comparison

dimitris-m:
Closes #291.

Plus small changes in other places, as noted in commits.

### DIFF
--- a/libs/paths/Ppath.ml
+++ b/libs/paths/Ppath.ml
@@ -287,12 +287,26 @@ let remove_prefix root path =
 (* Builder entry points *)
 (*****************************************************************************)
 
+let capitalize_drive_letter path =
+  if String.length path >= 2 &&
+     Char.equal path.[1] ':' &&
+     path.[0] >= 'a' && path.[0] <= 'z'
+  then
+    String.capitalize_ascii path
+  else
+    path
+
+let cwd () =
+  let cwd = Unix.getcwd () in
+  if not Sys.win32 then cwd
+  else capitalize_drive_letter cwd
+
 (*
    Make a path absolute, using getcwd() if needed.
    I hesitated to put this into Fpath_ since Fpath is purely syntactic.
 *)
 let make_absolute path =
-  if Fpath.is_rel path then Fpath.(v (Unix.getcwd ()) // path)
+  if Fpath.is_rel path then Fpath.(v (cwd ()) // path)
   else
     (* Here, we must make a syscall, because we are making an unnormalized path absolute
        so that we can compare it to a normalized path.


libs/paths/Ppath.ml
```diff
@@ -287,12 +287,27 @@ let remove_prefix root path =
 (* Builder entry points *)
 (*****************************************************************************)
 
+let capitalize_drive_letter path =
+  if String.length path >= 2 &&
+     Char.equal path.[1] ':' &&
+     path.[0] >= 'a' && path.[0] <= 'z'
+  then
+    let upper = Char.uppercase_ascii path.[0] in
```
corneliuhoffman: can one not use String.capitalize_ascii?

dimitris-m: I left a comment on the commit message: since some windows filesystems like NTFS can be case sensitive, I did not want to change the whole path.

dimitris-m:
> This looks good to me, I don't have a  window machine to test though

I tested with a python script on windows, maybe I can paste it on the PR description.

libs/paths/Ppath.ml
corneliuhoffman: yes but String.capitalize_ascii just capitalizes the first char. I think that is equivalent to your hack

dimitris-m: oh, ok yes

dimitris-m: fixed

Issue #301, opened 2025-06-18
Support `.semgrepignore` or the `--semgrepignore-filename` logic with `--baseline-commit`


tom-paz:
# Feature Request: Support .semgrepignore files with --baseline-commit --experimental

## Is your feature request related to a problem? Please describe.

When using `--baseline-commit --experimental` for differential scanning, the `.semgrepignore` file from the source branch is not being respected for changed files. Files that have been modified between the baseline and source branch should still be ignored if they match patterns in `.semgrepignore`, but currently they are scanned regardless of ignore patterns.

For example:
```bash
opengrep scan --baseline-commit=abc123 --experimental
```

The differential scan correctly identifies changed files between commits, but ignores the `.semgrepignore` patterns when deciding which of those changed files to actually scan.

## Describe the solution you'd like

The `--baseline-commit --experimental` differential scanning should respect the `.semgrepignore` file from the source branch. When determining which changed files to scan, it should exclude any files that match ignore patterns, even if they were modified between the baseline and source commits.

## Describe alternatives you've considered

- Not using differential scanning when ignore files are needed (defeats the purpose of efficient scanning)
- Manually filtering results after scanning (inefficient and error-prone)

## Use case

Consider a project where files have changed between the baseline commit and source branch:
- `src/app.js` (modified)
- `tests/app.test.js` (modified) 
- `dist/bundle.min.js` (modified)

With a `.semgrepignore` file containing:
```
tests/
dist/
```

**Expected behavior**: When running `opengrep scan --baseline-commit=main --experimental`, the differential scan should:
1. Detect all three files as changed
2. Apply ignore patterns from `.semgrepignore`
3. Only scan `src/app.js` (exclude the test and dist files)

**Current broken behavior**: 
- All three changed files are scanned, including `tests/app.test.js` and `dist/bundle.min.js`
- The `.semgrepignore` patterns are ignored for the differential file selection
- Result: Wasted time scanning files that should be excluded, and potential false positives in ignored files

This would enable efficient differential scanning that respects project ignore patterns, making it practical for CI/CD pipelines that need to scan only relevant changed files.

## Additional context

This issue specifically occurs with the combination of `--baseline-commit` and `--experimental` flags during differential scanning. The feature should filter the list of changed files through the `.semgrepignore` patterns before scanning, ensuring that ignored files are excluded even if they were modified between commits.

PR #303, opened 2025-06-18
Adding ignore logic to --baseline-commit 

tom-paz:
#301 

Here's a PR description for the changes:

### Problem
Previously, when using `--baseline-commit` for differential scans, the baseline scan would ignore `.semgrepignore` patterns and scan ALL files in the repository, even if those files were excluded from the head scan. This led to inconsistent results where ignored files could still produce findings in the baseline comparison.

### Solution
This PR introduces the `--enable-semgrep-ignore` flag that enables `.semgrepignore` filtering for baseline scans. When enabled, files matching ignore patterns are excluded from both the head AND baseline scans, ensuring consistent results.

### Usage
```bash
# Use default .semgrepignore file
opengrep scan --experimental --baseline-commit=<HASH> --enable-semgrep-ignore

# Use custom ignore file
opengrep scan  --experimental  --baseline-commit=<HASH> --enable-semgrep-ignore --semgrepignore-filename=.myignore
```

### Key Features
- **Opt-in behavior**: The feature is disabled by default to maintain backward compatibility
- **Consistent filtering**: Files ignored in the head scan are also ignored in the baseline scan
- **Custom ignore files**: Supports `--semgrepignore-filename` for custom ignore patterns
- **Clear validation**: Requires `--baseline-commit` to be specified when using `--enable-semgrep-ignore`


src/osemgrep/cli_scan/Diff_scan.ml
```diff
@@ -171,17 +171,28 @@ let scan_baseline_and_remove_duplicates (caps : < Cap.chdir ; Cap.tmp >)
...
+                        Find_targets.get_target_fpaths conf.targeting_conf
+                          conf.target_roots
+                      in
+                      (* Performing a scan on the same set of files for the
+                         baseline that were previously scanned for the head.
+                         In Interfile mode, the matches are influenced not
+                         only by the file displaying matches but also by its
+                         dependencies. Hence, merely rescanning files with
+                         matches is insufficient. *)
+                      (all_in_baseline, paths_in_scanned)
```
dimitris-m: I need to understand why something changed here, this path should not be executed since it's for the PRO engine.

Have you tested this and does it have any effect??

src/osemgrep/cli_scan/Diff_scan.ml
```diff
@@ -200,7 +211,7 @@ let scan_baseline_and_remove_duplicates (caps : < Cap.chdir ; Cap.tmp >)
 (*****************************************************************************)
 
 let scan_baseline (caps : < Cap.chdir ; Cap.tmp >) (conf : Scan_CLI.conf)
-    (profiler : Profiler.t) (baseline_commit : string) (targets : Fpath.t list)
+    (profiler : Profiler.t) (baseline_commit : string) (_targets : Fpath.t list)
```
dimitris-m: There needs to be a good reason and a clear explanation why a parameter is to be ignored.

src/osemgrep/cli_scan/Diff_scan.ml
tom-paz: So you dont want me to patch the PRO flow ?  is that what you mean ?

dimitris-m: Anything PRO will eventually be removed.
It has no effect at all, unless if I'm missing something critical!

dimitris-m:
Have a look at https://github.com/opengrep/opengrep/pull/326. 

Let me know if this works for your use-case, I tested locally and the results seem correct.

PR #305, opened 2025-06-23
Release version v1.4.2

### LINKED ISSUE #291: Opengrep fails to scan on Windows due to case sensitivity in path comparison

dimitris-m:
## Bug fixes

* Fix #291: cwd can begin with lowecase drive letter on windows by @dimitris-m in #300

**Full Changelog**: https://github.com/opengrep/opengrep/compare/v1.4.1...v1.4.2


corneliuhoffman (commented):
LGTM

Issue #297, opened 2025-06-13
PHP 5->Php 8.4 parser


corneliuhoffman:
## **Describe the bug**

 In issue #295 we discovered 2 things
1. the menhir parser (default) is not updated to latest php
2. there is a bug in the tree-sitter parser that interpolates strings wrongly. 

While we should eventually fix the interpolation issue in tree-sitter this does not help since the pattern parsing in php is done via menhir.
In digging in the php updates I found other unparsable syntax:

issues remainig for the parser:
1. [ ] static return type:

```php
function f(): static {}
```
2. [x]  Match expressions:
```php
match($x) {
    1 => 'one',
    2 => 'two',
};
```
3. [x]  Enums:
```php
enum Status {
  case Active;
  case Inactive;
}
```
4. [x]  Null coallescing assignment
```php
$a ??= 'default';
```
5. [x]  Array spread:
```php
$parts = ['apple', 'pear'];
$fruits = ['banana', 'orange', ...$parts, 'watermelon'];
```
and non-numeric
```php
$arr = ["a" => 1, ...$b];
```
6. [ ]  Readonly props or classes
```php
readonly class C { public int $i; }
class C { public readonly int $i; }
```
7. [ ]  Dnf types, standalone types
```php
function f((A&B)|C $x) {}
```
8. [ ]  propery hooks
```php
public string $credits { get { /*...*/ } }
```
9. [ ]  Asymmetric visibility
```php
public private(set) int $i;
```
10. [x]  new without parentheses chaining
```php
new C($a)->meth();
```
11. [ ] [Class constant visibility](https://wiki.php.net/rfc/class_const_visibility), introduced in PHP version 7.1. (thanks @bblumenwiese)

Example:
```php
class C { public const string P = 'Public'; }
class C { protected const string P = 'Protected'; }
class C { private const string P = 'Private'; }
```

### **To Reproduce**
Steps to reproduce the behavior.

This bug is somewhat subtle and so not very obvious to reproduce, there are two possible ways
1. in Parse_target2.ml comment out [this line](https://github.com/opengrep/opengrep/blob/80a6332c484b9e66bfc00d172c986da84f0523fb/src/parsing_languages/Parse_target2.ml#L129), make the project and try to parse any of the above.
2. do not comment the project but create code containing one of the above as well as the following syntax:
```php
function f($filename) {
    "convert $filename";
}
```
then apply the following rule 
```yaml
rules:
- id: embedded_word
[...]

PR #306, opened 2025-06-23
Match and enum in php

### LINKED ISSUE #304: Some issues with enums in php
In improving the php parser I arrived at teh following issue:
In the treesitter parser (and in my version of the menhir), the enum
```php
enum Status {
    case Active;
    case Inactive;
}
```
is essentially the same as the class
```php
final class Status {
    public static Status $Active;
    public static Status $Inactive;

    private function __construct() {}

    public static function cases(): array {
        return [self::$Active, self::$Inactive];
    }
}
```

In particular the cases become the static properties $Active and $Inactive receive a dollar sign (in fact in treesitter they eroneously do not) in the parsed version. This creates a problem:
In opengrep there is a function `Match_rules.is_relevant_rule_for_xtarget` which selects regexps and identifiers from the rule and uses PCRE to quickly figure out which of the rules are relevant for each file. because the extra $ does not exist in the actual code, the rule (even if identical to the code) will be deigned irrelevant for the file and so not found.

A simple solution would be to modify the PCRE for languages that use $ for variables and drop the $ in the search.

corneliuhoffman:
this resolves #304 as well as 2 cases in #297 

It also fixes an issue in switch (and match) that prevented ellipses in the body of switch. In particular
```php
switch($F) 
   {...}
  ```
  and 
  ```php
  match ($F)
  {...}
  ```
  are now valid rules.

### DIFF
--- a/languages/php/menhir/parser_php.mly
+++ b/languages/php/menhir/parser_php.mly
@@ -714,7 +714,7 @@ unticked_class_declaration:
          c_enum_type = None;
        }
      }
-  (* in the following we assum that the cases of a statement appear before constants,  methods and use constructs.*)
+  (* in the following we assume that the cases of a statement appear before constants,  methods and use constructs.*)
    | T_ENUM ident_class_name implements_list 
     "{" enum_single* member_declaration* "}"
   {let members = $5 @ $6 in  { c_type = Enum $1; c_name = $2; c_extends = None; c_tparams = None;
--- a/languages/php/menhir/parser_php.mly
+++ b/languages/php/menhir/parser_php.mly
@@ -71,8 +71,9 @@ let mk_param s =
     p_modifier = None;
     p_variadic = None;
   }
-let mk_comm_list list =  List.rev (List.fold_left (fun acc x -> Right(Tok.FakeTok(",", None)) :: Left(x) :: acc) [] list)
-let make_class_var tok1 body tok2 =ClassVariables (NoModifiers(tok1), None, mk_comm_list body,tok2 )
+
+let make_class_vars single = 
+  ClassVariables(NoModifiers(Tok.FakeTok("", None)),None, [Left(single)],Tok.FakeTok("", None))
 
 let mk_var (s, tok) =
   match s with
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: languages/php/menhir/parser_php.mly [hunk 1]
```diff
@@ -690,14 +714,35 @@ unticked_class_declaration:
          c_enum_type = None;
        }
      }
- | T_ENUM ident_class_name ":" type_php type_constr_opt
-    "{" enum_statement* "}"
-     { { c_type = Enum $1; c_name = $2; c_extends = None; c_tparams = None;
-         c_implements = None; c_body = ($6, $7, $8);
+  (* in the following we assum that the cases of a statement appear before constants,  methods and use constructs.*)
```
dimitris-m: typo: `assum`

tests/rules/enum.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+function f($filename)
+{
+  // ruleid: test
+  bad("convert $filename");
+}
```
dimitris-m: do we need this one?

tests/rules/enum.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - patterns:
+    - pattern: $SOURCE
+    - pattern-inside: bad($SOURCE)
+  - patterns:  
+    - pattern: $SOURCE
+  pattern-sources:
+  - patterns:
+    - focus-metavariable: $SOURCE
+    - pattern-either:
+        - pattern: function $_(..., $SOURCE, ...) { ... }
```
dimitris-m: Maybe we can remove this rule

tests/rules/match-rules.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+function f($filename) {
+    // ruleid: test
+    bad("convert $filename");
+}
```
dimitris-m: Maybe this is not needed, but we could have a test to show that tainting works in the presence of the new constructs.

tests/rules/enum.yaml
```diff
@@ -0,0 +1,26 @@
+rules:
+- id: enum_rule
+  pattern: "enum Status {
+    ...}"
```
dimitris-m: Maybe we could have a more specific rule? 

Or at least a tainting test that requires enums to parse and transform correctly.

tests/rules/enum.php
corneliuhoffman: we did because it is the easiest way of making sure the file is not parsed by treesitter

tests/rules/enum.yaml
corneliuhoffman: same as above

tests/rules/match-rules.yaml
```diff
@@ -0,0 +1,25 @@
+rules:
+- id: match_test
+  message: "match_test"
+  languages:
+    - php
+  severity: ERROR
+  pattern: match($filename) {
+              1 => bad("convert $filename"),
+              2 => 'two',
+              default => 'more',}
```
dimitris-m: would that also work?

```
pattern:
  match($F) {
  ...
  }
```

PR #307, opened 2025-06-24
fix(install): reduce github api calls, add retry logic

JonZeolla:
This reduces the number of GitHub API calls during install. In using the install script we found that it will sometimes fail, likely due to GitHub rate limiting

It does that by using the `/latest/` redirect from the github download endpoint to identify what the latest version is, so we don't need to list releases and then select one.

It also adds retrying if there's failures on the download, and it fixes a small bug where before it wouldn't update the latest symlink if the requested version was already downloaded.

Note that technically the latest symlink points to the latest version installed via `install.sh` and not the latest release of the opengrep project, but I didn't have a great alternative here (plus that was the existing pattern, so I left it)

## Testing

<details><summary>Happy Path</summary>

```console
$ rm -rf ~/.opengrep/
$ ./install.sh

*** Installing Opengrep v1.4.2 for Darwin (arm64) ***

############################################################################################################################################################################################################# 100.0%

Successfully installed Opengrep binary to /Users/jonzeolla/.opengrep/cli/v1.4.2/opengrep

To launch Opengrep now, type:
opengrep

To check Opengrep version, type:
opengrep --version

$ ~/.opengrep/cli/latest/opengrep --version
1.4.2
$ ./install.sh -v v1.4.1

*** Installing Opengrep v1.4.1 for Darwin (arm64) ***

############################################################################################################################################################################################################# 100.0%

Successfully installed Opengrep binary to /Users/jonzeolla/.opengrep/cli/v1.4.1/opengrep

To launch Opengrep now, type:
opengrep

To check Opengrep version, type:
opengrep --version

$ ~/.opengrep/cli/latest/opengrep --version
1.4.1
$ ./install.sh -v v1.4.2

*** Installing Opengrep v1.4.2 for Darwin (arm64) ***

Destination binary /Users/jonzeolla/.opengrep/cli/v1.4.2/opengrep already exists.

To launch Opengrep now, type:
opengrep

To check Opengrep version, type:
opengrep --version

$ ~/.opengrep/cli/latest/opengrep --version
1.4.2
$ ./install.sh -v v1.4.0

*** Installing Opengrep v1.4.0 for Darwin (arm64) ***

############################################################################################################################################################################################################# 100.0%

Successfully installed Opengrep binary to /Users/jonzeolla/.opengrep/cli/v1.4.0/opengrep

To launch Opengrep now, type:
opengrep

To check Opengrep version, type:
opengrep --version

$ ~/.opengrep/cli/latest/opengrep --version
1.4.0
$ ./install.sh

*** Installing Opengrep v1.4.2 for Darwin (arm64) ***

Destination binary /Users/jonzeolla/.opengrep/cli/v1.4.2/opengrep already exists.

To launch Opengrep now, type:
opengrep

To check Opengrep version, type:
opengrep --version

$ ~/.opengrep/cli/latest/opengrep --version
1.4.2
```
</details>
<details><summary>Connectivity issues</summary>

To test this I edited `install.sh` to make invalid urls

```console
$ ./install.sh # I broke the URL on purpose
Failed to resolve latest version. The redirect from https://github.com/opengrep/opengrep/releases/latest/downloadfail/opengrep_osx_arm64 did not work.
Please check your internet connection or specify a version with -v
$ ./install.sh -v v1.4.2 # I broke the URL on purpose

*** Installing Opengrep v1.4.2 for Darwin (arm64) ***

curl: (56) The requested URL returned error: 404

⚠️  attempt 1/3 failed for: curl --fail --location --progress-bar https://github.com/opengrep/opengrep/releases/download/v1.4.2/opengrep_osx_arm64fail; retrying in 2s
curl: (56) The requested URL returned error: 404

⚠️  attempt 2/3 failed for: curl --fail --location --progress-bar https://github.com/opengrep/opengrep/releases/download/v1.4.2/opengrep_osx_arm64fail; retrying in 2s
curl: (56) The requested URL returned error: 404

⚠️  attempt 3/3 failed for: curl --fail --location --progress-bar https://github.com/opengrep/opengrep/releases/download/v1.4.2/opengrep_osx_arm64fail

Error: Failed to download version v1.4.2
Error: Unable to fetch available versions. Please check your internet connection.
$ ls ~/.opengrep
ls: /Users/jonzeolla/.opengrep: No such file or directory
```
</details>
<details><summary>User requests a version that doesn't exist</summary>

```console
$ ./install.sh -v nonexistant

*** Installing Opengrep nonexistant for Darwin (arm64) ***

curl: (56) The requested URL returned error: 404

⚠️  attempt 1/3 failed for: curl --fail --location --progress-bar https://github.com/opengrep/opengrep/releases/download/nonexistant/opengrep_osx_arm64; retrying in 2s
curl: (56) The requested URL returned error: 404

⚠️  attempt 2/3 failed for: curl --fail --location --progress-bar https://github.com/opengrep/opengrep/releases/download/nonexistant/opengrep_osx_arm64; retrying in 2s
curl: (56) The requested URL returned error: 404

⚠️  attempt 3/3 failed for: curl --fail --location --progress-bar https://github.com/opengrep/opengrep/releases/download/nonexistant/opengrep_osx_arm64

Error: Failed to download version nonexistant
Error: Version nonexistant does not appear to be a valid opengrep release
Available versions (latest 3):
v1.4.2
v1.4.1
v1.4.0
```
</details>


JonZeolla:
Ah, it seems I didn't finish my WIP. That is my fault, I will send something within the next hour

install.sh
```diff
@@ -8,33 +10,75 @@ print_usage() {
...
+        ((n++))
+    done
+}
+
+
 # Function to get available versions - already checked when running main
 get_available_versions() {
     command -v curl >/dev/null 2>&1 || { echo >&2 "Required tool curl could not be found. Aborting."; exit 1; }
-    curl -s https://api.github.com/repos/opengrep/opengrep/releases | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
+    # This extracts and validates the version tag(s) without requiring jq
```
dimitris-m: But there was no use of `jq`, not sure if this comment is needed?

JonZeolla: I'm fine with removing it, it's just that my first thought when I read this code is "this would be way cleaner as `curl -s https://api.github.com/repos/opengrep/opengrep/releases | jq -r '.[0].tag_name'`", so the comment is helpful to be explicit that the reason for doing it this way is to avoid another host dependency

install.sh
```diff
@@ -71,22 +115,64 @@ main () {
...
 
     # check if binary already exists
     if [ -f "${INST}/opengrep" ]; then
         echo "Destination binary ${INST}/opengrep already exists."
     else
+        # Download to temp file first
+        TEMP_FILE=$(mktemp)
+        trap 'rm -f $TEMP_FILE' EXIT
+
+        if ! retry curl --fail --location --progress-bar "${URL}" > "${TEMP_FILE}"; then
```
JonZeolla, line 208: Moving the conversation here:
[Comment 1](https://github.com/opengrep/opengrep/pull/307#issuecomment-3002033389)

> As a first note, if the requested version is not found, then why retry?
> 
> Will review in more detail as soon as possible.

[Comment 2](https://github.com/opengrep/opengrep/pull/307#issuecomment-3005095383)

> @dimitris-m because sometimes the failure is due to a GitHub rate limit / failure, not because the release isn't actually there.

[Comment 3](https://github.com/opengrep/opengrep/pull/307#issuecomment-3005259747)

> But in case of rate limit, what is the point of immediately retrying?
>
> I'd rather have the http code inform the error message and fail in such cases, which I assume to be infrequent.

JonZeolla: I've been using this script for about a day and I've hit this problem when building in a GitHub Action pipelines a small handful of times (I think maybe 4 or 5 times). This added retry has succeeded ever since. It's a small sample size, but I'm willing to bet something as simple as `curl --fail || curl --fail` would even help a great bit, but having the 2 second delay seemed prudent

install.sh
dimitris-m: I think we can remove the comment safely.

install.sh
```diff
@@ -8,33 +10,75 @@ print_usage() {
...
+
 # Function to get available versions - already checked when running main
 get_available_versions() {
     command -v curl >/dev/null 2>&1 || { echo >&2 "Required tool curl could not be found. Aborting."; exit 1; }
-    curl -s https://api.github.com/repos/opengrep/opengrep/releases | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
+    # This extracts and validates the version tag(s) without requiring jq
+    retry curl -s --fail https://api.github.com/repos/opengrep/opengrep/releases \
+      | grep '"tag_name":' \
+      | sed -E 's/.*"([^"]+)".*/\1/' \
+      | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$'
```
dimitris-m: I see why one may want to add this, but it restricts us a bit, for example if some future release deviates from that format? For example v1.5.0-alpha.

So I'm not sure about this change, if something is found in the releases URL, then it's a release so why block it here?

install.sh
```diff
@@ -8,33 +10,75 @@ print_usage() {
...
+    [[ $1 =~ ^[0-9]+$ ]] && { retry_count=$1; shift; }
+    [[ $1 =~ ^[0-9]+$ ]] && { retry_delay=$1; shift; }
+
+    local n=1
+    until "$@"; do
+        if (( n >= retry_count )); then
+            printf '⚠️  attempt %d/%d failed for: %s\n' "$n" "$retry_count" "$*" >&2
+            return 1
+        fi
+        printf '⚠️  attempt %d/%d failed for: %s; retrying in %ds\n' "$n" "$retry_count" "$*" "$retry_delay" >&2
```
dimitris-m: I'm personally cautious with fancy symbols like ⚠️.

Do all terminals support it? Does it interfere with logging (in clients)?

dimitris-m: Note also that we don't use such symbols for other actions, successful or other warnings and errors. 
So for consistency I would remove them.

install.sh
```diff
@@ -8,33 +10,75 @@ print_usage() {
...
+retry() {
+    # This facilitates arbitrary retry logic for the provided command
+    #
+    # USAGE:
+    #  # Use the defaults
+    #  retry curl --fail https://example.com
+    #  # Provide an updated count
+    #  retry 10 curl --fail https://example.com
+    #  # Provide an updated count and delay
+    #  retry 5 10 curl --fail https://example.com
```
dimitris-m: It seems the only use in the script itself is with default values.

So why not simplify?

install.sh
```diff
@@ -8,33 +10,75 @@ print_usage() {
...
+    local retry_count=3
+    local retry_delay=2
+
+    # Update the default values if provided
+    [[ $1 =~ ^[0-9]+$ ]] && { retry_count=$1; shift; }
+    [[ $1 =~ ^[0-9]+$ ]] && { retry_delay=$1; shift; }
```
dimitris-m: Is that actually used in the script?

install.sh
```diff
@@ -71,22 +115,64 @@ main () {
...
+    # Now set INST with the final version
+    INST="${PREFIX}/${FINAL_VERSION}"
 
     # check if binary already exists
     if [ -f "${INST}/opengrep" ]; then
         echo "Destination binary ${INST}/opengrep already exists."
     else
+        # Download to temp file first
+        TEMP_FILE=$(mktemp)
+        trap 'rm -f $TEMP_FILE' EXIT
```
dimitris-m, line 206: It's not very likely that a download will be interrupted in the middle, especially in a cloud environment.

What motivated you to add this?

dimitris-m (commented):
I made some minor improvements to our script, orthogonal to the material in this PR.
You will probably need to rebase.

Pls see also comments above.

dimitris-m:
We have now made a lot of changes on the install script, since PR #315 which added signature verification.

Please can you revise your PR if you still think there are additions you want to suggest? 

As for retry logic, I think this can be done by the client: you can create a script that repeats the installation script invocation as required.

Pls let me know if this PR should remain open.

install.sh
JonZeolla: I expected that we'd want users to explicitly opt-into non-stable releases like an alpha, etc. and that we'd add support for that in future work when that was desirable

install.sh
JonZeolla: I'm fine removing it

install.sh
JonZeolla: I'm good with that; I was using this for testing to find some numbers which worked well given the scenario

install.sh
JonZeolla: I'm good with removing; same answer as https://github.com/opengrep/opengrep/pull/307#discussion_r2180515632

install.sh
JonZeolla: It's just my standard approach to avoid downloading files directly to their final runtime location, because I've experienced many interruptions/blips over the years which have wasted a lot of my time troubleshooting 😅

More common when someone is using this to download over wifi for use on their local computer, not CI

install.sh
JonZeolla: Removed in ee1923ddde7ecb755f2dba25fe8d65509dd5284e

install.sh
JonZeolla: Removed in ee1923ddde7ecb755f2dba25fe8d65509dd5284e

install.sh
JonZeolla: Removed in ee1923ddde7ecb755f2dba25fe8d65509dd5284e

install.sh
```diff
@@ -230,6 +296,10 @@ main() {
         echo "with a symlink from ${LATEST}/opengrep"
     fi
 
+    # Always update the symlink to point to the requested version
+    rm -f "${LATEST}" || exit 1
+    ln -s "${INST}" "${LATEST}" || exit 1
+
```
JonZeolla: It seems like this was "borrowed" into https://github.com/opengrep/opengrep/commit/8935a5951851e6d28639a2f1e807350453dfa051
```suggestion
```

install.sh
```diff
@@ -139,7 +161,33 @@ main() {
         exit 1
     fi
 
-    URL="https://github.com/opengrep/opengrep/releases/download/${VERSION}/${DIST}"
+    if [ "$REQUESTED_VERSION" = "latest" ]; then
```
dimitris-m, line 163: But latest is the default when no version is passed.

Why complicate the script.

install.sh
dimitris-m: The updated script cleans up any files on error, so this is not needed.

install.sh
```diff
@@ -38,20 +38,41 @@ check_has_curl() {
     }
 }
 
+retry() {
```
dimitris-m, line 41: There is now some logic that depends on the http return code.

For example if the .cert gives 404, we assume it's an older version.

I'm not a big fan of having retry logic in this script. Given that it cleans after itself, why not follow the unix philosophy? If someone has connectivity issue they can create their own retry logic.

install.sh
JonZeolla: Removed in ee1923ddde7ecb755f2dba25fe8d65509dd5284e

install.sh
JonZeolla: Removed in d2ff233f85ada093cbc2f1443830ea9a1aaa72f2

install.sh
JonZeolla: Regarding https://github.com/opengrep/opengrep/pull/307#issuecomment-3028079088, specifically:
> As for retry logic, I think this can be done by the client: you can create a script that repeats the installation script invocation as required.

While I agree that it _is_ possible, I don't think it's a great UX. For instance, if you're using this installation script in a `Dockerfile` now you're wrapping the `install.sh` with your own custom installer, or doing error management in the `Dockerfile` itself.

dimitris-m (commented):
I have some concerns about this PR.

- It states that less calls are made: this is not true (am I wrong?)

- It adds retry logic which may interfere with other logic; and honestly I never saw an install script dealing with such issues. The duckdb one was a kind of inspiration (to the original author) and we think it's a very good one. I'm simply not convinced about the need for this functionality _within_ the script. I can understand the need for it as a wrapper.

- It adds new logic about `latest` as a version, which I think is not needed, is it? The default is to download the latest one if there is no version.

In summary, only the retry logic is "new behaviour", and I'm not convinced it belongs here...

install.sh
dimitris-m: but it's enough to repeat until success, if you know that the version exists.
especially without `-v <version>`, it should only fail on connectivity issues (modulo `--verify-signatures`).

install.sh
JonZeolla: Because that requires two GETs (more likely to be rate limited), and the approach I took is a HEAD and a GET for the `latest` path.

JonZeolla: It does seem like a pretty minor change in the grand scheme, but the only reason I started down this path is because I **immediately** started having rate limiting issues as soon as I began using this script.

install.sh
JonZeolla: I tend not to retry infinitely if only to avoid a thundering herd problem during an upstream outage (like GitHub API downtime)

install.sh
```diff
@@ -150,19 +198,35 @@ main() {
...
     else
-        echo
-        echo "*** Installing Opengrep ${VERSION} for ${OS} (${ARCH}) ***"
-
         # cleanup on error
         trap '[ "$?" -eq 0 ] || cleanup_on_failure $INST' EXIT
 
+        # Download to temp file first
+        TEMP_FILE=$(mktemp)
+        trap 'rm -f $TEMP_FILE' EXIT
```
dimitris-m, line 206: I did not confirm but I was under the impression that a new trap on EXIT will overwrite the previous one.

dimitris-m:
Most of the concerns addressed in this PR have been addressed as the scripts evolved, see #796 for example.

The retry logic will not be added at this point, but if it were to be added, why not use curl's `--retry N --retry-max-time M...` etc?

PR #309, opened 2025-06-26
Improve install script

dimitris-m:
### Changes

- Do not assume that `~/.local/bin` is in the PATH, even if it exists, as this is not always the case;
- Better messages & small misprints


corneliuhoffman (approved):
lgtm

PR #315, opened 2025-07-01
Add cosign signature

### LINKED ISSUE #87: Sign Release Binaries for Security and Integrity
Signing Opengrep’s release binaries with a tool like Cosign would improve security by allowing users to verify their authenticity and integrity, helping prevent supply chain attacks.

corneliuhoffman:
Fixes #87
The PR has two additions:
1. changes in the `rolling-release` action to sign and add signatures to each image
2. changes to the `install` script to include signature verification:
from now on the way to run it is:
`./install (-v version)` this install the file and the sigs (if present) verifies the signatures and warns on signature error 
`./install.sh (-v version) --verify-signatures` this installs the file and the signs (if present), verifies the signatures and in the case of signature errors it deletes the installed files and  fails.

Note:
it is not obvious how to test the install script before having a relesea with signatures. However one can test it locally:
1. download the three files for your os from https://github.com/opengrep/opengrep/releases/tag/untagged-d11b21f8c54255d9cadb
2. move them to $HOME/.opengrep/cli/test and rename them removing the reference to the OS (i.e opengrep, opengrep.sig opengrep.cert)
3. point the $HOME/.opengrep/cli/latest to $HOME/.opengrep/cli/test
4. run the install.sh test with and without --verify-signatures

You can also directly run install.sh with and without --verify-signatures to see it failing

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -130,8 +130,6 @@ jobs:
         uses: sigstore/cosign-installer@v3
 
       - name: Sign all binaries in artifacts
-        env:
-          COSIGN_EXPERIMENTAL: "1"
         run: |
           files_to_sign=(
             artifacts/opengrep_manylinux_binary_x86_64/opengrep_manylinux_x86
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -9,6 +9,7 @@ on:
 
 permissions:
   contents: write
+  id-token: write  # for Cosign keyless
 
 jobs:
 
@@ -124,7 +125,32 @@ jobs:
           pushd opengrep_windows_binary_x86; mv opengrep.exe opengrep_windows_x86.exe; popd
 
           popd
+      
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@v3
 
+      - name: Sign all binaries in artifacts
+        env:
+          COSIGN_EXPERIMENTAL: "1"
+        run: |
+          files_to_sign=(
+            artifacts/opengrep_manylinux_binary_x86_64/opengrep_manylinux_x86
+            artifacts/opengrep_manylinux_binary_aarch64/opengrep_manylinux_aarch64
+            artifacts/opengrep_musllinux_binary_x86_64/opengrep_musllinux_x86
+            artifacts/opengrep_musllinux_binary_aarch64/opengrep_musllinux_aarch64
+            artifacts/opengrep_osx_binary_arm64/opengrep_osx_arm64
+            artifacts/opengrep_osx_binary_x86/opengrep_osx_x86
+            artifacts/opengrep_windows_binary_x86/opengrep_windows_x86.exe)
+            for bin in "${files_to_sign[@]}"; do
+            echo "Signing $bin..."
+            cosign sign-blob \
+              --yes \
+              --output-signature "$bin.sig" \
+              --output-certificate "$bin.cert" \
+              "$bin"
+          done
+
+          
       - name: Create or Update Rolling Release
         uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2
         with:
@@ -136,11 +162,25 @@ jobs:
             This is a rolling release from `main`.
           files: |
             artifacts/opengrep_manylinux_binary_x86_64/opengrep_manylinux_x86
+            artifacts/opengrep_manylinux_binary_x86_64/opengrep_manylinux_x86.cert
+            artifacts/opengrep_manylinux_binary_x86_64/opengrep_manylinux_x86.sig
             artifacts/opengrep_manylinux_binary_aarch64/opengrep_manylinux_aarch64
+            artifacts/opengrep_manylinux_binary_aarch64/opengrep_manylinux_aarch64.cert
+            artifacts/opengrep_manylinux_binary_aarch64/opengrep_manylinux_aarch64.sig
             artifacts/opengrep_musllinux_binary_x86_64/opengrep_musllinux_x86
+            artifacts/opengrep_musllinux_binary_x86_64/opengrep_musllinux_x86.cert
+            artifacts/opengrep_musllinux_binary_x86_64/opengrep_musllinux_x86.sig
             artifacts/opengrep_musllinux_binary_aarch64/opengrep_musllinux_aarch64
+            artifacts/opengrep_musllinux_binary_aarch64/opengrep_musllinux_aarch64.cert
+            artifacts/opengrep_musllinux_binary_aarch64/opengrep_musllinux_aarch64.sig
             artifacts/opengrep_osx_binary_arm64/opengrep_osx_arm64
+            artifacts/opengrep_osx_binary_arm64/opengrep_osx_arm64.cert
+            artifacts/opengrep_osx_binary_arm64/opengrep_osx_arm64.sig
             artifacts/opengrep_osx_binary_x86/opengrep_osx_x86
-            artifacts/opengrep_windows_binary_x86/opengrep_windows_x86.exe
+            artifacts/opengrep_osx_binary_x86/opengrep_osx_x86.cert
+            artifacts/opengrep_osx_binary_x86/opengrep_osx_x86.sig
+            artifacts/opengrep_windows_binary_x86/opengrep_windows_x86.exe 
+            artifacts/opengrep_windows_binary_x86/opengrep_windows_x86.exe.cert 
+            artifacts/opengrep_windows_binary_x86/opengrep_windows_x86.exe.sig 
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/install.sh
+++ b/install.sh
@@ -1,37 +1,69 @@
-#!/usr/bin/env bash -e
-# Opengrep installation script 
+#!/usr/bin/env bash 
+set -euo pipefail
+# Opengrep installation script
 
 print_usage() {
-    echo "Usage: $0 [-v version] [-l] [-h]"
-    echo "  -v    Specify version to install (optional, default: latest)"
-    echo "  -l    List available versions (latest 3)"
-    echo "  -h    Show this help message"
+
+    echo "Usage:"
+    echo "install.sh  -v version   Specify version to install (optional, default: latest) warns if you cannot verify signatures"
+    echo "install.sh  -v version --verify-signatures   Specify version to install (optional, default: latest) failing if you cannot verify signatures "
+    echo "install.sh  -l    List available versions (latest 3)"
+    echo "install.sh  -h    Show this help message"
 }
 
 # Function to get available versions - already checked when running main
 get_available_versions() {
-    command -v curl >/dev/null 2>&1 || { echo >&2 "Required tool curl could not be found. Aborting."; exit 1; }
+    command -v curl > /dev/null 2>&1 || {
+        echo >&2 "Required tool curl could not be found. Aborting."
+        exit 1
+    }
     curl -s https://api.github.com/repos/opengrep/opengrep/releases | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
 }
 
 # Function to validate version
 validate_version() {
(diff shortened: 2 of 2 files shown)


install.sh
```diff
@@ -144,46 +185,105 @@ main () {
...
+if { { $verify_signatures || [[ -n "$version" ]] || $list; } && $help; } || { { $verify_signatures || [[ -n "$version" ]] || $help; } && $list; }; then
+    echo "❌ incorrect arguments:"
+    print_usage
+    exit 1
+
+fi
+
+if $verify_signatures && ! $has_cosign; then
+    echo "❌ cosign is required for -verify-signatures but not installed."
+    echo "Go to https:/github.com/sigstore/cosign to install or run without the -verify-signatures flag to install without verifying."
```
dimitris-m: `--verify-signatures` (missing `-`)

install.sh
```diff
@@ -144,46 +185,105 @@ main () {
...
+    exit 1
+elif ! $has_cosign; then
+    echo "⚠️ cosign is required for -verify-signatures but not installed. Skipping signature validation"
+    echo "Go to https:/github.com/sigstore/cosign to install."
```
dimitris-m: `--verify-signatures` (missing `-`)

install.sh
```diff
@@ -144,46 +185,105 @@ main () {
...
+if command -v cosign &> /dev/null; then
+    has_cosign=true
+else
+    has_cosign=false
+fi
+
+help=false
+list=false
+verify_signatures=false
+version=""
```
dimitris-m: We now have a mix of lowecase and uppercase variables in the script, maybe retain them all as uppercase?

corneliuhoffman: done

install.sh
corneliuhoffman: changed

install.sh
corneliuhoffman: changed

.github/workflows/rolling-release.yml
```diff
@@ -124,7 +125,32 @@ jobs:
           pushd opengrep_windows_binary_x86; mv opengrep.exe opengrep_windows_x86.exe; popd
 
           popd
+      
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@v3
 
+      - name: Sign all binaries in artifacts
+        env:
+          COSIGN_EXPERIMENTAL: "1"
```
dimitris-m: Probably not required

install.sh
```diff
@@ -1,37 +1,68 @@
...
 print_usage() {
-    echo "Usage: $0 [-v version] [-l] [-h]"
-    echo "  -v    Specify version to install (optional, default: latest)"
-    echo "  -l    List available versions (latest 3)"
-    echo "  -h    Show this help message"
+
+    echo "the possible uses are:"
+    echo "$0  -v version   Specify version to install \(optional, default: latest\) warns if you cannot verify signatures"
+    echo "$0  -v version --verify-signatures   Specify version to install \(optional, default: latest\) failing if you cannot verify signatures "
+    echo "$0  -l    List available versions \(latest 3\)"
```
dimitris-m: This is now printing `\(...\)` in a few places

install.sh
```diff
@@ -1,37 +1,68 @@
...
-# Opengrep installation script 
+# Opengrep installation script
 
 print_usage() {
-    echo "Usage: $0 [-v version] [-l] [-h]"
-    echo "  -v    Specify version to install (optional, default: latest)"
-    echo "  -l    List available versions (latest 3)"
-    echo "  -h    Show this help message"
+
+    echo "the possible uses are:"
```
dimitris-m: I would adapt the original: 

```
echo "Usage:
```

install.sh
dimitris-m: Also, `$0` is `bash` in the typical usage so not sure it's good to add.

For example:

```bash
❯ curl -fsSL https://raw.githubusercontent.com/opengrep/opengrep/969ee80ea422bd8652244371ed22ad91109eee8a/install.sh | bash -s -- -h

Cosign installed, including signature verification.
the possible uses are:
bash  -v version   Specify version to install \(optional, default: latest\) warns if you cannot verify signatures
bash  -v version --verify-signatures   Specify version to install \(optional, default: latest\) failing if you cannot verify signatures
bash  -l    List available versions \(latest 3\)
bash  -h    Show this help message
```

Also, the message: `Cosign installed, including signature verification.` should not be printed as first message, but only when relevant I think.

dimitris-m: For `$0` maybe we can hardcode `install.sh`?

install.sh
```diff
@@ -1,37 +1,68 @@
...
 }
 
 # Function to validate version
 validate_version() {
-    local version="$1"
-    local available_versions
-    available_versions=$(get_available_versions)
-    if echo "$available_versions" | grep -q "^$version$"; then
+    local VERSION="$1"
+    local AVAILBLE_VERSIONS
```
dimitris-m: Misprint `AVAILBLE`

install.sh
```diff
@@ -1,37 +1,68 @@
...
         exit 1
     fi
 }
+validate_signature() {
+    local P="$1"
+    if $HAS_COSIGN; then
 
-main () {
-    local VERSION="$1"
+        echo "🔍 Verifying signatures for ${P}/opengrep.cert"
```
dimitris-m: Icon

install.sh
```diff
@@ -1,37 +1,68 @@
...
+            --signature "${P}/opengrep.sig" \
+            --certificate-identity-regexp "https://github.com/opengrep/opengrep.+" \
+            --certificate-oidc-issuer "https://token.actions.githubusercontent.com" \
+            "${P}/opengrep"; then
+            echo "Signature valid."
+            exit 0
+        else
+            if [[ "$VERIFY_SIGNATURES" == true ]]; then
+                echo "Error: Signature validation error—deleting installed package"
+                rm -rf "${P}"
```
dimitris-m: Here `rm -f "${P}"` is enough

install.sh
```diff
@@ -1,37 +1,68 @@
...
+            --cert "$P/opengrep.cert" \
+            --signature "${P}/opengrep.sig" \
+            --certificate-identity-regexp "https://github.com/opengrep/opengrep.+" \
+            --certificate-oidc-issuer "https://token.actions.githubusercontent.com" \
+            "${P}/opengrep"; then
+            echo "Signature valid."
+            exit 0
+        else
+            if [[ "$VERIFY_SIGNATURES" == true ]]; then
+                echo "Error: Signature validation error—deleting installed package"
```
```
echo "Error: Signature validation error, deleting installed package ${P}"
```

dimitris-m: Is it installed or just downloaded?

install.sh
```diff
@@ -1,37 +1,68 @@
...
+                rm -rf "${P}"
+                exit 1
+            else
+                echo "Warning  Signature validation error; the package is still installed."
+                echo "If this was not intended, delete and rerun with --validate-version."
```
dimitris-m: --verify-signatures

install.sh
```diff
@@ -144,46 +185,84 @@ main () {
...
+if { { $VERIFY_SIGNATURES || [[ -n "$VERSION" ]] || $LIST; } && $HELP; } || { { $VERIFY_SIGNATURES || [[ -n "$VERSION" ]] || $HELP; } && $LIST; }; then
+    echo "Error: incorrect arguments:"
+    print_usage
+    exit 1
+
+fi
+
+if $VERIFY_SIGNATURES && ! $HAS_COSIGN; then
+    echo "Error: cosign is required for --verify-signatures but not installed."
+    echo "Go to https:/github.com/sigstore/cosign to install or run without the --verify-signatures flag to install without verifying."
```
dimitris-m: Bad URL `https:/github...`

install.sh
```diff
@@ -73,11 +107,12 @@ main () {
     URL="https://github.com/opengrep/opengrep/releases/download/${VERSION}/${DIST}"
     echo
     echo "*** Installing Opengrep ${VERSION} for ${OS} (${ARCH}) ***"
-    echo
+    echo "${URL}"}
```
dimitris-m: I would leave as it was (no URL) as it's unlikely to be very useful.
Also: there's a stray `}`.

install.sh
```diff
@@ -73,11 +107,12 @@ main () {
...
 
     # check if binary already exists
     if [ -f "${INST}/opengrep" ]; then
         echo "Destination binary ${INST}/opengrep already exists."
+        validate_signature "${INST}"
```
dimitris-m: Do we really want that?

install.sh
```diff
@@ -144,46 +185,84 @@ main () {
...
     exit 0
 fi
 
-shift $((OPTIND -1))
+shift $((OPTIND - 1))
 
 if [ -z "$VERSION" ]; then
     VERSION=$(get_available_versions | head -1)
-else
+    else
```
dimitris-m: indentation

install.sh
```diff
@@ -144,46 +185,84 @@ main () {
...
+if $VERIFY_SIGNATURES && ! $HAS_COSIGN; then
+    echo "Error: cosign is required for --verify-signatures but not installed."
+    echo "Go to https:/github.com/sigstore/cosign to install or run without the --verify-signatures flag to install without verifying."
+    exit 1
+elif ! $HAS_COSIGN; then
+    echo "Warning: cosign is required for --verify-signatures but not installed. Skipping signature validation"
+    echo "Go to https:/github.com/sigstore/cosign to install."
+else
+    echo "Cosign installed, including signature verification."
+fi
```
dimitris-m: I would remove that, leaving only the warning in case it's not there.

install.sh
dimitris-m: Maybe we could drop the `-f` too and let the user decide if they want to delete, as long as this does not leave a broken install.

By the way, what is the state of things if this fails and the user tries to re-install without the flag?

install.sh
```diff
@@ -1,37 +1,68 @@
 #!/usr/bin/env bash -e
-# Opengrep installation script 
+# Opengrep installation script
```
dimitris-m: I know this was like that, but it seems more portable to have:

```bash
#!/usr/bin/env bash
set -e
```

and maybe with: `set -euo pipefail` to be safer?

install.sh
corneliuhoffman: it reinstalls and warns of bad validation

corneliuhoffman: actually it should likely be rm -r since the file(s) are inside of $P

install.sh
corneliuhoffman: well downloaded but the script is called install so ..

install.sh
corneliuhoffman: what do you mean?

install.sh
corneliuhoffman: I think } closes main

install.sh
corneliuhoffman: done

install.sh
corneliuhoffman: done

install.sh
corneliuhoffman: done

install.sh
corneliuhoffman: removed

install.sh
corneliuhoffman: done

install.sh
corneliuhoffman: removed

install.sh
corneliuhoffman: fixed

install.sh
corneliuhoffman: done

install.sh
corneliuhoffman: hardcoded

install.sh
dimitris-m: Are you sure? Don't think so. Checked out the branch and got linter error.

install.sh
dimitris-m: Ah yes

install.sh
dimitris-m: Missing slash /

corneliuhoffman: Oh man ... fixed

install.sh
corneliuhoffman: and unescaped

PR #324, opened 2025-07-04
Ruby local global bug

### LINKED ISSUE #323: ruby local global bug
**Describe the bug**
taint is not found in:
```ruby
 def opts(data) = Array(data).join(" ") # process options

  def test_calls4(user_input)
    opts = [user_input, "-c"] # local var
    # ruleid: test
    system("ls #{opts}")
  end
```
due to the same name for opts.

corneliuhoffman:
fixes #323 

 In the AS_to_IL there was a short hack to allow for ruby calling function without brackets and without arguments. The issue was that it did not consider local variable.

### DIFF
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -635,7 +635,7 @@ and expr_aux env ?(void = false) g_expr =
                 tok,
                 G.FN
                   (G.Id
-                    (("concat", _), { G.id_resolved = { contents = None }; _ }))
+                     (("concat", _), { G.id_resolved = { contents = None }; _ }))
               );
           _;
         },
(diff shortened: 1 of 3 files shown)


src/analyzing/AST_to_IL.ml
```diff
@@ -721,9 +721,13 @@ and expr_aux env ?(void = false) g_expr =
...
         match lval with
-        | { base = Var { ident; _ }; _ }
+        | { base = Var { ident; id_info; _ }; _ }
           when env.lang =*= Lang.Ruby
-               && IdentSet.mem (H.str_of_ident ident) env.ctx.entity_names ->
+               && IdentSet.mem (H.str_of_ident ident) env.ctx.entity_names
+               &&
+               match !(id_info.id_resolved) with
+               | None -> true
+               | _ -> false ->
```
dimitris-m: why not use `Option.is_none` here?

dimitris-m: Also, the last conjunct could become second (simple re-order) as it seems more performant when short-circuited?

dimitris-m (approved):
LGTM, nice.

Left some suggestions.

src/analyzing/AST_to_IL.ml
corneliuhoffman: done, was just annoyed by the nanny "=" an saw red :)

dimitris-m: > Also, the last conjunct could become second (simple re-order) as it seems more performant when short-circuited?

but confirm if this is the case, I could be wrong.

PR #325, opened 2025-07-04
Make tests invariant to version

dimitris-m:
### Changes

Ensures we don't need to manually edit test snapshots when bumping the version in releases.


maciejpirog (approved):
Good!

PR #330, opened 2025-07-06
fix: detect musl in install script

securityPirate:
Issue: 
The ldd --version check failed to reliably detect musl on Alpine.

Change:
 A direct verification for `/lib/ld-musl-*.so.*`. The path based check for musl presence when ldd --version output is ambiguous.

Validate:
Create a test.Dockerfile for testing
```
FROM alpine:latest
RUN apk add cosign curl
COPY install.sh .
RUN sh ./install.sh
```
Run
`Docker build . -f test.Dockerfile --no-cache --progress=plain`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -112,7 +112,7 @@ main() {
 
     # check and set "os_arch"
     if [ "$OS" = "Linux" ]; then
-        if ldd --version 2> /dev/null | grep -qi musl || [ -f /lib/ld-musl-*.so.* ]; then
+        if ldd /bin/sh 2>&1 | grep -qi musl; then
             if [ "$ARCH" = "x86_64" ] || [ "$ARCH" = "amd64" ]; then
                 DIST="opengrep_musllinux_x86"
             elif [ "$ARCH" = "aarch64" ] || [ "$ARCH" = "arm64" ]; then
--- a/install.sh
+++ b/install.sh
@@ -112,7 +112,7 @@ main() {
 
     # check and set "os_arch"
     if [ "$OS" = "Linux" ]; then
-        if ldd --version 2>&1 | grep -qi musl; then
+        if ldd --version 2> /dev/null | grep -qi musl || [ -f /lib/ld-musl-*.so.* ]; then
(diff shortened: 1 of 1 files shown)


dimitris-m (changes requested):
This is not accurate:

```
/ # if ldd --version 2>&1 | grep -qi 'musl'; then
>     echo "Using musl"
> fi
Using musl
```

but the exit code is 1 -- this is why the script fails.

Simply replace `--version` with `/bin/sh`.

The same check will be ok and Opengrep will install.

install.sh
```diff
@@ -151,7 +151,7 @@ main() {
         fi
     else
         echo
-        echo "*** Installing Opengrep ${VERSION} for ${OS} (${ARCH}) ***"
+        echo "*** Installing ${DIST} ${VERSION} for ${OS} (${ARCH}) ***"
```
dimitris-m: Please undo this change, it will confuse users.

install.sh
```diff
@@ -112,7 +112,7 @@ main() {
 
     # check and set "os_arch"
     if [ "$OS" = "Linux" ]; then
-        if ldd --version 2>&1 | grep -qi musl; then
+        if ldd --version 2> /dev/null | grep -qi musl || [ -f /lib/ld-musl-*.so.* ]; then
```
```suggestion
        if ldd /bin/sh 2>&1 | grep -qi musl; then
```

install.sh
dimitris-m:
```suggestion
        echo "*** Installing Opengrep ${VERSION} for ${OS} (${ARCH}) ***"
```

dimitris-m:
To be more accurate: 

```bash
/ # ldd --version 2>&1
musl libc (aarch64)
Version 1.2.5
Dynamic Program Loader
Usage: /lib/ld-musl-aarch64.so.1 [options] [--] pathname
/ # echo "$?"
1
/ # ldd /bin/sh 2>&1
	/lib/ld-musl-aarch64.so.1 (0xffffbbb0c000)
	libc.musl-aarch64.so.1 => /lib/ld-musl-aarch64.so.1 (0xffffbbb0c000)
/ # echo "$?"
0
```

and the script starts with:

```
set -euo pipefail
```

Issue #312, opened 2025-06-30
Kotlin: Taint analysis does not propagate through the Elvis operator (`?:`)


fxbru:
### Description

Hello Opengrep Team,

I've identified a limitation in the taint analysis engine for the Kotlin language. When a variable is assigned a value from a taint source using the Elvis operator (`?:`), the taint is not propagated correctly. This can lead to false negatives for rules relying on taint mode.

### Steps to Reproduce

1.  **Rule:** A basic taint-tracking rule to find format string injections.
    ```yaml
    rules:
    - id: format-string-injection
      mode: taint
      pattern-sources:
        - pattern: $INTENT.getStringExtra(...)
      pattern-sinks:
        - pattern: String.format($FORMAT, ...)
      message: Format String Injection
      languages: [kotlin]
      severity: ERROR
    ```

2.  **Kotlin Code:** A file containing two functions. The first uses the Elvis operator, and the second uses a direct assignment.
    ```kotlin
    // test.kt
    import android.content.Intent

    class MyActivity {
        // Taint flow is NOT detected in this function
        fun processIntentData1(intent: Intent) {
            val formatString = intent.getStringExtra("format_string") ?: return
            // The 'formatString' variable should be considered tainted here.
            String.format(formatString, "argument")
        }

        // Taint flow IS detected correctly in this function
        fun processIntentData2(intent: Intent) {
            val formatString = intent.getStringExtra("format_string")
            String.format(formatString, "argument")
        }
    }
    ```

### Expected Behavior

When running Opengrep with the rule and code above, the taint analysis should find a path from the source (`getStringExtra`) to the sink (`String.format`) in **both** `processIntentData1` and `processIntentData2`. Two findings should be reported.

### Actual Behavior

Opengrep only reports a finding for `processIntentData2`. The taint flow in `processIntentData1` is missed due to the use of the Elvis operator, resulting in a false negative.

---

This appears to be a limitation in how the analysis engine models this specific Kotlin operator. If the team agrees this is a valid issue, I would be happy to prepare a Pull Request to address it.

Thank you for your work on this great tool.

PR #331, opened 2025-07-07
fix(kotlin): Enable taint tracking through the Elvis operator (?:)

fxbru:
#312

**Description**
This PR fixes a false negative in the Kotlin taint analysis engine where taint flow was incorrectly terminated when encountering the Elvis operator (`?:`).

The root cause was that the generic AST-to-IL converter did not have a specific rule for the Elvis operator. It was treated as a generic binary operator, and the engine did not model the conditional data flow where the result could come from either the left-hand side or the right-hand side.

This fix addresses the issue by updating `src/analyzing/AST_to_IL.ml`. It introduces special handling for the `G.Op(Elvis, ...)` node, transforming it into an if/else structure in the Intermediate Language. This correctly models that the resulting value (and its taint) can flow from either branch of the expression, allowing the taint engine to follow the path correctly.

**Testing and Validation**
The fix was validated using a simple taint rule and a test case demonstrating the issue.

Rule:
```
rules:
- id: elvis-operator-taint-test
  mode: taint
  pattern-sources:
    - pattern: $INTENT.getStringExtra(...)
  pattern-sinks:
    - pattern: String.format($FORMAT, ...)
  languages: [kotlin]
```

Test Code:
```
// Taint flow was previously NOT detected in this function
fun processIntentWithElvis(intent: Intent) {
    val formatString = intent.getStringExtra("format_string") ?: return
    String.format(formatString, "argument")
}

// Taint flow was already correctly detected in this function
fun processIntentDirect(intent: Intent) {
    val formatString = intent.getStringExtra("format_string")
    String.format(formatString, "argument")
}
```

**Behavior Before This PR**
Findings: 1 (Only `processIntentDirect` was detected).

**Behavior After This PR**
Findings: 2 (Both `processIntentWithElvis` and `processIntentDirect` are now correctly detected).

This change makes the Kotlin taint analysis more accurate and robust against common language idioms.

### DIFF
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -570,68 +570,25 @@ and expr_aux env ?(void = false) g_expr =
       expr_lazy_op env op tok arg0 args eorig
   (* args_with_pre_stmts *)
   | G.Call ({ e = G.IdSpecial (G.Op op, tok); _ }, args) -> (
-    match op with
-    | G.Elvis when env.lang =*= Lang.Kotlin -> (
-        match Tok.unbracket args with
-        | [ G.Arg lhs_gen; G.Arg rhs_gen ] ->
-            begin
-              (* This translates 'lhs ?: rhs' into the IL equivalent of:
-               *
-               * if (lhs != null) {
-               * result = lhs;
-               * } else {
-               * result = rhs;
-               * }
-               *
-               * This structure allows the dataflow engine to see that 'result'
-               * can be tainted by either 'lhs' or 'rhs'.
-              *)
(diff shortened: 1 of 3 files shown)


fxbru:
Thanks for your feedback @corneliuhoffman.

I'll work on implementing this change entirely in Parse_kotlin_tree_sitter.ml.

dimitris-m:
> Thanks for your feedback @corneliuhoffman.
> 
> I'll work on implementing this change entirely in Parse_kotlin_tree_sitter.ml.

And please can you rebase on top on main, we have a lot of changes and it's nice to have a linear-looking history.
(Both PRs).

Issue #313, opened 2025-06-30
Kotlin: Taint analysis fails to propagate through `.let` scope function


fxbru:
### Description

Hello Opengrep Team,

I've encountered a case where taint analysis for Kotlin does not seem to track dataflow through the `.let` scope function. This results in false negatives for rules that rely on taint mode, as the engine loses the taint trail when a variable is passed into a `.let` block.

### Steps to Reproduce

1.  **Rule:** A minimal taint-tracking rule.
    ```yaml
    rules:
    - id: scope-function-taint-test
      mode: taint
      pattern-sources:
        - pattern: mySource(...)
      pattern-sinks:
        - pattern: mySink(...)
      message: Scope functions taint analysis test
      languages: [kotlin]
      severity: INFO
    ```

2.  **Kotlin Code:** A test file demonstrating the issue with both implicit `it` and explicitly named lambda variables.
    ```kotlin
    // test.kt
    fun mySource(): String = "tainted"
    fun mySink(data: String) { /* sink */ }

    // Case 1: Taint flow is NOT detected
    fun testTaintWithImplicitIt() {
        val taintedVar = mySource()
        taintedVar.let {
            mySink(it)
        }
    }

    // Case 2: Taint flow is NOT detected
    fun testTaintWithExplicitVar() {
        val taintedVar = mySource()
        taintedVar.let { v ->
            mySink(v)
        }
    }

    // Case 3: Taint flow IS detected (Control)
    fun testTaintWithDirectCall() {
        val taintedVar = mySource()
        mySink(taintedVar)
    }
    ```

### Expected Behavior

Opengrep should report **three** findings, identifying a taint flow from `mySource()` to `mySink()` in all three test functions.

### Actual Behavior

Opengrep only reports **one** finding, for `testTaintWithDirectCall`. The taint flow is not detected in the two cases that use the `.let` scope function.

---

This limitation seems to apply to other scope functions as well. If the team agrees this is a valid issue that needs addressing, I would be happy to contribute a Pull Request to help resolve it.

Thank you for your time and for this excellent tool.

PR #332, opened 2025-07-07
fix(kotlin): Enable taint tracking through scope functions (`let`, `also`, `use`, `takeIf`, `takeUnless`)

fxbru:
#313

### Description

This PR resolves a major source of false negatives in the Kotlin taint analysis engine where taint flow would be lost when passing through common scope functions like `let`, `also`, `use`, `takeIf` and `takeUnless`.

The issue was caused by two distinct problems in the analysis pipeline:

1.  **Missing Implicit Parameter:** The Kotlin parser did not generate an AST node for the implicit `it` parameter in single-argument lambdas, making it impossible for the engine to reason about it.

2.  **Missing Dataflow Edge:** The AST-to-IL converter lacked the semantic understanding to connect the receiver of a scope function call (e.g., `taintedVar` in `taintedVar.let { ... }`) to the lambda's parameter. This broke the dataflow graph.

This PR implements a two-part fix to address these issues comprehensively:

* **Parser Enhancement (`Parse_kotlin_tree_sitter.ml`):** The parser is updated to generate a synthetic `it` parameter in the AST for lambdas where it is implicit. This ensures the AST accurately reflects Kotlin's semantics.

* **IL Converter Update (`AST_to_IL.ml`):** The IL converter is enhanced to recognize calls to `let`, `also`, `use`, `takeIf` and `takeUnless`. It now rewrites the AST in-flight to inject an explicit `param = receiver` assignment at the beginning of the lambda's body.

Together, these changes create the necessary dataflow edge for the taint engine to correctly follow taint from a receiver object into the body of these critical scope functions.

### Testing and Validation

The fix was validated using a simple taint rule and a test case covering implicit `it`, explicit parameters, and multiple scope functions.

**Rule:**

```yaml
rules:
- id: scope-function-taint-test
  mode: taint
  pattern-sources:
    - pattern: mySource(...)
  pattern-sinks:
    - pattern: mySink(...)
  languages: [kotlin]
```

**Test Code:**
```kotlin
// Taint flow was previously NOT detected in these functions
fun testTaintWithLet(taintedVar: String) {
    taintedVar.let {
        mySink(it) // Taint should flow to 'it'
    }
}

fun testTaintWithAlso(taintedVar: String) {
    taintedVar.also { v ->
        mySink(v) // Taint should flow to 'v'
    }
}

fun testTaintWithUse(taintedStream: java.io.InputStream) {
    taintedStream.use { s ->
        mySink(s.toString()) // Taint should flow to 's'
    }
}

// Taint flow was already correctly detected here
fun testTaintWithDirectCall(taintedVar: String) {
    mySink(taintedVar)
}
```

**Behavior Before This PR**
Findings: 1 (Only `testTaintWithDirectCall` was detected).

**Behavior After This PR**
Findings: 4 (All four functions are now correctly detected).

This change significantly improves the accuracy of Kotlin taint analysis and establishes a clear pattern for adding support for other scope functions in the future.

### DIFF
--- a/languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
+++ b/languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
@@ -1424,10 +1424,15 @@ and lambda_literal (env : env) ((v1, v2, v3, v4) : CST.lambda_literal) =
         (* use this to delimit the Block below. *)
         let v2 = token env v2 (* "->" *) in
         (v1, v2)
-    (* note that even without parameters, 'it' can be used to
-     * represent an anonymous parameter.
-     *)
-    | None -> ([], v1)
+    | None ->
+        (*
+         * If a lambda has no explicit parameters, Kotlin provides an implicit 'it'.
+         * We create a synthetic parameter in the AST to represent it, which allows
+         * the dataflow engine to track taint into the lambda body.
+        *)
+        let fake_it_tok = Tok.fake_tok v1 "it" in
+        let it_param = G.Param (G.param_of_id ("it", fake_it_tok)) in
+        ([it_param], v1)
   in
   let v3 =
     match v3 with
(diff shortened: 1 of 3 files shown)


corneliuhoffman:
Here the change is a bit more subtle but I also think it belongs in the Parse_kotlin_tree_sitter.ml.

dimitris-m:
We want to improve Kotlin as you suggest, but we need to determine the best place to put those changes. 

Adding syntax to the parse tree is not ideal.

Please leave the PR open and we will return to this!

Thanks

dimitris-m:
We have decided to merge as is -- after rebasing.

PR #334, opened 2025-07-07
fix(kotlin): Enable taint tracking through the Elvis operator (?:)

fxbru:
This PR is a duplicate of #331 (closed), to resolve issue #312.
Changes are now narrowed to Parse_kotlin_tree_sitter.ml, and AST_to_IL remains untouched.

**Description**
This PR fixes a false negative in the Kotlin taint analysis engine where taint flow was incorrectly terminated when encountering the Elvis operator (`?:`).

To do this, we have to "desugar" the Elvis operator—that is, convert it from its special syntax (?:) into a more fundamental language construct (an if/else expression) at the earliest possible stage, which is inside Parse_kotlin_tree_sitter.ml.

**Testing and Validation**
The fix was validated using a simple taint rule and a test case demonstrating the issue.

Rule:
```
rules:
- id: elvis-operator-taint-test
  mode: taint
  pattern-sources:
    - pattern: $INTENT.getStringExtra(...)
  pattern-sinks:
    - pattern: String.format($FORMAT, ...)
  languages: [kotlin]
```

Test Code:
```
// Taint flow was previously NOT detected in this function
fun processIntentWithElvis(intent: Intent) {
    val formatString = intent.getStringExtra("format_string") ?: return
    String.format(formatString, "argument")
}

// Taint flow was already correctly detected in this function
fun processIntentDirect(intent: Intent) {
    val formatString = intent.getStringExtra("format_string")
    String.format(formatString, "argument")
}
```

**Behavior Before This PR**
Findings: 1 (Only `processIntentDirect` was detected).

**Behavior After This PR**
Findings: 2 (Both `processIntentWithElvis` and `processIntentDirect` are now correctly detected).

This change makes the Kotlin taint analysis more accurate and robust against common language idioms.

### DIFF
--- a/languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
+++ b/languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
@@ -1450,15 +1450,7 @@ and lambda_literal (env : env) ((v1, v2, v3, v4) : CST.lambda_literal) =
         (* use this to delimit the Block below. *)
         let v2 = token env v2 (* "->" *) in
         (v1, v2)
-    | None ->
-        (*
-         * If a lambda has no explicit parameters, Kotlin provides an implicit 'it'.
-         * We create a synthetic parameter in the AST to represent it, which allows
-         * the dataflow engine to track taint into the lambda body.
-        *)
-        let fake_it_tok = Tok.fake_tok v1 "it" in
-        let it_param = G.Param (G.param_of_id ("it", fake_it_tok)) in
-        ([it_param], v1)
+    | None -> ([], v1)
   in
   let v3 =
     match v3 with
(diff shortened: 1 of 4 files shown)


dimitris-m:
@fxbru, thanks for your contribution! 

We reviewed internally and we prefer the solution of #331 after all, that is, the solution in `AST_to_IL.ml`.

The reason for this is that we don't want to identify Elvis with conditionals, in all contexts.

So please can you cherry-pick or just copy the solution from #331 in a new PR, noting:

- please ensure your solution is rebased on main and keep only the solution (lines in 573 tp 634 in https://github.com/opengrep/opengrep/pull/331/commits/9f66b4525c7bcbe583fd6225024c68932d90b119 plus the tests); 
- see the comments that @maciejpirog will add now on how to improve your solution, because we believe there is a small mistake.

Thanks

dimitris-m (changes requested):
@fxbru, thanks for your contribution! 

We reviewed internally and we prefer the solution of #331 after all, that is, the solution in `AST_to_IL.ml`.

The reason for this is that we don't want to identify Elvis with conditionals, in all contexts.

So please can you cherry-pick or just copy the solution from #331 in a new PR, noting:

- please ensure your solution is rebased on main and keep only the solution (lines in 573 tp 634 in https://github.com/opengrep/opengrep/pull/331/commits/9f66b4525c7bcbe583fd6225024c68932d90b119 plus the tests); 
- see the comments that @maciejpirog will add now on how to improve your solution, because we believe there is a small mistake.

Thanks

PR #337, opened 2025-07-08
Add support for Apex

maciejpirog:
Add support for Salesforce's Apex (https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_intro_what_is_apex.htm)

- Supports `.cls` and `.trigger` file extensions.

- This PR includes a basic support for query expressions, which includes deep expressions and ellipses.

---

The support for SOQL and SOSL queries is a bit limited at the moment. One way to deal with them is to use deep expressions. For example, in a taint rule, one can write a rules like:

```yaml
    message: found potential injection $P
    pattern-sources:
      - pattern: unsafeGetData()
    pattern-sinks:
      - pattern: |
          [<... $P ...>]
```

which matches code like

```apex
String userData = unsafeGetData();
...
[SELECT c1, c2 FROM t1 WHERE c1 = :userData AND c2 = 100 LIMIT 10];
```


src/osemgrep/language_server/scan_helpers/Scan_helpers.ml
```diff
@@ -88,56 +88,14 @@ let run_semgrep ?(targets : Fpath.t list option) ?rules ?git_ref
...
-                       pro binary available. Running with the OSS engine \
-                       instead.");
-              Core_runner.mk_core_run_for_osemgrep (Core_scan.scan session.caps)
+            if pro_intrafile then
+              Logs.warn (fun m ->
+                  m
+                    "Pro intrafile is enabled, but the pro engine is not \
+                      available, as the user is not logged in, or there is no \
+                      pro binary available. Running with the OSS engine \
+                      instead.");
```
dimitris-m, line 97: Maybe delete this too?

src/parsing/Parsing_plugin.ml
```diff
@@ -129,11 +129,6 @@ let make ?(optional = false) lang =
 (* Plugins *)
 (*****************************************************************************)
 
-module Apex = struct
-  let is_optional, register_parsers, is_available, parse_pattern, parse_target =
-    make Lang.Apex
-end
-
```
dimitris-m, line 136: nit: I guess there are no more plugins? Something to remove in the future?

languages/apex/generic/Parse_apex_tree_sitter.ml
```diff
@@ -0,0 +1,7343 @@
...
+  | `Pat_last_year x
+  | `Pat_next_year x
+  | `Pat_this_fiscal_quar x
+  | `Pat_last_fiscal_quar x
+  | `Pat_next_fiscal_quar x
+  | `Pat_this_fiscal_year x
+  | `Pat_last_fiscal_year x
+  | `Pat_next_fiscal_year x ->
+      let v = str env x in
+      R.Token v
```
dimitris-m, line 263: I wonder why we use `R.token` instead of proper token: `let token = H.token` as in csharp.

languages/apex/generic/Parse_apex_tree_sitter.ml
```diff
@@ -0,0 +1,7343 @@
...
+
+(* NEW *)
+and accessor_list (env : env) ((v1, v2, v3) : CST.accessor_list)
+    : (G.attribute list * G.ident * G.function_body) list bracket =
+  let v1 = (* "{" *) token_ env v1 in
+  let v2 =
+    List.map (accessor_declaration env) v2
+  in
+  let v3 = (* "}" *) token_ env v3 in
+  v1, v2, v3
```
dimitris-m, line 752: not important: 
in some places we have tuples with explicit parentheses, in others (here) we don't.

dimitris-m (approved):
I recommend to add one commit that removes the `OLD` stuff that have been translated in a satisfactory way, to avoid duplication with `Boilerplate.ml` and to reduce the size of the file.

And as discussed we should open issues for the stuff that has not been implemented yet.

Other than that... 🚀

languages/apex/generic/Parse_apex_tree_sitter.ml
dimitris-m: I understand now there is probably a good reason.

maciejpirog: Because they serve different purposes. `H.token` takes a CST token to `AST_generic.tok`, while `R.Token` is a constructor of raw tree. In particular `AST_generic.tok` by itself is not a proper expression or statement (some constructors in the generic syntax want tokens, usually as brackets, but some don't), but `R.Token` is a proper constructor of syntax (via `AST_generic.RawTree (R.Token ...) : expr_kind`).

`H.token` is used to translate tokens from CST to AST generic's tokens. Not to get confused with the "OLD" functions, it is called `token_` (with underscore).

languages/apex/generic/Parse_apex_tree_sitter.ml
```diff
@@ -5980,41 +5982,50 @@ and sobject_return (env : env) ((v1, v2) : CST.sobject_return) : raw =
     match v2 with
     | Some (v1, v2, v3, v4, v5, v6, v7, v8) ->
         let v1 = (* "(" *) str env v1 in
-        let v2 = selected_fields env v2 in
+        let v2 =
+          R.List
+            (List.map
+               (fun e -> R.Any (G.E e))
+               (selected_fields env v2))
+        in
```
dimitris-m, line 2552: A bit of a naive question: if that stuff is inside Raw (last line of function), does this have any effect?

languages/apex/generic/Parse_apex_tree_sitter.ml
maciejpirog: fixed now

dimitris-m (approved):
Epic work

PR #341, opened 2025-07-10
fix: flake.nix

spotdemo4:
Currently, the nix flake is completely broken. This PR fixes it and cleans it up a bit.
- Bumped `ocaml-base-compiler` 4.14.2 -> 5.3.0
- Bumped `mirage-runtime` 4.4.2 -> 4.9.0
- Replaced `nixpkgs-fmt` -> `alejandra`
- Replaced most instances of semgrep with opengrep
- Removed all of the submodule warnings/errors, instead just set `self.submodules = true`

```zsh
> nix build
trace: [opam-nix] mirage-random: mirage-random is deprecated
> ls result/bin
opengrep  pyopengrep
```

Unfortunately, the biggest culprit is memprof-limits. I don't know of a way of getting opam-nix to work if memprof-limits depends on [install-memprof-limits-dev.sh](https://github.com/opengrep/opengrep/blob/main/scripts/install-memprof-limits-dev.sh) getting run first to create the pin. Instead, this PR creates opam.template files that pin memprof-limits to the equivalent hash using [pin-depends](https://ocaml.org/docs/managing-dependencies#adding-a-git-dependency-to-a-dune-project-file). If there is a better way of accomplishing this let me know.

### DIFF
--- a/dune
+++ b/dune
@@ -34,4 +34,5 @@
   languages
   src
   tools
+  opam
 )
--- a/Makefile
+++ b/Makefile
@@ -250,7 +250,6 @@ install-deps-for-semgrep-core:
 	cd libs/ocaml-tree-sitter-core \
 	&& ./configure \
 	&& ./scripts/install-tree-sitter-lib
-	./scripts/install-memprof-limits-dev.sh
 	make install-opam-deps
 
 # Install OCaml dependencies (globally) from *.opam files.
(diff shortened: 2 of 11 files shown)


dimitris-m:
Would be good to have nix working but our opam files are auto-generated from dune-project, and this should remain the case.

So for now, we cannot merge this as is. Happy to also think of some solution...

dune
```diff
@@ -34,4 +34,5 @@
   languages
   src
   tools
+  opam
```
dimitris-m, line 37: does that make a difference? there are no dune files there.

dimitris-m:
> > Would be good to have nix working but our opam files are auto-generated from dune-project, and this should remain the case.
> > So for now, we cannot merge this as is. Happy to also think of some solution...
> 
> With the two opam.template files:
> 
> https://github.com/opengrep/opengrep/blob/b1dded4f189b06c404a197316d645d769531a392/opam/semgrep.opam.template#L1-L3
> 
> 
> https://github.com/opengrep/opengrep/blob/b1dded4f189b06c404a197316d645d769531a392/opam/process_limits.opam.template#L1-L3
> 
> the corresponding semgrep.opam and process_limits.opam files are still auto-generated from dune-project. I never manually touched a .opam file, I just ran `dune build`. The template files just append pin-depends to the end of the generated files:
> 
> https://github.com/opengrep/opengrep/blob/b1dded4f189b06c404a197316d645d769531a392/opam/semgrep.opam#L98-L100

If that is the case, great.

I'll review in more detail.

dune
dimitris-m: maybe it saves from having to do `dune build opam/*.opam`

dimitris-m: which we do manually right now

dimitris-m:
Btw, if you need to update your branch please rebase, don't merge, else I might have to squash your commits.

dune
spotdemo4: I needed to do this in order to get the template files to work (which was a pita to figure out). I think it's because dune doesn't check the ./opam directory for .opam.template files by default

dimitris-m:
I'll try to test this this week and we should aim to merge it

dimitris-m:
Closing since we already merged #668, but thanks for the contribution

PR #349, opened 2025-07-11
feat: release opengrep-core

### LINKED ISSUE #347: Release static binaries for opengrep-core
In order to get this packaged in [nixpkgs](https://github.com/NixOS/nixpkgs), we'll need a link to some static binaries for opengrep-core. Otherwise we'd need to put every dependency on nixpkgs to build the core from scratch, which is too much to maintain. 

Semgrep does this by extracting semgrep-core from the Pypi release: [semgrep-core.nix](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/python-modules/semgrep/semgrep-core.nix#L23-L30)

This would work, but it'd be easier if this repo just released the static binaries with all of the other assets in the github release for each version update:
- opengrep-core_linux_x86.tar.gz
- opengrep-core_linux_aarch64.tar.gz
- opengrep-core_osx_x86.tar.gz
- opengrep-core_osx_aarch64.tar.gz

spotdemo4:
closes #347

example of a successful release: https://github.com/spotdemo4/opengrep/releases/tag/v1.6.0

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -107,6 +107,9 @@ jobs:
           tar xzf ./opengrep-osx-arm64/artifacts.tgz --strip=1
           tar czf opengrep-core_osx_aarch64.tar.gz opengrep-core
 
+          tar xzf ./opengrep-core-and-dependent-libs-w64-artifact/artifacts.tgz
+          zip -r -j opengrep-core_windows_x86.zip artifacts
+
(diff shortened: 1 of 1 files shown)


dimitris-m (commented):
Thanks for that!
Let's add the windows binary too, if you have the time?

spotdemo4:
> Thanks for that! Let's add the windows binary too, if you have the time?

done :+1:

Issue #335, opened 2025-07-07
Non-deterministic abstract_content and fingerprints across runs


aayachnes:
**Describe the bug**
I found a bug in opengrep (and probably semgrep) where the `fingerprint` and `abstract_content` fields of results are non-deterministic across multiple runs of opengrep (v1.5.0, though probably also in earlier versions). This poses a problem for customers that rely on the `fingerprint` field to be unchanging across runs for the same file.

Some important things to note:
1. The reproduction example is for Kotlin, but the issue was also found in Ruby and Java files.
2. If we only scan a single file with opengrep, this issue is far less likely to occur. If scanning multiple files, this issue is extremely likely to occur. There may be some kind of race condition/non-determinism when multiple files are in the mix.

**To Reproduce**
1. Download the attached zip file. Ensure you have `opengrep v1.5.0` installed.
2. Run `opengrep.sh`; three JSON output files should be created (`opengrep1.json`, `opengrep2.json`, `opengrep3.json`).
3. Run `md5sum opengrep*.json`; most likely one or more files have different checksums. Run `diff` on two files with different checksums. You should see a result like this:

```
162c162
<             "abstract_content": "MessageDigest MessageDigest MessageDigest"
---
>             "abstract_content": "MessageDigest"
209c209
<         "fingerprint": "4f1c6842d9905b356cd15c7f08db2205382ed6ce1c8071986540557d8f3edead428f466b04192a08f9df03cd8660c6976969b574ab2bf04b4692f364b722a598_0",
---
>         "fingerprint": "4ac1c6d3f696697db0851b20189796f612cf9f55884544ce08d84e02196377ca18357feaedc5da9dfa7ff40529c376bc3dae9c62d0f5ee4a5a279d5030237cf4_0",
```

4. If all files have the same checksum, run `opengrep.sh` again.

**Expected behavior**
- I expect the opengrep results for the same file and rule to be identical.

**Screenshots**

**What is the priority of the bug to you?**

- [X] P0: blocking your adoption of Opengrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
- Mac OSX, Opengrep v1.5.0

**Use case**
- This bug is blocking adoption as we rely on the fingerprint to stay the same for the same file and rule.

maciejpirog:
Possibly not a concurrency issue. Here's a run on my machine (note the `-j 1` in the script):
```bash
$ cat opengrep.sh                    
[...]

PR #356, opened 2025-07-17
For variables, calculate the range in the `extra` section using the original token

maciejpirog:
Back to #335:

In the example provided in #335 by @aayachnes, we could observe non-deterministically `MessageDigest` or `MessageDigest MessageDigest MessageDigest` as the abstract content of the metavariable.

The reason why we saw the triple token was that constant propagation (or maybe name resolution?) was linking the occurrence of the matched variable with the import

```
java.security.MessageDigest
```

and tried to print it as the value of the metavariable.

For some reason, however, the identifier `java.security.MessageDigest` is treated as `DotAccess`, in which each segment is given the token of the occurrence of the variable (**!!!**), hence it is printed out as triple `MessageDigest` (dots in between become fake tokens, because `DottedIdentifier` does not hold dot tokens, while `DotAccess` does, for some weird reason).

This PR does not solve the nondet problem (sometimes the constant propagation does not seem to link to the import, but it is not clear why).

### DIFF
--- a/src/reporting/Metavar_replacement.ml
+++ b/src/reporting/Metavar_replacement.ml
@@ -30,11 +30,18 @@ let metavar_string_of_any any =
           x = 1; y = x + 1; ...
      we have y = 2 but there is no source location for 2.
      Handle such cases *)
-  any |> AST_generic_helpers.ii_of_any
-  |> List.filter Tok.is_origintok
-  |> List.sort_uniq Tok.compare_pos
-  |> List_.map Tok.content_of_tok
-  |> Core_text_output.join_with_space_if_needed
+  (* FIXME: If the match is a variable, extract the variable, not the
+     propagated value / definition assigned by name resolution. This is a
+     temporary fix for the nondeterministic behaviour bug (issue #335) *)
+  match any with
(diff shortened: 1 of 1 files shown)


src/reporting/Metavar_replacement.ml
```diff
@@ -30,11 +30,18 @@ let metavar_string_of_any any =
...
+  (* FIXME: If the match is a variable, extract the variable, not the
+     propagated value / definition assigned by name resolution. This is a
+     temporary fix for the nondeterministic behaviour bug (issue #335) *)
+  match any with
+  | G.E { e_range = Some (l1, l2); _ } when Int.equal l1.pos.bytepos l2.pos.bytepos ->
+      l1.str
+  | _ ->
+      any |> AST_generic_helpers.ii_of_any
+      |> List.filter Tok.is_origintok
+      |> List.sort Tok.compare_pos
```
maciejpirog, line 42: Note that `sort_uniq` was changed back to `sort`

dimitris-m (approved):
Ok, let's merge this in.

I confirm that only the 1st issue remains fixed (but your solution is cleaner).

The second issue still happens, ie, we still get nondeterminictic fingerprints in that case.

PR #359, opened 2025-07-21
Add new rule option `max-match-per-file`

dimitris-m:
## Changes

- Revert the global `--max-match-per-file` parameter's behaviour: since this check happens before de-duplication, we retain the original behaviour of Semgrep in which no matches are retained for such targets.

- Add a `max-match-per-file`option to rule options, which allows to report deduplicated matches only up to that limit. Useful with rules that typically result in many matches per file.

How to use the new rule option: 

``` yaml
rules:
- id: some-rule-with-many-matches-per-file
  options:
    max_match_per_file: 1 # any positive integer.
```

### DIFF
--- a/src/core/Core_match.ml
+++ b/src/core/Core_match.ml
@@ -153,8 +153,34 @@ and rule_id = {
   langs : Lang.t list;
   (* used for debugging (could be removed at some point) *)
   pattern_string : string;
+  options : rule_id_options option
 }
 [@@deriving show, eq]
+
+and rule_id_options  = {
+  max_match_per_file : int option;
+  (* maximum number of matches per file. *)
+}
+[@@deriving show, eq]
+
+let rule_id_options_of_rule_options (opts : Rule_options.t) =
+  { max_match_per_file = opts.max_match_per_file }
+
+let rule_id_options_of_rule_options_opt (opts : Rule_options.t option) =
+  Option.map rule_id_options_of_rule_options opts
+
+let rule_ids_to_map (opts: rule_id list) =
+  Rule_ID.Map.of_list (List_.map (fun r -> (r.id, r)) opts)
+
+let rule_ids_to_rule_id_options_map (opts: rule_id list) =
+  Rule_ID.Map.of_list
+    (List_.filter_map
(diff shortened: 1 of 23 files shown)


src/core/Core_match.ml
```diff
@@ -153,9 +153,35 @@ and rule_id = {
...
 [@@deriving show, eq]
 
+and rule_id_options  = {
+  max_match_per_file : int option;
+  (* maximum number of matches per file. *)
+}
+[@@deriving show, eq]
+
+let rule_id_options_of_rule_options (opts : Rule_options.t) =
+  { max_match_per_file = opts.max_match_per_file }
```
corneliuhoffman, line 167: Not blocking.
I am a bit confused by this, is it here for future more complicated options? It is used with
```ocaml
 Some Core_match.{max_match_per_file = Some limit; _}
 ```
 so in principle it can even stay as a Rule_options.t

File: src/reporting/Core_json_output.ml [hunk 2]
```diff
@@ -137,12 +137,59 @@ let core_unique_key (c : Out.core_match) : key =
...
+    |> Assoc.group_assoc_bykey_eff
+  in
+
+  let per_files_and_rules =
+    per_files
+    |> List_.map (fun (path, matches) ->
+          let matches_by_rule =
+            matches
+            |> List_.map (fun (Out.{check_id; _} as m : Out.core_match) -> (check_id, m))
+            |> Assoc.group_assoc_bykey_eff
```
corneliuhoffman: As discussed this is kind of terrible function altogether. maybe we should just use Assoc.group_by

corneliuhoffman (approved):
LGTM modulo minor things

src/core/Core_match.ml
dimitris-m: The type `rule_id` which contains `rule_id_options` must have `[@@deriving show, eq]` and `Rule_options.t` comes from atd and does not implement `equal` so it cannot be used directly.

And also I wanted to keep only what is relevant, in the spirit of `rule_id` vs `Rule.t` (`Rule.rule_info`)

Issue #339, opened 2025-07-09
Followup on #335


dimitris-m:
We closed #335 with #338 but we have still not determined the root cause. 

It is related to resolving imports and how this affects metavariables, but requires further investigation.

aayachnes:
I'm seeing non-determinism with this taint rule. Run `opengrep.sh` and then `cmp.sh` - `cmp.sh` will show that the results are different.

[example.zip](https://github.com/user-attachments/files/21168983/example.zip)

maciejpirog:
Possibly not a concurrency issue. Here's a run on my machine (note the `-j 1` in the script):
```bash
$ cat opengrep.sh                    
#!/bin/bash

set -e

run_opengrep() {
  local suffix="$1"
  opengrep scan -j 1 --disable-version-check --config rule.yaml --timeout-threshold 0 --output tmp.json --json --no-git-ignore files
  jq . tmp.json > "results/opengrep${suffix}.json"
  rm tmp.json
}

run_opengrep 1
run_opengrep 2
run_opengrep 3
$ sh opengrep.sh >/dev/null 2>/dev/null
$ sh cmp.sh                            
e2835e20856490f9a3aef65b3ae4b85d  results/opengrep1.json
3c79ff3b3b4f1cc525672d6c898620d8  results/opengrep2.json
3c79ff3b3b4f1cc525672d6c898620d8  results/opengrep3.json
```

dimitris-m:
Related: https://github.com/opengrep/opengrep/issues/362

maciejpirog:
Some more problems with fingerprints and nondeterministic results have been indeed found. Fixed in https://github.com/opengrep/opengrep/releases/tag/v1.8.1

PR #364, opened 2025-07-24
Improve deduplication and sorting of results

maciejpirog:
Improvements in the implementation of deduplication and sorting of matches and results. Makes the implementation simpler and deterministic.

## Summary of changes

- A custom comparison in `Core_match.uniq`
- `List.sort_uniq` instead of a hashtable-based deduplication
- A more deterministic comparison of matches in `Semgrep_output_utils.compare_match`.

## Rationale:

- Code simplification
- Fixes the nondeterminism bug (#339), hopefully
- Slight performance improvement

## Data:

Using a hashtable in `Core_match.uniq` is an overkill. Running opengrep-rules on a few mid-sized repos revealed that this function is run on lists of the following lengths:

| Length of input | # calls for pmd | # calls for jellyfin |
| ---: | ---: | ---: |
| 0 | 80800 | 43326 |
| 1 | 639 | 270 |
| 2 - 9| 1976 | 274 
| 10-99 | 2356 | 385 |
| 100-999 | 803 | 448 |
| 1000-9999 | 58 | 78 |
| 10000 - ... | 0 | 0 |

This means that for a real-life set of rules and real-life repos, only about 1% of applications to `uniq` are to lists of length greater or equal to 100, and, moreover, there are no cases in which the list is of length greater or equal to 10000. I such cases, the overhead of creating a hashtable in each of these cases and their internal working makes it less efficient then `List.sort_uniq`.

Also, looking at the arguments to `uniq`, most often they come more or less sorted already. In such cases the chunked mergesort that is used in the implementation of `List.sort.uniq`  is very efficient.

## Experiments

- In the great scheme of things, the perf improvement is neglegable,
- I can no longer reproduce the nondet bug,
- The tests pass,
- Compared the results using opengrep-rules on the 7 repos from the perf script, they all match

### DIFF
--- a/src/core/Core_match.ml
+++ b/src/core/Core_match.ml
@@ -215,22 +215,23 @@ end)
 
 (* Deduplicate matches *)
 let uniq (pms : t list) : t list =
-  (* perf: Previously used an initial size of 1024 but profiling with memtrace
-      revealed that this was oversized and contributed to 14% of all major heap
-      usage in an interfile run over juice-shop. Setting it to 8 meant it was
-      undersized in some cases and led to additional allocations as the table
-      gets resized. Change this only with some caution.
-      DOUBT(iago): What is the actual observed effect in terms of runtime and
-        peak memory usage?
-  *)
-  let size = List.length pms in
-  let matches_tbl = Tbl.create size in
-  (* dedup matches *)
-  pms
-  |> List.iter (fun match_ ->
-         if not (Tbl.mem matches_tbl match_) then Tbl.add matches_tbl match_ ());
-  (* list matches *)
-  Tbl.fold (fun m () acc -> m :: acc) matches_tbl []
+  let digest (t : t) =
+    (t.rule_id.id,
+     t.path,
+     (fst t.range_loc).pos.bytepos,
+     (snd t.range_loc).pos.bytepos,
+     t.env)
+  in
+  let cmp t1 t2 =
+    let c = compare (digest t1) (digest t2) in
+    if c <> 0 then c else
+      match t1.taint_trace, t2.taint_trace with
+      | None, None -> 0
+      | None, _ -> 1
+      | Some _, None -> -1
+      | Some t1, Some t2 -> compare (Lazy.force t1) (Lazy.force t2)
+  in
+  List.sort_uniq cmp pms
 [@@profiling]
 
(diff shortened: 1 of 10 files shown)


src/core/Core_match.ml
```diff
@@ -215,22 +215,23 @@ end)
...
+     (snd t.range_loc).pos.bytepos,
+     t.env)
+  in
+  let cmp t1 t2 =
+    let c = compare (digest t1) (digest t2) in
+    if c <> 0 then c else
+      match t1.taint_trace, t2.taint_trace with
+      | None, None -> 0
+      | None, _ -> 1
+      | Some _, None -> -1
```
dimitris-m: `_, None` would also work, since if the lhs was `None` it would have been caught by the case above.

src/core/Core_match.ml
```diff
@@ -215,22 +215,23 @@ end)
...
+      | Some t1, Some t2 -> compare (Lazy.force t1) (Lazy.force t2)
```
dimitris-m: I think the same things we discussed offline for comparing digests (using `compare`) also apply here. 

`Taint_trace.t` derives `ord` so it has compare.

src/core/Semgrep_output_utils.ml
```diff
@@ -168,7 +168,11 @@ let compare_match (a : core_match) (b : core_match) =
   let (bloc : location) = { path = b.path; start = b.start; end_ = b.end_ } in
 
   let c = compare_location aloc bloc in
-  if c <> 0 then c else compare_match_extra a.extra b.extra
+  if c <> 0 then c else
+  let e = compare_match_extra a.extra b.extra in
+  if e <> 0 then e else
+  (* make a deep comparison to make sure the order is always deterministic *)
+  compare a b
```
dimitris-m: Not sure about this, why is it needed? 

Here we need something custom, because `Semgrep_output_v1_t.core_match` does not implement / derive ord....

dimitris-m:
I also can no longer reproduce the fingerprint issue on this branch.

src/core/Core_match.ml
maciejpirog: Yep, I added deriving `ord` to `Taint_trace.t`, but forgot to use it :). Fixing now ...

PR #367, opened 2025-07-25
Fix bug in propagating taints through composites and dictionaries

corneliuhoffman:
This PR deals with the various differences between semgrep 1.52 and opengrep.
It turns out the issue is quite funny. 
In opengrep 1.8.1 a new construct `is_exact` was introduced. It was supposed to avoid finding false positive given by sub-expressions. Before that, the sub-expressions finds hid an actual bug that did not propagate taints in dictionaries an composites. Once the sub-expression were removed new false negatives sprung up in ruby. 
This fixes all of them (I think). It also fixes similar issues in javascript (see test) and potentially other languages.

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -131,6 +131,7 @@ let propagate_through_indexes env =
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
+let add_taints_from_shape shape = Taints.union (Shape.gather_all_taints_in_shape shape)
 
 let log_timeout_warning (taint_inst : Taint_rule_inst.t) opt_name timeout =
   match timeout with
@@ -156,6 +157,7 @@ let map_check_expr env check_expr xs =
   in
   (List.rev rev_taints_and_shapes, lval_env)
 
+
 let union_map_taints_and_vars env check xs =
   let taints, lval_env =
     xs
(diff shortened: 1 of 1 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -1172,7 +1173,15 @@ and check_tainted_expr env exp : Taints.t * S.shape * Lval_env.t =
     | Composite ((CTuple | CArray | CList), (_, es, _)) ->
         let taints_and_shapes, lval_env = map_check_expr env check es in
         let tuple_shape = Shape.tuple_like_obj taints_and_shapes in
-        (Taints.empty, tuple_shape, lval_env)
+        let all_taints =
+          taints_and_shapes
+          |> List.fold_left
+               (fun acc (taints, shape) ->
+                 acc |> Taints.union taints
+                 |> Taints.union (Shape.gather_all_taints_in_shape shape))
```
dimitris-m: This seems duplicated in a couple more places, maybe it could be a function and reused? With inlining if you think it's appropriate.

corneliuhoffman: done, just the later because it appears a lot

tests/rules/dict_js_test.js
```diff
@@ -0,0 +1,5 @@
+function redirectToBoard(bsn) {
+//ruleid: test
+    redirect_to({ id: bsn }.values);
```
dimitris-m, line 3: indentation of comment

tests/rules/dict_ruby_test.rb
```diff
@@ -0,0 +1,4 @@
+def redirect_to_board(bsn)
+  # ruleid: test
+sink_to {id: bsn}.values
```
dimitris-m, line 3: indentation

PR #370, opened 2025-07-28
Make `--experimental` bypass `argv[0]`

dimitris-m:
## Changes 

- Ensure that when `--experimental` is passed, `argv[0]` does not need to be in the specified values `opengrep` and `opengrep-cli` in order for the new OCaml CLI to be invoked.
- Allow `--experimental` to be passed when `argv[0] = "opengrep"`.

### DIFF
--- a/src/main/Main.ml
+++ b/src/main/Main.ml
@@ -144,8 +144,6 @@ let () =
         Array.mem "--experimental" argv
       in
       match argv0, experimental with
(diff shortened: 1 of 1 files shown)


dimitris-m:
> Why didn't we do it earlier?

Honestly we should add flags and not depend on `argv[0]` at all... to be continued.

PR #373, opened 2025-07-29
Use trie instead of hashtable in entropy

maciejpirog:
No need to use a hashtable when an array suffices

### DIFF
--- a/src/engine/Entropy.ml
+++ b/src/engine/Entropy.ml
@@ -1,7 +1,36 @@
 (*
-   A special built-in string analyzer that estimates whether a string
-   looks like a secret.
+   Ananalyzer that estimates whether a string looks like a secret.
 *)
+
+(* A specialized trie in which keys are A-Z strings of length 3 *)
+module Trie = struct
+  type t = float option array array array
+
+  let create () : t =
+    let empty = Array.make_matrix 26 26 None in
+    Array.init 26 (fun _ -> Array.map Array.copy empty)
+
+  let index_of_char c = Char.code c - Char.code 'A'
+
+  let insert (trie : t) (k : string) (v : float) : unit =
+      let a = index_of_char k.[0] in
+      let b = index_of_char k.[1] in
+      let c = index_of_char k.[2] in
+      trie.(a).(b).(c) <- Some v
+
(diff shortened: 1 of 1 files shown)


src/engine/Entropy.ml
```diff
@@ -1,8 +1,37 @@
 (*
-   A special built-in string analyzer that estimates whether a string
-   looks like a secret.
+   Ananalyzer that estimates whether a string looks like a secret.
 *)
 
+(* A sepcialized trie in which keys are A-Z strings of length 3 *)
```
corneliuhoffman: specialised (so proud of myself for finding this)

src/engine/Entropy.ml
```diff
@@ -1,8 +1,37 @@
...
+      let a = index_of_char k.[0] in
+      let b = index_of_char k.[1] in
+      let c = index_of_char k.[2] in
+      t.(a).(b).(c) <- Some v
+
+  let is_valid_key s =
+    String.length s = 3 &&
+    String.for_all (fun c -> c >= 'A' && c <= 'Z') s
+
+  let lookup (trie : float option array array array) (s : string) : float option =
```
corneliuhoffman: is trie not of type t?

src/engine/Entropy.ml
maciejpirog: fixed

src/engine/Entropy.ml
maciejpirog: fixed

maciejpirog:
> this looks ok but i still don't understand the point of saving a file with the counts of trinomes, read it and the process it every time. Can we not just do that once, serialize and save in that file? there are also some obnoxious refs as well :)))

Of course, but since secret detection does not seem to contribute much to the run time, not a priority.

> Also since this is so easy can we not save the array as a flat 26_26_26 array and process the trinome position as 26_26_a + 26*b +c. I think it would be mildly more efficient

I tend to use this kind of arithmetic to the compiler, I would definitely get it wrong.

corneliuhoffman (approved):
lgtm

dimitris-m (approved):
lgtm

PR #375, opened 2025-07-30
Fix: inline metavars in incremental json output

dimitris-m:
When the option is set, inline metavariables in json output (but not in text output since this would go to waste).

### DIFF
--- a/src/osemgrep/cli_scan/Scan_subcommand.ml
+++ b/src/osemgrep/cli_scan/Scan_subcommand.ml
@@ -137,9 +137,9 @@ let output_and_exit_from_fatal_core_errors_exn ~exit_code
  * the use of a mutex below. *)
 let file_match_hook_mutex = Mutex.create ()
 
-let mk_file_match_hook (conf : Scan_CLI.conf) (rules : Rule.rules)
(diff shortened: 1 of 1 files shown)


src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -219,14 +220,17 @@ let choose_output_format_and_match_hook (caps : < Cap.stdout >)
       _;
     } ->
       ( Output_format.Incremental,
-        Some (mk_file_match_hook conf rules (incremental_text_printer caps)) )
+        Some (mk_file_match_hook ~inline_metavars:false (* because text format *)
```
corneliuhoffman, line 223: do you need that? it is false to start with

corneliuhoffman (approved):

src/osemgrep/cli_scan/Scan_subcommand.ml
dimitris-m: I added it for clarity

dimitris-m: in that sense the parameter need not be optional of course.... maybe I'll remove

dimitris-m: which I did now ^

PR #376, opened 2025-07-30
Precompile regexp stored in the CondRegexp constructor

maciejpirog:
Before: regex used in `metavariable-regex` was stored as a string and compiled to pcre's regex every time it was matched.

After: we precompile the regex. We keep both version in the `CondRegexp` constructor, so that in places where we need the string representation, it is still available.


src/metachecking/Check_rule.ml
```diff
@@ -132,7 +132,7 @@ let unknown_metavar_in_comparison r f =
              | CondName _ ->
                  (Set.empty, [])
              | CondRegexp (_, regex, _) ->
-                 (Mvar.mvars_of_regexp_string regex |> Set_.of_list, [])
+                 (Mvar.mvars_of_regexp_string regex.pattern |> Set_.of_list, [])
```
dimitris-m, line 135: But `mvars_of_regexp_string` is defined like this: 

```ocaml
let mvars_of_regexp_string s =
  Pcre2_.pcre_compile s |> Pcre2_.pcre_regexp |> Pcre2.names |> Array.to_list
  |> Common.(List_.map (fun s -> spf "$%s" s))
```

So we could avoid the compilation here perhaps, passing the compiled regex?

src/optimizing/Analyze_rule.ml
```diff
@@ -354,7 +354,7 @@ and metavarcond_step1 ~is_id_mvar x =
   | R.CondEval _ -> None
   | R.CondNestedFormula _ -> None
   | R.CondRegexp (mvar, re, const_prop) ->
-      if is_id_mvar mvar then Some (MvarRegexp (mvar, re, const_prop)) else None
+      if is_id_mvar mvar then Some (MvarRegexp (mvar, re.pattern, const_prop)) else None
```
dimitris-m, line 357: I think `MvarRegexp` should also take the full pcre object, for example we have code like this were compilation happens: 

```ocaml
let or_step2 (Or xs) =
  let step1_to_step2 =
    List_.map (function
      | StringsAndMvars ([], _) -> raise GeneralPattern
      | StringsAndMvars (xs, _) -> Idents xs
      | Regexp re_str -> Regexp2_search (Pcre2_.pcre_compile re_str)
      | MvarRegexp (_mvar, re_str, _const_prop) ->
          (* The original regexp is meant to apply on a substring.
              We rewrite them to remove end-of-string anchors if possible. *)
          let re =
            match
              Pcre2_.remove_end_of_string_assertions
                (Pcre2_.pcre_compile re_str)
            with
            | None -> raise GeneralPattern
            | Some re -> re
          in
          Regexp2_search re)
```

src/metachecking/Check_rule.ml
maciejpirog: The regex used for matching is compiled with `ANCHOR` flag, this one without any flags. I'm pretty sure it does not change the names in the expression, but wanted to play it safe.

src/optimizing/Analyze_rule.ml
maciejpirog: same as above: no `ANCHOR` flag + `remove_end_of_string_assertions`, which we cannot do after we compile

dimitris-m (commented):
Perhaps we could also adapt: 

`eval_regexp_matches`

in `src/engine/Eval_generic.ml`?

dimitris-m (approved):

src/metachecking/Check_rule.ml
dimitris-m: Ok. From a quick look this does not seem to be in any hot path, we can ignore.

src/optimizing/Analyze_rule.ml
dimitris-m: Ok -- and same comment as above, we can ignore I think.

dimitris-m:
> > Perhaps we could also adapt:
> > `eval_regexp_matches`
> > in `src/engine/Eval_generic.ml`?
> 
> I don't understand. It is adapted in this PR to accept compiled regexps

Sorry, my bad.

PR #379, opened 2025-08-01
When comparing locations, compare only selected fields

maciejpirog:
perf revealed 1.16% time is spent in `caml_string_compare` called from `Tok.equal_location`. Hence:

To compare tokens, we compare:

- `bytepos`
- `file`

We do not compare:

- content of a token,
- line and col

as they are determined by `bytepos` and `file`

### DIFF
--- a/libs/lib_parsing/Pos.ml
+++ b/libs/lib_parsing/Pos.ml
@@ -60,8 +60,8 @@ type t = {
    *)
   bytepos : int; (* 0-based *)
   (* Those two fields can be derived from bytepos (See complete_position() *)
-  line : int; (* 1-based *)
-  column : int; (* 0-based *)
+  line : int [@eq.ignore][@ord.ignore]; (* 1-based *)
+  column : int [@eg.ignore][@ord.ignore]; (* 0-based *)
   (* TODO: use an Src.t/Origin.t instead? (see spacegrep Src_file.source *)
   file : Fpath_.t;
 }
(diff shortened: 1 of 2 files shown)


libs/lib_parsing/Pos.ml
```diff
@@ -60,10 +60,10 @@ type t = {
    *)
   bytepos : int; (* 0-based *)
   (* Those two fields can be derived from bytepos (See complete_position() *)
-  line : int; (* 1-based *)
-  column : int; (* 0-based *)
+  line : int [@eq.ignore]; (* 1-based *)
+  column : int [@eg.ignore]; (* 0-based *)
   (* TODO: use an Src.t/Origin.t instead? (see spacegrep Src_file.source *)
-  file : Fpath_.t;
+  file : Fpath_.t [@eq.ignore];
```
dimitris-m: Could this complicate things when we move to interfile?

dimitris-m: and, why not ignore for `ord` too, these same fields?

dimitris-m: I suggest add a clear warning (comment), why this is defined with these fields ignored.

libs/lib_parsing/Tok.ml
```diff
@@ -55,7 +55,7 @@ open Sexplib.Std
  *)
 type location = {
   (* the content of the "token" *)
-  str : string;
+  str : string [@eq.ignore];
```
dimitris-m: Same as before, if we can do that, then `ord` could also have this field ignored?

libs/lib_parsing/Pos.ml
maciejpirog: Changed this, we now compare on `file` as well

libs/lib_parsing/Tok.ml
maciejpirog: Yes. Fixed

PR #383, opened 2025-08-11
Make entropy data non-lazy

maciejpirog:
Before: in Entropy.ml, the lazy variable `data_tables` leads to a bug, because it can be forced by different threads simultaneously, which is not allowed in OCaml.

After: `data_tables` is no longer lazy.


dimitris-m (approved):
LGTM

PR #385, opened 2025-08-12
fix failure in to-ast-generic for C and C++

maciejpirog:
For C and C++ files that were parsed (or partially parsted), we sometimes encountered failure during the to-ast-generic step for incorrect programs or programs that use the preprocessor. After this fix we never fail: if the file is (at least partially) parsed, we always produce an AST.


dimitris-m (approved):
lgtm (discussed offline)

PR #390, opened 2025-08-15
Fix: install script `sed` call should be posix-compliant

dimitris-m:
The previous version only worked on GNU sed, so ok for homebrew sed but not Mac OSX default sed.


corneliuhoffman (approved):
lgtm

PR #395, opened 2025-08-20
Fix AIK_react-dangerouslysetinnerhtml bug

corneliuhoffman:
This PR fixes a bug in Naming_AST for TS. The issue is that the visitor for PatTyped missed the case of Record patterns and, as a consequence, the AIK_react-dangerouslysetinnerhtml failed to find a tainting in the following simple code:
```typescript
export const ErrorDetails = ({ error}: { error: TestError}) => { __html: formatDiffMessage(error)}
```


dimitris-m (approved):
lgtm

Issue #399, opened 2025-08-25
Fatal error: index out of bounds when using --baseline flag


tom-paz:
**Describe the bug**
  Opengrep crashes with a fatal "index out of bounds" error when using
  the `--baseline` flag. The crash occurs during nosemgrep (ignore
  pattern) processing when the system attempts to read line numbers that
   exceed the actual number of lines in a file. This results in a
  complete failure of the baseline scan with an array bounds exception.

  **To Reproduce**
  1. Run opengrep with the `--baseline` flag on a repository:
     ```bash
     opengrep scan   <target_directory>  --config <rules> --baseline <commit_hash>

  2. The scan fails when processing files where match positions
  reference line numbers beyond the file's actual line count
  3. Error message follows this pattern:
  [ERROR]: Uncaught exn in Processor.post_process: 
  Common.ErrorOnFile("lines_of_file(): index out of bounds on index 
  <N>", "<file_path>")
  Fatal error: exception Common.ErrorOnFile("lines_of_file(): index out
  of bounds on index <N>", "<file_path>")

  Expected behavior
  The baseline scan should complete successfully without crashing by:
  - Properly validating line ranges before attempting file access
  - Gracefully handling cases where match positions reference
  non-existent lines
  - Providing meaningful error messages instead of fatal crashes
  - Continuing the scan even when individual files have line count
  discrepancies

  What is the priority of the bug to you?
  - P0: blocking your adoption of Opengrep or workflow

  Environment
  - Local build from git repository
  - Platform: macOS (Darwin 24.6.0)
  - Any repository where file line counts may vary between baseline and
  head commits

  Use case
  Fixing this bug will enable:
  1. Reliable use of baseline/differential scanning across diverse
  codebases
  2. Gradual migration to new rules without being blocked by existing
  findings
  3. CI/CD pipelines that depend on --baseline functionality
  4. Better development workflows where only new findings are surfaced
  5. Scanning of repositories with files that may have inconsistent line
   counts between commits

  Technical Details
  The error originates in UFile.lines_of_file_exn when array bounds
  checking fails. The call stack shows the issue propagates through
  nosemgrep processing:
  - UFile.lines_of_file_exn - Array access fails on out-of-bounds index
  - Nosemgrep.rule_match_nosem - Calls file reading with potentially
  invalid line ranges
  - Nosemgrep.produce_ignored - Processes matches for ignore patterns
  - Pre_post_core_scan.Nosemgrep_processor.post_process -
[...]

PR #400, opened 2025-08-25
Fixing baseline compare nosemgrep line count bug

tom-paz:
Opengrep crashes with a fatal "index out of bounds" error when using
  the --baseline flag. The crash occurs during nosemgrep (ignore
  pattern) processing when the system attempts to read line numbers that
   exceed the actual number of lines in a file, particularly when files
  differ between baseline and head commits.

  Root Cause

  The error originates in UFile.lines_of_file_exn where arr.(i) fails
  when trying to access line indices beyond the file's actual line
  count. This happens when match positions reference line numbers that
  don't exist in the current file state.

  Fix

  - Added bounds checking in UFile.lines_of_file_exn to validate array
  indices before access
  - Used filter_map to silently skip out-of-bounds line indices instead
  of crashing
  - Added defensive limits in Nosemgrep.rule_match_nosem to cap line
  ranges (max 1000 lines from start)

  This ensures baseline/differential scanning works reliably even when
  match positions reference non-existent lines, enabling gradual rule
  migration without fatal crashes.

  Fixes: #399

### DIFF
--- a/libs/commons/UFile.ml
+++ b/libs/commons/UFile.ml
@@ -313,11 +313,12 @@ let lines_of_file_exn (start_line, end_line) file : string list =
   (* This is the case of the empty file. *)
   | [| "" |] -> []
   | _ ->
-      let max_index = Array.length arr - 1 in
       lines
-      |> List_.filter_map (fun i ->
-             if i >= 0 && i <= max_index then
-               Some arr.(i)
-             else
-               (* Skip out-of-bounds indices silently to prevent crashes *)
-               None)
+      |> List_.map (fun i ->
+             try arr.(i) with
+             | Invalid_argument s ->
+                 let exn =
+                   Common.ErrorOnFile
+                     (spf "lines_of_file(): %s on index %d" s i, file)
+                 in
+                 Exception.catch_and_reraise exn)
--- a/libs/commons/UFile.ml
+++ b/libs/commons/UFile.ml
@@ -308,11 +308,14 @@ let find_first_match_with_whole_line path ?split term =
  *)
 let lines_of_file_exn (start_line, end_line) file : string list =
   let arr = cat_array file in
-  let lines = List_.enum start_line end_line in
   match arr with
   (* This is the case of the empty file. *)
   | [| "" |] -> []
   | _ ->
+      let max_index = Array.length arr - 1 in
+      let safe_start = max 0 (min start_line max_index) in
+      let safe_end = max safe_start (min end_line max_index) in
+      let lines = List_.enum safe_start safe_end in
       lines
       |> List_.map (fun i ->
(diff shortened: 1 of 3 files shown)


tom-paz:
Yea your right, removed that naive file number 👍

libs/commons/UFile.ml
```diff
@@ -313,12 +313,17 @@ let lines_of_file_exn (start_line, end_line) file : string list =
...
-                 Exception.catch_and_reraise exn)
+      |> List_.filter_map (fun i ->
+             if i >= 0 && i <= max_index then
+               try Some arr.(i) with
+               | Invalid_argument s ->
+                   let exn =
+                     Common.ErrorOnFile
+                       (spf "lines_of_file(): %s on index %d" s i, file)
+                   in
+                   Exception.catch_and_reraise exn
```
dimitris-m: If you already checked the bounds in the `if`, then the exception can never happen.

tom-paz: fixed

libs/commons/UFile.ml
```diff
@@ -308,11 +308,14 @@ let find_first_match_with_whole_line path ?split term =
  *)
 let lines_of_file_exn (start_line, end_line) file : string list =
   let arr = cat_array file in
-  let lines = List_.enum start_line end_line in
   match arr with
   (* This is the case of the empty file. *)
   | [| "" |] -> []
   | _ ->
+      let max_index = Array.length arr - 1 in
+      let safe_start = max 0 (min start_line max_index) in
```
dimitris-m, line 316: can `start_line` or `max_index` be < 0 ? 

I'd guess that no ?

dimitris-m, line 316: I would delete this line unless if there's a specific reason I'm missing.

libs/commons/UFile.ml
```diff
@@ -308,11 +308,14 @@ let find_first_match_with_whole_line path ?split term =
...
+      let safe_end = max safe_start (min end_line max_index) in
+      let lines = List_.enum safe_start safe_end in
```
dimitris-m, line 318:
```suggestion
      let safe_end_line = min end_line max_index in
      let lines = List_.enum start_line safe_end_line in
```

libs/commons/UFile.ml
dimitris-m: maybe keep just `min start_line max_index`?

libs/commons/UFile.ml
dimitris-m: Note also that the function is called ..._exn which indicates it may raise, which here you are trying to avoid.

I suspect this issue is only happening with diff scan.

PR #403, opened 2025-08-27
Scala: Support metavars as elements in interpolated strings

### LINKED ISSUE #106: Support for metavariables in interpolated strings for Scala
When looking for patterns in interpolated strings like 
```yaml
- pattern: s"..."
```
it is possible to determine if your tainted user input goes into the interpolated string.
But what is not possible is something like this:
```yaml
- focus-metavariable: $TAINTED
- pattern: s"$BAD...$TAINTED"
- metavariable-regex:
      metavariable: $BAD
      regex: This text cannot be followed by a tainted value
```

It should be possible to add metavariables into interpolated strings like s"..." and f"..." for Scala.
I think a second issue arises if you also want to add taint analysis. Variables are also identified with a $ eg:
```scala
val name = "James"
val age = 30
println(s"$name is $age years old") 
```

**Describe alternatives you've considered**

**Use case**
What will this feature enable for you?

**Additional context**

maciejpirog:
Closes #106 

In this PR:
- The lexer does not ignore `$` inside an interpolated string when it is followed by a metavar identifier
- In the `sgrep` mode (i.e., when parsing patterns) the parser distinguishes between variables and metavariables  by checking if the identifier starts with `$`.  In non-`sgrep` mode (i.e., when parsing targets), the parser treats all identifiers as variables.


languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
```diff
@@ -2303,9 +2303,17 @@ and interpolatedString ~inPattern in_ =
...
             (* ast: Ident(x) *)
-            xs += EncapsDollarIdent x
+            begin
+                (* if the identifier starts with $, it is a metavariable, not a variable
+                 * inside an interpolated string. However, if we are not in semgrep_mode,
+                 * we treat these identifiers as normal variables, so we strip the $ *)
+                match s.[0], Domain.DLS.get Flag_parsing.sgrep_mode with
+                | '$', true -> xs += EncapsExpr (Name (Id x, []))
+                | '$', false -> xs += EncapsDollarIdent (String.sub s 1 (String.length s - 1), t)
+                | _ -> xs += EncapsDollarIdent x
```
dimitris-m: It would probably be more efficient to check `Char.equal s.[0] '$'` and only in that case read from DLS.

tests/patterns/scala/metavars_in_interpolated_strings_2.scala
```diff
@@ -0,0 +1,4 @@
+
+//ERROR: match
+println(s"this $x value comes $x")
```
dimitris-m: weird text, maybe:

`println(s"this $x value is $x")`

dimitris-m (approved):
lgtm, left some minor comments.

languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
maciejpirog: fixed

tests/patterns/scala/metavars_in_interpolated_strings_2.scala
maciejpirog: fixed

PR #408, opened 2025-08-29
Fix comparison of tokens and locations

maciejpirog:
- Replace faulty ppx for `ord` and `eq` for positions and locations with direct definition
- Add tests

### DIFF
--- a/libs/lib_parsing/Pos.ml
+++ b/libs/lib_parsing/Pos.ml
@@ -60,12 +60,22 @@ type t = {
    *)
   bytepos : int; (* 0-based *)
   (* Those two fields can be derived from bytepos (See complete_position() *)
-  line : int [@eq.ignore][@ord.ignore]; (* 1-based *)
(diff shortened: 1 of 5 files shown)


libs/lib_parsing/Pos.ml
```diff
@@ -60,12 +60,22 @@ type t = {
...
+  column : int; (* 0-based *)
   (* TODO: use an Src.t/Origin.t instead? (see spacegrep Src_file.source *)
   file : Fpath_.t;
 }
-[@@deriving show, eq, ord, sexp]
+[@@deriving show, sexp]
+
+let compare (pos1 : t) (pos2 : t) : int =
+  let cmp_bpos = Int.compare pos1.bytepos pos2.bytepos in
+  if cmp_bpos =|= 0
```
dimitris-m: I find `Int.equal` much clearer...

maciejpirog: I agree, fixed

PR #409, opened 2025-09-01
Direct implementation of is_metavar_name and `is_metavar_ellipsis`, not via memoized regex

maciejpirog:
A direct implementation of `is_metavar_name`  and `is_metavar_ellipsis`, which are used **a lot** during matching. The direct implementation should be faster and more GC-friendly.

Quick unreliable benchmarks:

name | main | fix | diff(s) | diff(%) | same-results
-- | -- | -- | -- | -- | --
jellyfin | 81.70 | 79.38 | -2.32 | -2.84% | True
lemur | 35.96 | 33.85 | -2.11 | -5.87% | True
ComfyUI-Custom-Scripts | 14.34 | 14.08 | -0.26 | -1.82% | True
pmd | 105.70 | 90.56 | -15.14 | -14.32% | True
leakcanary | 9.83 | 9.76 | -0.07 | -0.72% | True

### DIFF
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -2491,12 +2491,42 @@ let special_multivardef_pattern = "!MultiVarDef!"
 (* Semgrep hacks *)
 (*****************************************************************************)
 
+let is_metavar_name_body len start s =
+    let check_first c =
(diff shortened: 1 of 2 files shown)


dimitris-m (approved):
lgtm, modulo duplication of some functions as discussed offline

PR #439, opened 2025-10-17
Add --taint-intrafile to python

dimitris-m:


fopina:
Hey @dimitris-m , one of the things I’ve been eager to see in opengrep is inter file taint

just noticed this MR and I see it’s “intra”, not inter but no description of any kind to understand what the flag enables
Would you be able to clarify?

PR #442, opened 2025-10-19
vb.net: lexer

maciejpirog:
# Support for VB.NET part 1 (lexer)

To make the process of reviewing easier, the support for VB.NET will be added in steps. This is step 1: **lexer**. This PR merges to the `mpir/vbnet` branch. Only when all parts are reviewed, I will merge it with `main`.

Notes:

1. The entry point is `Vbnet_tokenize.tokenize`

2. The type of the token produced by the lexer is:

```ocaml
type token_kind =
  | Identifier
  | Keyword
  | Operator
  | ...

type t = {
  kind: token_kind;
  content: string;
  tok : Tok.t
}
```

Of course only `Tok.t` will ultimately land in the generic AST. The other fields will be used by the parser.

3. There are some helper functions for the parser (such as `ghost_token`), which are not used at the moment.

4. We use ocamllex, which works only on bytes, while real-life VB.NET uses a lot of Unicode non-ascii characters. Hence, there is a crude support for UTF-8 by a byte-by-byte analysis. It is not perfect yet, as some correct characters are not recognised, I will keep fixing this continuously.

5. We need a stateful lexer to support interpolated strings. Since ocamllex does not support state by itself, the usual solution is to add a `ref` that keeps the state. ~~We need to be careful because we use multiucore, but instead of the ugly imperative `DLS`, I propose to use a local `ref` + **effect handlers**. I think this solution is more functional (as in "functional programming"), and in general a more modern approach to handling "local global" state.~~

### DIFF
--- a/languages/vbnet/Unit_vbnet_tokenize.ml
+++ b/languages/vbnet/Unit_vbnet_tokenize.ml
@@ -37,5 +37,37 @@ let tests =
          ; Punctuation, "}"
          ; EOF, ""])
       in
-      assert (s = r))
+      assert (s = r));
+
+    Testo.create "lexer (nested interpolated strings)" (fun () ->
+      let s = tokenize "$\"{\"xyz\" a $\"{b \"xyz\"}\"}\"" |> List.map preview in
+      let r =
+      T.([ Operator, "$\""
+         ; StringLiteral, "\"xyz\""
+         ; Identifier, "A"
+         ; Operator, "$\""
+         ; Identifier, "B"
+         ; StringLiteral, "\"xyz\""
+         ; Operator, "\""
+         ; Operator, "\""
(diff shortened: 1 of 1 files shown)


maciejpirog:
> First question, why not use sedlex which supports unicode?
> 
> https://github.com/ocaml-community/sedlex

Because we already have good integration with ocamllex in terms of converting ocamllex's `lexbuf` tokens to our `Tok` tokens, and I want to leverage on that

...is what I thought a month ago when I started to work on this. Only by the end of implementation I learnt that it's not true, and we need to do steps that are now implemented in `complete_location_in_token` and `fix_pos`.

Another reason is that LLMs are very good with ocamllex, not so much with sedlex

PR #443, opened 2025-10-20
vb.net: generated raw parser

maciejpirog:
# Support for VB.NET part 2 (raw parser)

To make the process of reviewing easier, the support for VB.NET will be added in steps. This is step 2: **raw parser**. This PR merges to the `mpir/vbnet` branch. Only when all parts are reviewed, I will merge it with `main`.

Notes:

1. Example use:

```ocaml
Vbnet_tokenize.tokenize "2 + 2" |> Vbnet_parser.run Vbnet_parser.expression
```

2. The actual contribution is the content of the file `Vbnet_parser.ml`. The rest is the script used to generate it. The intention is that the parser will **NOT** be regenerated from the grammar ever again, but we will directly edit the `Vbnet_parser.ml` file.

3. The parser is a bundle of mutually-recursive backtracking combinator parsers, all of which return a value of `AST_generic.any`. In the next parts, the file `Vbnet_parser.ml` will be edited to return specific kinds of AST_generic.

An alternative would be to extend the scripts so that they include semantic actions, but that would require a substantial amount of unnecessary work.

4. The lexer or parser does not contain any opengrep specific constructs (ellipsis, metavars). It is easy to add them later, the syntax of the language does not conflict with them in any way.

5. The "ghost tokens" are tokens that you cannot match, but you can peek (so when you match a token, ghost tokens are skipped). The only ghost tokens are newlines, because they are semantically relevant it a couple of places (e.g. to distinguish between single- and multi-line versions of subs, blocks, and lambdas), but can be skipped everywhere else.

### DIFF
--- a/scripts/vbnet/grammar.py
+++ b/scripts/vbnet/grammar.py
@@ -1,0 +1,193 @@
+import common
+from lark import Lark, Transformer, v_args, exceptions
+import sys
+import re
+from pprint import pprint
+
+# parser of antlr grammars
+
+grammar = r"""
+    start: rules
+
+    dir_token_type: "%token_type" LINE
+
+    dir_token_match: "%token_match" LINE
+
+    dir_token_ghost: "%token_ghost" LINE
+
+    dir_inline_code: "% " LINE
+
+    rules: rule*
+
+    rule: NAME ":" productions ";"
+
+    productions: production ("|" production)*
+
+    production: modified_item*
+
+    modified_item: base_item modifier?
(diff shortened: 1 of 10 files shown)


scripts/vbnet/grammar.py
```diff
@@ -0,0 +1,193 @@
+import common
+from lark import Lark, Transformer, v_args, exceptions
```
dimitris-m, line 2: this is the only external dependency right?

maciejpirog: yes

PR #444, opened 2025-10-21
Rename the functions and simpler eval

corneliuhoffman:
- added Go specific method names
- changed the eval so you do not lose methods/functions with the same name.

TODO: improve the graph construction to incorporate methods with the same name (like this [file ](https://github.com/sampion88/dotnet-samples/blob/main/framework/wcf/Extensibility/MessageEncoder/Text/CS/library/CustomTextMessageEncoder.cs#L65)in c#)

### DIFF
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -293,351 +293,320 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
     Common.with_time (fun () -> lazy_force lazy_ast_and_errors)
   in
   (* TODO: 'debug_taint' should just be part of 'res'
-   * (i.e., add a "debugging" field to 'Report.match_result'). *)
-  match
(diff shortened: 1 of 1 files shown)


src/tainting/Function_call_graph.ml
```diff
@@ -51,23 +51,60 @@ module Dot = Graph.Graphviz.Dot (struct
...
+  (* Non-pointer receiver: func (r Type) ... *)
+  | G.ParamReceiver { ptype = Some { t = G.TyN (G.Id ((name, _), _)); _ }; _ }
+    :: _ ->
+      Some name
+  (* Pointer receiver: func (r *Type) ... *)
+  | G.ParamReceiver { ptype = Some { t = G.TyPointer (_, { t = G.TyN (G.Id ((name, _), _)); _ }); _ }; _ }
+    :: _ ->
+      Some name
+  | _ -> None
+
```
dimitris-m, line 82: duplication

PR #446, opened 2025-10-22
vbnet: add opengrep primitives to raw parser

maciejpirog:
# Support for VB.NET part 3 (add opengrep primitives to raw parser)

To make the process of reviewing easier, the support for VB.NET will be added in steps. This is step 3: **add opengrep primitives to raw parser**. This PR merges to the `mpir/vbnet` branch. Only when all parts are reviewed, I will merge it with `main`.

Notes:

1. We add metavariable, ellipsis, and deep ellipsis.

2. We don't add names ellipsis and typed metavariables. Should we?

### DIFF
--- a/src/tests/Test.ml
+++ b/src/tests/Test.ml
@@ -120,6 +120,7 @@ let tests (caps : Cap.all_caps) =
       Unit_include_filter.tests;
       Unit_parsing.tests ();
       Unit_vbnet_tokenize.tests;
+      Unit_vbnet_parser.tests;
       Unit_entropy.tests;
       Parser_regexp.Unit_parsing.tests;
       Unit_ReDoS.tests;
--- a/languages/vbnet/Unit_vbnet_parser.ml
+++ b/languages/vbnet/Unit_vbnet_parser.ml
@@ -1,0 +1,35 @@
+module T = Vbnet_tokenize
+module P = Vbnet_parser
+
+let (>>>) (s : string) (p : 'a P.parser) : unit =
+  let r = T.tokenize s |> P.run p in
+  assert (List.length r = 1);
+  assert ((List.hd r).P.next |> List.length = 1 (* EOF *) )
+
+let (>/>) (s : string) (p : 'a P.parser) : unit =
+  let r = T.tokenize s |> P.run p in
+  assert (List.length r = 0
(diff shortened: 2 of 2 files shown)


scripts/vbnet/grammar.txt
```diff
@@ -542,14 +542,22 @@ parameter_list
...
 
 implements_clause
-  : ':'? 'Implements' type (',' type)*
+  : ':'? 'Implements' type_list_elem (',' type_list_elem)*
+  ;
+
+type_list_elem
+  : type
+  | '...'
+  | '<...' type '...>'
```
dimitris-m, line 560: Just curious, why is this needed?

src/tests/Test.ml
```diff
@@ -120,6 +120,7 @@ let tests (caps : Cap.all_caps) =
       Unit_include_filter.tests;
       Unit_parsing.tests ();
       Unit_vbnet_tokenize.tests;
+      Unit_vbnet_parser.tests;
```
dimitris-m, line 123: Not important: maybe these vbnet tests should be nested under `Unit_parsing.tests`?

scripts/vbnet/grammar.txt
dimitris-m, line 1: So the parser was regenerated based on these changes, and then edited?

dimitris-m (approved):
Left some minor comments

scripts/vbnet/grammar.txt
maciejpirog: This is element of a list of types that instantiate another type. The syntax is as follows, say you want to instantiate a key-value map:

`SomeCollection (Of String, Integer)` 

Now, you want to be able to match these types as well, e.g.

`SomeCollection (Of $TYPE, Integer)`

or

`SomeCollection (Of ..., Integer)`

scripts/vbnet/grammar.txt
maciejpirog: just regenerated from a modified grammar. no manual changes yet

src/tests/Test.ml
maciejpirog: So far we only have these small smoke tests for lexing and parsing. I will add more comprehensive tests soon

PR #449, opened 2025-10-24
Fix for several methods with the same name

corneliuhoffman:
The algo is as follows:
For each call we look for all the funcs/methods that have this name. If more than one has the same arity with the calling function we ignore the def, warn and behave as in no taint-intrafile. Otherwise we pick the one and continue as before. See the new test to see this in action

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -212,7 +212,6 @@ let any_is_best_sanitizer env any =
 let any_is_best_source ?(is_lval = false) env any =
   env.taint_inst.preds.is_source any
   |> List.filter (fun (m : R.taint_source TM.t) ->
-         (* Remove sources that should match exactly but do not here. *)
          match m.spec.source_by_side_effect with
          | Only -> is_lval && TM.is_exact m
          (* 'Yes' should probably require an exact match like 'Only' but for
@@ -631,7 +630,7 @@ let effects_of_call_func_arg fun_exp fun_shape args_taints =
             (S.show_shape fun_shape));
       []
 
-let get_signature_for_object db method_name object_mappings obj =
+let get_signature_for_object db method_name object_mappings obj arity =
   (* Method call: obj.method() *)
   let obj_str = fst obj.ident in
   (* Use object initialization mappings to determine obj's class *)
@@ -659,10 +658,10 @@ let get_signature_for_object db method_name object_mappings obj =
       let method_sig_id_with_class =
         Shape_and_sig.{ class_name = Some class_name; name = Some method_name }
       in
-      Shape_and_sig.(lookup_signature db method_sig_id_with_class)
+      Shape_and_sig.(lookup_signature db method_sig_id_with_class arity)
   | None -> None
 
-let lookup_signature_with_object_context env fun_exp object_mappings =
+let lookup_signature_with_object_context env fun_exp object_mappings arity =
   match env.signature_db with
   | None ->
       Log.debug (fun m -> m "TAINT_SIG: No signature database available");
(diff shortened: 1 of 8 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -1503,7 +1502,7 @@ let check_function_call env fun_exp args
     (args_taints : (Taints.t * S.shape) argument list) :
     (Taints.t * S.shape * Lval_env.t) option =
   let sig_result =
-    if env.taint_inst.options.taint_intrafile then lookup_signature env fun_exp
+    if env.taint_inst.options.taint_intrafile then lookup_signature env fun_exp (List.length args)
```
dimitris-m, line 1505: Maybe we should add a comment about partial application? For later consideration.

src/tainting/Dataflow_tainting.ml
```diff
@@ -1558,10 +1557,7 @@ let check_function_call env fun_exp args
                      lval_env |> Lval_env.add var offset taints ))
              (Taints.empty, Bot, env.lval_env))
   | None ->
-      Log.info (fun m ->
-          m "No taint signature found for `%s' `%b' "
-            (Display_IL.string_of_exp fun_exp)
-            env.taint_inst.options.taint_intrafile);
+
```
dimitris-m, line 1560: Here we deleted a log, maybe it was useful to have?

corneliuhoffman: moved it to shape_and_sig

src/tainting/Shape_and_sig.ml
```diff
@@ -686,39 +687,90 @@ module FunctionMap = Map.Make (struct
...
+  let signatures = FunctionMap.find_opt func_name db.signatures in
+  match signatures with
+  | Some sigs when not (SignatureSet.is_empty sigs) ->
+      let filtered_sigs =
+        SignatureSet.filter (fun x -> Int.equal arity x.arity) sigs
+      in
+
+      if SignatureSet.cardinal filtered_sigs > 1 then
+        Log.debug (fun m ->
+            m "TAINT_SIG: There are several methods with the name %s the defintion was ignored"
```
dimitris-m: misprint: defintion

src/tainting/Shape_and_sig.ml
```diff
@@ -686,39 +687,90 @@ module FunctionMap = Map.Make (struct
...
+              (show_name func_name.name));
+      if SignatureSet.is_empty filtered_sigs then
+        Log.info (fun m ->
+            m "No taint signature found for `%s' " (show_name func_name.name));
+
+      (* If there is only one signature with the given arity, return it *)
+      if SignatureSet.cardinal filtered_sigs =*= 1 then
```
dimitris-m: Not sure how cheap it is to calculate `SignatureSet.cardinal filtered_sigs`, but it could be done once and reused, before line 732.

dimitris-m: Probably not worth it though, as we expect very small sets anyway.

tests/rules/cross_function_tainting/test_csharp_method_overloading.cs
dimitris-m, line 1: This test is ran with `--taint-intrafile`?

tests/rules/cross_function_tainting/test_csharp_method_overloading.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        this.data = input;
+    }
+
+    static void Main()
+    {
+        C c = new C();
+        string tainted = source();
+        c.Process(tainted);
+        // OK: csharp_method_overload not found, double method with this arity
```
dimitris-m, line 24: How about without `--taint-intrafile`?

Is the default for `c` to not become tainted in such cases, by side-effect for example?

PR #453, opened 2025-10-28
Depth based equality

corneliuhoffman:
We fix a "maximum depth" and decide that two nested objects are the same  (as far as the fixed point is concerned) if they are small and the same or deeper than the max depth and equal up to the max dept.

For example in the code below of course we do not know how big thew input is and so in the existing version the loop never finds a fixed point. We consider that after obtaining an object whose depth is more than 50 probably there will be no more new taints. 
```typescript
function bad(input: string) {
  const obj: any = {};

  for (let item of input) {
    obj[item] = [obj[item], item];
  }

  return obj;
}

sink(bad(source()));
```

### DIFF
--- a/src/tainting/Function_call_graph.ml
+++ b/src/tainting/Function_call_graph.ml
@@ -41,6 +41,7 @@ module FuncVertex = struct
     let cmp_opt = Option.compare compare_il_name in
     let c1 = cmp_opt f1.class_name f2.class_name in
     if c1 <> 0 then c1 else cmp_opt f1.name f2.name
+
   let hash (f : fn_id) =
     let h1 =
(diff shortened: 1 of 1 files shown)


src/tainting/Function_call_graph.ml
```diff
@@ -41,7 +41,6 @@ module FuncVertex = struct
     let cmp_opt = Option.compare compare_il_name in
     let c1 = cmp_opt f1.class_name f2.class_name in
     if c1 <> 0 then c1 else cmp_opt f1.name f2.name
-
```
dimitris-m, line 44: That space makes sense, maybe this is accidental?

corneliuhoffman: yes, accidental

corneliuhoffman: changed

PR #465, opened 2025-11-11
Vb.net: AST generic

maciejpirog:
# Support for VB.NET part 5 (AST generic)

To make the process of reviewing easier, the support for VB.NET will be added in steps. This is step 5: **AST generic**. This PR merges to the `mpir/vbnet` branch. Only when all parts are reviewed, I will merge it with `main`.

Notes:

1. The parser creates the generic AST directly, with no intermediate data types, so it should be a tiny bit easier on the GC compared to other languages,

2. There are still some todos, marked in the code. They mostly concern complicated but not so important details of types (such as type constraints).

3. LINQ is quite different from the one in C# (see https://stackoverflow.com/questions/6515037/differences-in-linq-syntax-between-vb-net-and-c-sharp), so we cannot simply reuse the code from `Parse_csharp_tree_sitter.ml` to generate AST generic. For now, we keep LINQ as `RawTree` with embedded expression. This allows for matching, not sure yet about tainting. To be fixed ASAP.

4. Only to minimal smoke tests are included. More tests to come in a separate PR (when the rest of the infrastructure is plugged in).

5. This PR also provides some minor tweaks to the lexer and the parser.

### DIFF
--- a/languages/vbnet/Vbnet_parser.ml
+++ b/languages/vbnet/Vbnet_parser.ml
@@ -248,6 +248,12 @@ let stmt_of_stmts (xs : G.stmt list) =
 let raw_token (t : T.t) : G.any RT.t =
   RT.Token (t.content, t.tok)
 
+let s_range (tl : Tok.t) (tr : Tok.t) (stmt : G.stmt) : G.stmt =
+  let l = Tok.unsafe_loc_of_tok tl in
+  let r = Tok.unsafe_loc_of_tok tr in
+  stmt.s_range <- Some (l, r);
+  stmt
+
 (* parser *)
 
 let rec compilation_unit : G.program parser = fun __n -> (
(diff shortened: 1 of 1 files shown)


languages/vbnet/Vbnet_parser.ml
```diff
@@ -177,1210 +181,1323 @@ let xOptional (r : G.any option) : G.any =
...
+      let ns, n = List.concat_map idents_of_name ns |> split_last in
+      G.IdQualified
+        { name_last = n;
+          name_middle = Some (QDots ns);
+          name_top = None;
+          name_info = G.empty_id_info ()
+        }
+
+let add_attrs_to_type (t : G.type_) (attrs : G.attribute list) : G.type_=
+  { t with t_attrs = attrs @ t.t_attrs }
```
dimitris-m, line 234: missing space before =

maciejpirog: fixed

PR #467, opened 2025-11-15
Vb.net: tests and fixes

maciejpirog:
# Support for VB.NET part 7 **Tests and fixes**

To make the process of reviewing easier, the support for VB.NET will be added in steps. This is step t:  **Tests and fixes**. This PR merges to the `mpir/vbnet` branch. Only when all parts are reviewed, I will merge it with `main`.

Notes:

- Most of the tests are adjusted tests for C# and Java. Some are for VB-specific constructs, such as the `With` block.

- This PR also contains fixes for the issues that surfaced while adding tests:

    - Bug in `Analyze_rule`, which skipped files that did not contain identifiers from the rule if they were in different case, even for languages that are case insensitive.

    - Mark the variable introduced by `With`  as `hidden` (so that opengrep does not complain about a variable with no range)

    - Fix pattern parsing for assignment/comparison (`$X = $Y` is assignment, `($X = $Y)` is comparison) 

    - Add import statements to parsing of patterns

    - Allow `_` in metavars


tests/patterns/vb/anonymous_metavar.vb
dimitris-m, line 1: Missing .sgrep file?

tests/patterns/vb/dots_method_chaining.vb
dimitris-m, line 1: Missing .sgrep file?

tests/patterns/vb/equal-sign-comparison.sgrep
dimitris-m, line 1: This is a nice approach

dimitris-m (approved):
lgtm

tests/patterns/vb/anonymous_metavar.vb
maciejpirog: Nope, the same test for different languages can share a .sgrep file in the `POLYGLOT` directory. In this case the corresponding .sgrep file is: https://github.com/opengrep/opengrep/blob/main/tests/patterns/POLYGLOT/anonymous_metavar.sgrep

tests/patterns/vb/dots_method_chaining.vb
maciejpirog: Same as above. The file is: https://github.com/opengrep/opengrep/blob/main/tests/patterns/POLYGLOT/dots_method_chaining.sgrep

tests/patterns/vb/equal-sign-comparison.sgrep
maciejpirog: I will add a wiki page with explanation of this and some TODOs

tests/patterns/vb/anonymous_metavar.vb
dimitris-m: I keep forgetting this...

PR #469, opened 2025-11-20
HOFs

corneliuhoffman:
# Higher-Order Function (HOF) Taint Analysis - PR Documentation

## Overview

This PR introduces comprehensive support for tracking taint through higher-order functions (HOFs) in opengrep's cross-function taint analysis. 
**Problem**: Before this PR, taint analysis could not track data flow through HOF calls like:
```javascript
let userInput = getSource();
let results = [userInput].map(x => x);  // Taint was lost here
sink(results[0]);  // Not detected as a vulnerability
```

**Solution**: We now model HOF behavior with `ToSinkInCall` effects that represent "call this callback with tainted data", combined with a call graph that tracks both direct calls and HOF callbacks for proper analysis ordering.

---

### 1. Path-Based Function Identification (`fn_id`)

**File**: `src/tainting/Shape_and_sig.ml:695-710`

**Before**:
```ocaml
type fn_id = { class_name : IL.name option; name : IL.name option }
```

**After**:
```ocaml
type fn_id = IL.name option list
```

**Why**: The old structure couldn't represent nested functions. The new path-based representation supports arbitrary nesting:
- `[None; Some "foo"]` - Top-level function `foo`
- `[Some "MyClass"; Some "method"]` - Method in a class
- `[None; Some "outer"; Some "inner"]` - Nested function `inner` inside `outer`
- `[Some "Class"; Some "method"; Some "_tmp"]` - Lambda inside a class method

**Helper functions added**:
```ocaml
let make_fn_id_with_class class_name method_name = [Some class_name; Some method_name]
let make_fn_id_no_class name = [None; Some name]
let show_fn_id fn_id = ... (* e.g., "MyClass::method::nested" *)
let get_fn_name fn_id = ... (* Extract last element *)
```

---

### 2. Builtin Signature Database

**New File**: `src/tainting/Builtin_models.ml` (471 lines)

Standard library HOFs (like `Array.map`, `filter`, Python's `map()`) don't have user-defined signatures. This module provides built-in models.

**Key concept**: `ToSinkInCall` effect represents "call a callback with tainted data":
```ocaml
Effect.ToSinkInCall {
  callee = <callback expression>;
  arg = { name = "callback"; index = 0 };  (* Which param is the callback *)
  args_taints = [tainted_data];            (* What data to pass *)
}
```

**Three signature patterns**:

1. **MethodHOF** - `arr.map(callback)`: Passes `this` (the array) to the callback
   ```ocaml
   (* Signature models: callback receives BThis (the receiver) *)
   let this_taint = Taint.{ orig = Var { base = BThis; offset = [] }; tokens = [] }
   ```

2. **FunctionHOF** - `map(callback, data)`: Passes one parameter to another
   ```ocaml
   (* Signature models: callback receives data from another parameter *)
   let data_param_taint = Taint.{ orig = Var { base = BArg data_arg; offset = [] }; ... }
   ```

3. **ReturningFunctionHOF** - Ruby's `arr.map` returns a function that takes a block
   ```ocaml
   (* Returns Shape.Fun with nested signature for deferred callback *)
   let return_effect = Effect.ToReturn {
     data_taints = this_taint_set;
     data_shape = Shape.Fun returned_fun_sig;  (* The function shape *)
     ...
   }
   ```

**Supported languages and functions** (`get_hof_configs`):

| Language | Method HOFs | Function HOFs |
|----------|-------------|---------------|
| JavaScript/TypeScript | `map`, `flatMap`, `filter`, `forEach`, `find`, `findIndex`, `some`, `every`, `reduce`, `reduceRight` | - |
| Python | - | `map`, `filter` |
| Ruby | - | `map`, `each`, `select`, `filter`, `flat_map`, `collect`, `find`, `detect` (returning-function pattern) |
| Java | `map`, `filter`, `forEach`, `flatMap` | - |
| Kotlin | `map`, `filter`, `forEach`, `flatMap`, `find`, `any`, `all` (arity 0 and 1) | - |
| Swift | `map`, `filter`, `forEach`, `flatMap`, `compactMap`, `first`, `contains` | - |
| Scala | `map`, `filter`, `foreach`, `flatMap`, `find`, `exists`, `forall` | - |
| C# | `Select`, `Where`, `ForEach`, `SelectMany`, `First`, `Any`, `All` | - |
| Rust | `map`, `for_each`, `filter`, `flat_map`, `find`, `any`, `all` | - |
| PHP | - | `array_map`, `array_filter`, `array_walk` |
| Julia | - | `map`, `foreach`, `filter` |
| C++ | - | `for_each` (arity 3), `transform` (arity 4) |
| Elixir | - | `Enum.map`, `Enum.each`, `Enum.filter`, `Enum.flat_map`, `Enum.find` |

---

### 3. Labeled Call Graph with HOF Edges

**File**: `src/tainting/Function_call_graph.ml` (major rewrite, +869 lines)

**Before**: Simple graph `FuncGraph = Persistent.Digraph.Concrete(FuncVertex)`

**After**: Labeled imperative graph with edge metadata:
```ocaml
type call_edge = {
  callee_fn_id : fn_id;   (* Full path of callee *)
  call_site : Tok.t;      (* Token at the call site *)
}

module FuncGraph = Graph.Imperative.Digraph.ConcreteLabeled(FuncVertex)(...)
```

**Why labeled edges**:
- Multiple calls to the same function from different sites need to be distinguished
- The call site token allows signature instantiation to find the correct callee
- HOF callbacks need different edges than regular calls

**Key functions**:

1. `identify_callee` - Resolves a call expression to its `fn_id`:
   - Checks nested functions in the same scope first
   - Then class methods
   - Then top-level functions
   - Handles `this.method()`, `obj.method()`, qualified names

2. `extract_calls` - Extracts all (fn_id, call_site_tok) pairs from a function body

3. `extract_hof_callbacks` - Extracts HOF callback edges:
   - Uses `Builtin_models.get_hof_configs` to know which methods are HOFs
   - Detects user-defined HOFs via `detect_user_hof`
   - Creates special HOF edges with fake tokens like `<hof_callback:foo>`

4. `detect_user_hof` - Checks if a function calls any of its parameters:
   ```ocaml
   let detect_user_hof (fdef : G.function_definition) : (string * int) list
   ```

---

### 4. Graph Reachability Optimization

**New File**: `src/tainting/Graph_functor.ml` (132 lines)

For large codebases, analyzing every function is expensive. This module computes the minimal subgraph needed for a specific rule.

**Key function**: `nearest_common_descendant_subgraph`
```ocaml
(* Subgraph containing all paths from sources to sinks *)
val nearest_common_descendant_subgraph : graph -> vertex -> vertex -> graph
```

**Algorithm**:
1. Compute forward-reachable sets from source functions (`R1`) and sink functions (`R2`)
2. Intersect to get functions reachable from both (`R1 ∩ R2`)
3. Find minimal SCCs in intersection (handles cycles from mutual recursion)
4. Reverse BFS from minima to include all ancestors

**Usage in Match_tainting_mode.ml**:
```ocaml
let source_functions = find_functions_containing_ranges ast source_ranges in
let sink_functions = find_functions_containing_ranges ast sink_ranges in
let relevant_graph = compute_relevant_subgraph call_graph source_functions sink_functions in
```

---

### 5. Parent Path Visitor

**File**: `src/analyzing/Visit_function_defs.ml` (+167 lines)

**New visitor**: `visitor_with_parent_path` tracks the full path to each function during AST traversal.

```ocaml
class visitor_with_parent_path = object
  val parent_path : IL.name option list ref = ref []
  (* Tracks: [class; outer_func; inner_func; ...] *)
end

val fold_with_parent_path :
  ('acc -> G.entity option -> IL.name option list -> G.function_definition -> 'acc)
  -> 'acc -> G.program -> 'acc
```

**Why**: The old `fold_with_class_context` only tracked class names. For nested functions, we need the full path to build correct `fn_id`s.

---

### 6. ToSinkInCall Effect

**File**: `src/tainting/Sig_inst.ml:38-48, 876-1107`

New effect type added to signature instantiation:
```ocaml
type call_effect =
  | ToReturn of Effect.taints_to_return
  | ToSink of Effect.taints_to_sink
  | ToLval of Taint.taints * IL.name * Taint.offset list
  | ToSinkInCall of {       (* NEW *)
      callee : IL.exp;
      arg : Taint.arg;
      args_taints : Effect.args_taints;
    }
```

**Instantiation logic** (simplified):
1. When encountering `ToSinkInCall` effect during signature instantiation:
2. Look up the callback's signature (from lval_env shape, or via lookup_sig)
3. Instantiate the callback with the provided args_taints
4. If no signature found, preserve the effect for later resolution
5. Add depth limiting to prevent infinite recursion on recursive HOFs

**Key change**: `instantiate_function_signature` now takes optional `lookup_sig` and `depth` parameters:
```ocaml
let rec instantiate_function_signature lval_env taint_sig
    ~callee ~args args_taints
    ?(lookup_sig : (IL.exp -> int -> Signature.t option) option)
    ?(depth : int = 0) () : call_effects option
```

---

### 7. Signature Lookup via Call Graph

**File**: `src/tainting/Dataflow_tainting.ml:667-730`

New function queries the call graph to resolve function calls:
```ocaml
let lookup_callee_from_graph graph caller_fn_id call_tok : fn_id option =
  (* Find edge with matching call_site token *)
  all_edges
  |> List.find_opt (fun edge ->
      Int.equal (Tok.compare label.call_site call_tok) 0)
  |> Option.map (fun edge -> label.callee_fn_id)
```

**Lookup order**:
1. Check call graph for edge with matching call site token
2. If found, use the pre-computed `callee_fn_id` from the edge label
3. Fall back to direct name lookup
4. Fall back to builtin signature database

---

### 8. Ruby Class Method Fix

**File**: `languages/ruby/generic/ruby_to_generic.ml:794-815`

**Bug**: Ruby class methods (`def self.method_name`) were being parsed with empty entity names, causing all class methods to collapse into a single node.

**Before**:
```ocaml
| SingletonM e ->
    let e = expr e in
    let ent = G.basic_entity ("", fake t "") in  (* Empty name! *)
    ...
```

**After**:
```ocaml
| SingletonM e ->
    let method_id =
      match e with
      | DotAccess (_, _, mn) -> (
          match method_name mn with
          | Left id -> id
          | Right _ -> ("", fake t ""))
      | ScopedId (Scope (_, _, SM mn)) -> ...
      | _ -> ("", fake t "")
    in
    let e = expr e in
    let ent = G.basic_entity method_id in  (* Correct name! *)
    ...
```

**Impact**: Call graph now correctly shows `Service.default_integration` and `Service.closest_group_integration` as separate nodes instead of both being `Service.`.

---

### 9. Parameter Shape for HOF Detection

**File**: `src/tainting/Taint_signature_extractor.ml:123-214`

Parameters now get `Shape.Arg` shape instead of just taints:
```ocaml
let add_param_to_env il_lval taint_set taint_arg env =
  let param_shape = Shape.Arg taint_arg in
  Taint_lval_env.add_lval_shape il_lval taint_set param_shape env
```

**Why**: When a parameter is used as a callback in a HOF call, we need to know it's an "argument" to create proper `ToSinkInCall` effects. The `Arg` shape carries the parameter's index.

---

### 10. Match_tainting_mode Integration

**File**: `src/engine/Match_tainting_mode.ml` (+343 lines of changes)

Major refactoring to use the new infrastructure:

1. **Shared call graph**: Computed once per file, shared across rules
   ```ocaml
   let shared_call_graph_opt =
     if taint_intrafile then
       Some (Function_call_graph.build_call_graph ~lang ~object_mappings ast)
     else None
   ```

2. **Relevant subgraph optimization**:
   ```ocaml
   let source_functions = find_functions_containing_ranges ast source_ranges in
   let sink_functions = find_functions_containing_ranges ast sink_ranges in
   let relevant_graph = compute_relevant_subgraph call_graph source_functions sink_functions in
   ```

3. **Topological ordering for analysis**:
   ```ocaml
   let analysis_order = Topo.fold (fun fn acc -> fn :: acc) relevant_graph [] |> List.rev
   ```

4. **Kotlin trailing lambda support**: Functions with lambda as last parameter get signatures extracted at arity and arity-1:
   ```ocaml
   if Lang.equal lang Lang.Kotlin && arity >= 2 then
     let last_param_is_lambda = ... in
     if last_param_is_lambda then
       extract_signature_with_file_context ~arity:(arity - 1) ...
   ```

---

## Data Flow Example

Consider this TypeScript code:
```typescript
function source() { return tainted; }
function sink(x) { /* dangerous */ }
function process(items) {
  return items.map(x => transform(x));
}
function transform(x) { return x; }

let data = [source()];
let result = process(data);
sink(result[0]);
```

**Analysis flow**:

1. **Build call graph**:
   ```
   <toplevel> --calls--> source (at source() call)
   <toplevel> --calls--> process (at process(data) call)
   <toplevel> --calls--> sink (at sink(...) call)
   process --calls--> transform (via map HOF callback edge)
   ```

2. **Compute relevant subgraph**: Functions between source and sink

3. **Topological order**: `transform`, `process`, `<toplevel>` (leaves first)

4. **Extract signatures**:
   - `transform`: `ToReturn { data_taints = {Arg(x, 0)} }` (returns its input)
   - `process`:
     - `map` builtin has `ToSinkInCall { callee = callback, args_taints = [BThis] }`
     - When lambda `x => transform(x)` is analyzed, we look up `transform`'s signature
     - Result: `ToReturn { data_taints = {Arg(items, 0)} }` (returns transformed items)

5. **Instantiate at call sites**:
   - `process(data)` where `data` is tainted from `source()`
   - Substitute `Arg(items, 0)` → taint from `data`
   - Result flows to `sink(result[0])` → **FINDING!**

---

## Test Coverage

**New test files** (17 languages):
- `test_hof_comprehensive_*.{js,ts,py,rb,java,kt,scala,swift,cs,rs,lua,php,go,cpp,c,julia,ex}`

Each test covers:
1. Basic HOF taint propagation (`map`, `filter`, `forEach`)
2. Chained HOFs (`arr.map(...).filter(...)`)
3. Reduce with accumulator
4. Nested callbacks
5. User-defined HOFs (functions that call their parameters)
6. Cross-function taint through HOFs

---

## Configuration

No new configuration options. HOF support is automatically enabled when `taint_intrafile` is true.

---

## Performance Considerations

1. **Call graph is shared**: Computed once per file, reused across all rules
2. **Subgraph optimization**: Only functions on paths between sources and sinks are analyzed
3. **Depth limiting**: Recursive HOF instantiation is limited by `taint_MAX_VISITS_PER_NODE`
4. **Builtin signatures are pre-computed**: Language-specific HOF models are created once

---

## Benchmark Script

A benchmark script is provided at `benchmark_taint.sh` for comparing OpenGrep vs Semgrep taint analysis performance:

```bash
# Usage
./benchmark_taint.sh [--no-semgrep] <yaml_file_or_dir> <target_file_or_dir> [num_runs] [sha1] [sha2] ...

# Examples
./benchmark_taint.sh rules/sqli.yaml target_repo/ 3
./benchmark_taint.sh --no-semgrep rules/ target_repo/ 5
./benchmark_taint.sh rules/ target_repo/ 3 main abc123  # Compare current vs main vs abc123
```

**Options:**
- `--no-semgrep`: Skip Semgrep benchmarks (useful when Semgrep is not installed or for quick OpenGrep-only tests)
- `num_runs`: Number of runs to average (default: 3)
- `sha1, sha2, ...`: Git commits/branches to compare (will checkout, compile, and benchmark each)

**Output:** Summary table with findings count and timing (avg ± std) for each configuration

### DIFF
--- a/languages/elixir/generic/Elixir_to_generic.ml
+++ b/languages/elixir/generic/Elixir_to_generic.ml
@@ -210,8 +210,7 @@ let find_max_placeholder (e : G.expr) : int =
         | G.OtherExpr
             (("PlaceHolder", _), [ G.E { e = G.L (G.Int (Some n, _)); _ } ]) ->
             max_found := max !max_found (Int64.to_int n)
-        | _ -> ());
-        super#visit_expr env expr
+        | _ -> super#visit_expr env expr)
     end
   in
   visitor#visit_expr () e;
--- a/perf/opengrep-scripts/bench.py
+++ b/perf/opengrep-scripts/bench.py
@@ -19,13 +19,13 @@ changes_while_running = False
 # SETUP
 
 # git url, git commit, number of repetitions of the test for a given repo
-repositories = [ ("https://github.com/jellyfin/jellyfin","a0931baa8eb879898f4bc4049176ed3bdb4d80d1", 5)
+repositories = [ ("https://github.com/jellyfin/jellyfin","a0931baa8eb879898f4bc4049176ed3bdb4d80d1", 3)
                , ("https://github.com/grafana/grafana", "afcb55156260fe1887c4731e6cc4c155cc8281a2", 1)
                , ("https://gitlab.com/gitlab-org/gitlab", "915627de697e2dd71fe8205853de51ad3794f3ac", 1)
-               , ("https://github.com/Netflix/lemur", "28b9a73a83d350b1c7ab71fdd739d64eec5d06aa", 5)
-               , ("https://github.com/pythongosssss/ComfyUI-Custom-Scripts","943e5cc7526c601600150867a80a02ab008415e7", 5)
-               , ("https://github.com/pmd/pmd", "81739da5caff948dbcd2136c17532b65c726c781", 5)
-               , ("https://github.com/square/leakcanary", "bf5086da26952e3627f18865bb232963e4d019c5", 5)
+               , ("https://github.com/Netflix/lemur", "28b9a73a83d350b1c7ab71fdd739d64eec5d06aa", 3)
+               , ("https://github.com/pythongosssss/ComfyUI-Custom-Scripts","943e5cc7526c601600150867a80a02ab008415e7", 3)
+               , ("https://github.com/pmd/pmd", "81739da5caff948dbcd2136c17532b65c726c781", 3)
+               , ("https://github.com/square/leakcanary", "bf5086da26952e3627f18865bb232963e4d019c5", 3)
                ]
 
 # Global config set by CLI args
@@ -35,7 +35,34 @@ config = {
     "with_semgrep": False,
     "taint_intrafile": False,
     "reps": None,  # Override default reps if set
+    "timeout": 9,
 }
+
+def scan_cmd(repo, intrafile=False):
+    """Opengrep scan command for repo scanning."""
+    cmd = ["pipenv", "run", "opengrep", "scan",
+            "-c", config["rules"],
+            repo,
+            "-j", num_cpus,
+            "--timeout", str(config["timeout"]),
+            "--max-memory", "4000",
+            "--max-target-bytes", "200000"]
+    if intrafile:
+        cmd.append("--taint-intrafile")
+    return cmd
+
+def semgrep_scan_cmd(repo, pro_intrafile=False):
+    """Semgrep scan command for repo scanning."""
+    cmd = ["semgrep", "scan",
+           "-c", config["rules"],
+           repo,
+           "-j", num_cpus,
+           "--timeout", str(config["timeout"]),
+           "--max-memory", "4000",
+           "--max-target-bytes", "200000"]
+    if pro_intrafile:
+        cmd.append("--pro-intrafile")
+    return cmd
 
 def regular_cmd(repo):
     return ["pipenv", "run", "opengrep", "scan",
@@ -129,8 +156,31 @@ def run(cmd, cwd=None):
     # my_env["PIPENV_PIPFILE"] = "../opengrep/cli/Pipfile"
     print(f"Running: {' '.join(cmd)}")
     sys.stdout.flush()
-    result = subprocess.run(cmd, cwd=cwd, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+    result = subprocess.run(cmd, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+    if result.returncode != 0:
+        print(f"ERROR: Command failed with exit code {result.returncode}")
+        if result.stdout:
+            print(f"STDOUT: {result.stdout}")
+        if result.stderr:
+            print(f"STDERR: {result.stderr}")
+        raise subprocess.CalledProcessError(result.returncode, cmd)
     return hash_output(result.stdout)
+
+def cleanup_rules_dir(rules_dir):
+    """Remove non-rule yaml files from a rules directory."""
+    import glob
+    import shutil
+    # Remove .github directory entirely
+    github_dir = os.path.join(rules_dir, ".github")
+    if os.path.exists(github_dir):
+        shutil.rmtree(github_dir)
+        print(f"  Removed {github_dir}")
+    # Remove common non-rule files
+    for pattern in ["template.yaml", ".pre-commit-config.yaml"]:
+        f = os.path.join(rules_dir, pattern)
+        if os.path.exists(f):
+            os.remove(f)
+            print(f"  Removed {f}")
 
 def clone_specific_commit(repo_url, commit_hash, name):
     if os.path.exists(name):
@@ -148,11 +198,8 @@ def clone_specific_commit(repo_url, commit_hash, name):
 
         print(f"Done: {name} at commit {commit_hash}")
 
-def setup():
-    os.makedirs("results", exist_ok=True)
-
+def setup_opengrep_rules():
     clone_specific_commit("https://github.com/opengrep/opengrep-rules", "f1d2b562b414783763fd02a6ed2736eaed622efa", "rules")
-
     run(["rm", "-f", "stats/web_frameworks.yml"], cwd="rules")
     run(["rm", "-f", "stats/cwe_to_metacategory.yml"], cwd="rules")
     run(["rm", "-f", "stats/metacategory_to_support_tier.yml"], cwd="rules")
@@ -163,7 +210,9 @@ def setup():
     run(["rm", "-f", ".github/workflows/pre-commit.yml"], cwd="rules")
     run(["rm", "-f", "template.yaml"], cwd="rules")
     run(["rm", "-f", ".pre-commit-config.yaml"], cwd="rules")
-
+        
+def setup():
+    os.makedirs("results", exist_ok=True)
     for r in repos:
         clone_specific_commit(r["url"], r["sha"], "repos/" + r["name"])
 
@@ -221,6 +270,13 @@ def single_run_cmd(cmd, label="", run_num=None, total_runs=None, start_time_all=
     start_time = time.time()
     output_data = []
     last_pct = -1
+    spinner_chars = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏']
+    spinner_idx = 0
+    got_first_pct = False
+
+    # Show initial "loading rules" message
+    sys.stdout.write(f"\r  {label}: {spinner_chars[0]} Loading rules...   ")
+    sys.stdout.flush()
 
     # Use pty to get real-time output with progress
     master_fd, slave_fd = pty.openpty()
@@ -247,12 +303,20 @@ def single_run_cmd(cmd, label="", run_num=None, total_runs=None, start_time_all=
                         pct = int(matches[-1])
                         if pct != last_pct:
                             last_pct = pct
+                            got_first_pct = True
                             elapsed = time.time() - start_time
                             bar_width = 30
                             filled = int(bar_width * pct / 100)
                             bar = '█' * filled + '░' * (bar_width - filled)
                             sys.stdout.write(f"\r  {label}: [{bar}] {pct}% {elapsed:.1f}s   ")
                             sys.stdout.flush()
+            else:
+                # No output yet, show spinner
+                if not got_first_pct:
+                    elapsed = time.time() - start_time
+                    spinner_idx = (spinner_idx + 1) % len(spinner_chars)
+                    sys.stdout.write(f"\r  {label}: {spinner_chars[spinner_idx]} Loading rules... {elapsed:.1f}s   ")
+                    sys.stdout.flush()
         except OSError:
             break
 
@@ -375,7 +439,7 @@ def run_custom_benchmark(rules, target, reps, with_semgrep, taint_intrafile, lab
         run_offset += reps
 
         duration, consistent, findings, files = run_with_cmd(
-            semgrep_cmd(rules, target, pro_intrafile=True), reps, f"{label_prefix}semgrep-pro",
+            semgrep_cmd(rules, target, pro_intrafile=True), reps, f"{label_prefix}semgrep (pro-intrafile)",
             run_offset, total_runs, start_time_all)
         results.append({"name": f"{label_prefix}Semgrep (pro-intrafile)", "duration": duration, "consistent": consistent, "findings": findings, "files": files})
         run_offset += reps
@@ -384,14 +448,14 @@ def run_custom_benchmark(rules, target, reps, with_semgrep, taint_intrafile, lab
     duration, consistent, findings, files = run_with_cmd(
         custom_cmd(rules, target), reps, f"{label_prefix}opengrep",
         run_offset, total_runs, start_time_all)
-    results.append({"name": f"{label_prefix}OpenGrep", "duration": duration, "consistent": consistent, "findings": findings, "files": files})
+    results.append({"name": f"{label_prefix}Opengrep", "duration": duration, "consistent": consistent, "findings": findings, "files": files})
     run_offset += reps
 
     if taint_intrafile:
         duration, consistent, findings, files = run_with_cmd(
             custom_cmd(rules, target, intrafile=True), reps, f"{label_prefix}opengrep-intra",
             run_offset, total_runs, start_time_all)
-        results.append({"name": f"{label_prefix}OpenGrep (taint-intrafile)", "duration": duration, "consistent": consistent, "findings": findings, "files": files})
+        results.append({"name": f"{label_prefix}Opengrep (taint-intrafile)", "duration": duration, "consistent": consistent, "findings": findings, "files": files})
         run_offset += reps
 
     return results, run_offset
@@ -443,7 +507,7 @@ def go_custom(args):
             run_offset += reps
 
             duration, consistent, findings, files = run_with_cmd(
-                semgrep_cmd(rules, target, pro_intrafile=True), reps, "semgrep-pro",
+                semgrep_cmd(rules, target, pro_intrafile=True), reps, "semgrep (pro-intrafile)",
                 run_offset, total_runs, start_time_all)
             all_results["semgrep"].append({"name": "Semgrep (pro-intrafile)", "duration": duration, "consistent": consistent, "findings": findings, "files": files})
             run_offset += reps
@@ -513,6 +577,128 @@ def go_custom(args):
         consistent_str = "yes" if r['consistent'] else "no"
         print(f"{r['name']:<{name_width}}  {r['files']:<8.0f}  {r['findings']:<8.0f}  {r['duration']:<8.2f}  {consistent_str}")
 
+def go_default(rules=None):
+    """Default mode: clone repos, run with semgrep-rules, with-semgrep, taint-intrafile."""
+    os.makedirs("results", exist_ok=True)
+    log_to_file(f"DEFAULT PERF TEST {ts}\n\n")
+
+    # Clone repos
+    for r in repos:
+        clone_specific_commit(r["url"], r["sha"], "repos/" + r["name"])
+
+    if rules is None:
+        rules = "rules"
+
+    # Clean up non-rule yaml files
+    print(f"Cleaning up non-rule files in {rules}...")
+    cleanup_rules_dir(rules)
+
+    # Get version info
+    version, git_sha = get_opengrep_version()
+    print(f"\nOpengrep: {version} ({git_sha[:12]})")
+    print(f"Rules: {rules}")
+    print()
+
+    # Build opengrep
+    print("=== Building Opengrep ===")
+    make_opengrep()
+
+    # Run -f -t on each repo
+    all_results = {}
+    for repo_idx, r in enumerate(repos):
+        repo_path = "repos/" + r["name"]
+        reps = r["reps"]
+        print(f"\n=== [{repo_idx+1}/{len(repos)}] {r['name']} ({reps} reps) ===")
+
+        results, _ = run_custom_benchmark(
+            rules, repo_path, reps,
+            with_semgrep=True,
+            taint_intrafile=True,
+            label_prefix=f"{r['name']}: "
+        )
+        all_results[r["name"]] = results
+
+    # Print summary as markdown table
+    print(f"\n## BENCHMARK RESULTS\n")
+
+    # Collect all config names from first repo
+    first_repo = list(all_results.keys())[0]
+    config_names = [r["name"].split(": ")[1] for r in all_results[first_repo]]
+    repo_names = list(all_results.keys())
+
+    # Header row
+    header = "| Metric |"
+    separator = "|--------|"
+    for repo in repo_names:
+        header += f" {repo} |"
+        separator += "--------|"
+    print(header)
+    print(separator)
+
+    # Time section header
+    print("| **Time (seconds)** |" + " |" * len(repo_names))
+
+    # Time rows
+    for i, cfg in enumerate(config_names):
+        row = f"| {cfg} |"
+        for repo in repo_names:
+            val = all_results[repo][i]["duration"]
+            row += f" {val:.1f} |"
+        print(row)
+
+    # Speedup row (if we have both Semgrep and Opengrep)
+    if len(config_names) >= 3:
+        row = "| **Speedup (SG/OG)** |"
+        for repo in repo_names:
+            sg_time = all_results[repo][0]["duration"]
+            og_time = all_results[repo][2]["duration"]
+            speedup = sg_time / og_time if og_time > 0 else 0
+            row += f" {speedup:.2f}x |"
+        print(row)
+
+    # Findings section header
+    print("| **Findings** |" + " |" * len(repo_names))
+
+    # Findings rows
+    for i, cfg in enumerate(config_names):
+        row = f"| {cfg} |"
+        for repo in repo_names:
+            val = int(all_results[repo][i]["findings"])
+            row += f" {val} |"
+        print(row)
+
+    # Save to CSV (transposed: metrics as rows, repos as columns)
+    print(f"\nSaving to results/results-default-{ts}.csv")
+    with open(f"results/results-default-{ts}.csv", 'w', newline='') as csvfile:
+        fieldnames = ["Metric"] + repo_names
+        writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
+        writer.writeheader()
+
+        # Time rows
+        for i, cfg in enumerate(config_names):
+            row = {"Metric": f"{cfg} (time)"}
+            for repo in repo_names:
+                row[repo] = f"{all_results[repo][i]['duration']:.1f}"
+            writer.writerow(row)
+
+        # Speedup row (Semgrep / Opengrep)
+        if len(config_names) >= 3:  # Has both Semgrep and Opengrep
+            row = {"Metric": "Speedup (SG/OG)"}
+            for repo in repo_names:
+                sg_time = all_results[repo][0]["duration"]  # Semgrep
+                og_time = all_results[repo][2]["duration"]  # Opengrep
+                speedup = sg_time / og_time if og_time > 0 else 0
+                row[repo] = f"{speedup:.2f}x"
+            writer.writerow(row)
+
+        # Findings rows
+        for i, cfg in enumerate(config_names):
+            row = {"Metric": f"{cfg} (findings)"}
+            for repo in repo_names:
+                row[repo] = int(all_results[repo][i]["findings"])
+            writer.writerow(row)
+
+
 def go():
     parser = argparse.ArgumentParser(description='Benchmark opengrep performance')
     parser.add_argument('sha1', nargs='?', help='First SHA/branch to compare')
@@ -527,29 +713,23 @@ def go():
     args = parser.parse_args()
 
     # Check for partial custom mode args
-    if args.rules and not args.target:
-        print("Error: -f/--rules requires -t/--target")
-        sys.exit(1)
     if args.target and not args.rules:
         print("Error: -t/--target requires -f/--rules")
         sys.exit(1)
 
+    setup_opengrep_rules()
+        
     # Custom mode: -f and -t specified
     if args.rules and args.target:
         go_custom(args)
         return
 
-    # Original mode: SHA comparison
+    # Default mode: run on all predefined repos
+    # -f without -t = custom rules on all repos
+    # no args = semgrep-rules on all repos
     if not args.sha1:
-        parser.print_help()
-        print("\nExamples:")
-        print("  bench.py <sha1>                              # Compare sha1 vs current")
-        print("  bench.py <sha1> <sha2>                       # Compare sha1 vs sha2")
-        print("  bench.py -f rules.yaml -t target/            # Custom benchmark (current build)")
-        print("  bench.py -f rules.yaml -t target/ -s main    # Custom benchmark comparing main vs current")
-        print("  bench.py -f rules.yaml -t target/ -s sha1 -s sha2  # Compare two SHAs")
-        print("  bench.py -f rules.yaml -t target/ --with-semgrep --taint-intrafile")
-        sys.exit(1)
+        go_default(rules=args.rules)
+        return
 
     setup()
     log_to_file(f"PERF TEST {ts}\n\n")
--- a/src/analyzing/Visit_function_defs.ml
+++ b/src/analyzing/Visit_function_defs.ml
@@ -80,7 +80,8 @@ class ['self] visitor_with_class_context =
               (* Go into nested functions but do NOT revisit the function definition again! *)
               let body = H.funcbody_to_stmt fdef.G.fbody in
               self#visit_stmt f body
-          | G.DefStmt (ent, G.VarDef { vinit = Some { e = G.Lambda fdef; _ }; _ }) ->
+          | G.DefStmt
+              (ent, G.VarDef { vinit = Some { e = G.Lambda fdef; _ }; _ }) ->
               (* Handle lambda assignments in class fields *)
               f (Some ent) !current_class fdef;
               let body = H.funcbody_to_stmt fdef.G.fbody in
@@ -137,86 +138,56 @@ let fold_with_class_context
    - Method: [Some class_name]
    - Nested function: [None; Some parent_func; Some nested_func] (excluding current function)
 *)
+
+let append_to_parrent_path parent_path class_il func_il =
+  let visitor_parent_path =
+    if parent_path = [] then [ class_il ] else parent_path
+  in
+  let current_fn_id = visitor_parent_path @ [ func_il ] in
+  (visitor_parent_path, current_fn_id)
+
 class ['self] visitor_with_parent_path =
   object (self : 'self)
     inherit [_] G.iter_no_id_info as super
     val current_class : G.name option ref = ref None
     val parent_path : IL.name option list ref = ref []
 
-    (* Helper to convert G.name to IL.name *)
-    method private g_name_to_il_name (g_name : G.name) : IL.name option =
-      match g_name with
-      | G.Id ((str, tok), id_info) ->
-          let id_info = { id_info with G.id_resolved = ref None } in
-          Some IL.{ ident = (str, tok); sid = G.SId.unsafe_default; id_info }
-      | _ -> None
-
-    (* Helper to get IL.name from entity *)
-    method private entity_to_il_name (ent : G.entity) : IL.name option =
-      match ent.G.name with
-      | G.EN name -> self#g_name_to_il_name name
-      | _ -> None
-
     method! visit_definition f ((ent, def_kind) as def) =
       match def_kind with
       | G.ClassDef _cdef ->
-          let old_class = !current_class in
-          (current_class :=
-             match ent.name with
-             | EN name -> Some name
-             | _ -> None);
-          super#visit_definition f def;
-          current_class := old_class
+          let newv =
+            match ent.name with
+            | EN name -> Some name
+            | _ -> None
+          in
+          Common.save_excursion_unsafe current_class newv (fun () ->
+              super#visit_definition f def)
       | G.FuncDef fdef ->
           (* Build fn_id path: [class_option; ...parent_path...; current_func] *)
-          let class_il = Option.bind !current_class self#g_name_to_il_name in
-          let func_il = self#entity_to_il_name ent in
-          let current_fn_id =
-            if !parent_path = [] then
-              (* Top-level or class method *)
-              [class_il; func_il]
-            else
-              (* Nested function - append to parent path *)
-              !parent_path @ [func_il]
-          in
+          let class_il = Option.map AST_to_IL.var_of_name !current_class in
+          let func_il = AST_to_IL.name_of_entity ent in
 
           (* Call the visitor function with parent path (without current function) *)
-          let visitor_parent_path =
-            if !parent_path = [] then [class_il] else !parent_path
+          let visitor_parent_path, current_fn_id =
+            append_to_parrent_path !parent_path class_il func_il
           in
           f (Some ent) visitor_parent_path fdef;
 
           (* Push current function onto path stack for nested functions *)
-          let old_path = !parent_path in
-          parent_path := current_fn_id;
-
-          (* Go into nested functions *)
-          let body = H.funcbody_to_stmt fdef.G.fbody in
-          self#visit_stmt f body;
-
-          (* Restore path *)
-          parent_path := old_path
+          Common.save_excursion_unsafe parent_path current_fn_id (fun () ->
+              let body = H.funcbody_to_stmt fdef.G.fbody in
+              super#visit_stmt f body)
       | G.VarDef { vinit = Some { e = G.Lambda fdef; _ }; _ } ->
           (* Handle lambda assignments like: const f = () => {...} *)
-          let class_il = Option.bind !current_class self#g_name_to_il_name in
-          let func_il = self#entity_to_il_name ent in
-          let current_fn_id =
-            if !parent_path = [] then
-              [class_il; func_il]
-            else
-              !parent_path @ [func_il]
-          in
-
-          let visitor_parent_path =
-            if !parent_path = [] then [class_il] else !parent_path
+          let class_il = Option.map AST_to_IL.var_of_name !current_class in
+          let func_il = AST_to_IL.name_of_entity ent in
+          let visitor_parent_path, current_fn_id =
+            append_to_parrent_path !parent_path class_il func_il
           in
           f (Some ent) visitor_parent_path fdef;
-
-          let old_path = !parent_path in
-          parent_path := current_fn_id;
-          let body = H.funcbody_to_stmt fdef.G.fbody in
-          self#visit_stmt f body;
-          parent_path := old_path
+          Common.save_excursion_unsafe parent_path current_fn_id (fun () ->
+              let body = H.funcbody_to_stmt fdef.G.fbody in
+              self#visit_stmt f body)
       | __else__ -> super#visit_definition f def
 
     method! visit_field f field =
@@ -224,54 +195,34 @@ class ['self] visitor_with_parent_path =
       | G.F stmt -> (
           match stmt.G.s with
           | G.DefStmt (ent, G.FuncDef fdef) ->
-              let class_il = Option.bind !current_class self#g_name_to_il_name in
-              let func_il = self#entity_to_il_name ent in
-              let current_fn_id =
-                if !parent_path = [] then
-                  [class_il; func_il]
-                else
-                  !parent_path @ [func_il]
-              in
-
-              let visitor_parent_path =
-                if !parent_path = [] then [class_il] else !parent_path
+              let class_il = Option.map AST_to_IL.var_of_name !current_class in
+              let func_il = AST_to_IL.name_of_entity ent in
+              let visitor_parent_path, current_fn_id =
+                append_to_parrent_path !parent_path class_il func_il
               in
               f (Some ent) visitor_parent_path fdef;
-
-              let old_path = !parent_path in
-              parent_path := current_fn_id;
-              let body = H.funcbody_to_stmt fdef.G.fbody in
-              self#visit_stmt f body;
-              parent_path := old_path
-          | G.DefStmt (ent, G.VarDef { vinit = Some { e = G.Lambda fdef; _ }; _ }) ->
-              let class_il = Option.bind !current_class self#g_name_to_il_name in
-              let func_il = self#entity_to_il_name ent in
-              let current_fn_id =
-                if !parent_path = [] then
-                  [class_il; func_il]
-                else
-                  !parent_path @ [func_il]
-              in
-
-              let visitor_parent_path =
-                if !parent_path = [] then [class_il] else !parent_path
+              Common.save_excursion_unsafe parent_path current_fn_id (fun () ->
+                  let body = H.funcbody_to_stmt fdef.G.fbody in
+                  self#visit_stmt f body)
+          | G.DefStmt
+              (ent, G.VarDef { vinit = Some { e = G.Lambda fdef; _ }; _ }) ->
+              let class_il = Option.map AST_to_IL.var_of_name !current_class in
+              let func_il = AST_to_IL.name_of_entity ent in
+              let visitor_parent_path, current_fn_id =
+                append_to_parrent_path !parent_path class_il func_il
               in
               f (Some ent) visitor_parent_path fdef;
-
-              let old_path = !parent_path in
-              parent_path := current_fn_id;
-              let body = H.funcbody_to_stmt fdef.G.fbody in
-              self#visit_stmt f body;
-              parent_path := old_path
+              Common.save_excursion_unsafe parent_path current_fn_id (fun () ->
+                  let body = H.funcbody_to_stmt fdef.G.fbody in
+                  self#visit_stmt f body)
           | _ -> super#visit_field f field)
 
     method! visit_function_definition f fdef =
       (* Anonymous nested functions *)
       let visitor_parent_path =
         if !parent_path = [] then
-          [Option.bind !current_class self#g_name_to_il_name]
-        else
-          !parent_path
+          [ Option.map AST_to_IL.var_of_name !current_class ]
+        else !parent_path
       in
       f None visitor_parent_path fdef;
       (* No path change for anonymous functions - they don't add to the path *)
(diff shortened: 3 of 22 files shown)


languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -559,8 +559,43 @@ and map_expr env v : G.expr =
...
+                frettype = None;
+                fkind = (G.LambdaKind, tamp);
+                fbody = G.FBStmt body_stmt;
+              } in
+              G.Lambda fdef |> G.e
+          | _ ->
+              (* Fallback for other captures like &(&1/2) or malformed *)
+              G.OtherExpr (("Capture", tamp), [ G.E e ]) |> G.e)
+      | _ ->
+          (* For other capture forms (e.g., &(&1 + &2)), keep as OtherExpr *)
```
dimitris-m: This is also a lambda? If so, this means we won't be able to handle these in tainting?

corneliuhoffman: No it become a small lambda and it is changed to a lambda by ast_to _ILa d yes we can use in taints, you can see an example in the elixir test

src/core_scan/Core_scan.ml
```diff
@@ -490,6 +490,12 @@ let errors_of_timeout_or_memory_exn (exn : exn) (target : Target.t) : ESet.t =
         (E.mk_error
            ?rule_id:(TLS.get_default ~default:(fun () -> None) Rule.last_matched_rule)
            ~loc Out.StackOverflow)
+  | Memory_limit.ExceededMemoryLimit _ ->
+      Logs.warn (fun m -> m "ExceededMemoryLimit on %s" (Origin.to_string origin));
+      ESet.singleton
+        (E.mk_error
+           ?rule_id:(TLS.get_default ~default:(fun () -> None) Rule.last_matched_rule)
+           ~loc Out.OutOfMemory)
```
dimitris-m, line 498: Was this a bug, not having that case now that we used memprof-limits?

corneliuhoffman: possibly, was not sure what was killing my files so put it in just in case

src/matching/Generic_vs_generic.ml
```diff
@@ -3076,6 +3076,18 @@ and m_definition_kind a b =
   (* boilerplate *)
   | G.EnumEntryDef a1, B.EnumEntryDef b1 -> m_enum_entry_definition a1 b1
   | G.FuncDef a1, B.FuncDef b1 -> m_function_definition a1 b1
+  (* iso: FuncDef pattern can match VarDef with Lambda (arrow function) *)
+  | G.FuncDef a1, B.VarDef { B.vinit = Some { e = B.Lambda b1; _ }; _ } ->
+      if_config
+        (fun x -> x.arrow_is_function)
+        ~then_:(m_function_definition a1 b1)
+        ~else_:(fail ())
```
dimitris-m, line 3098: So we don't need the inverse, `G.VarDef a1, B.FuncDef ...`?

src/matching/Generic_vs_generic.ml
```diff
@@ -3076,6 +3076,18 @@ and m_definition_kind a b =
...
+  (* iso: FuncDef pattern can match FieldDefColon with Lambda (arrow function in object literal) *)
+  | G.FuncDef a1, B.FieldDefColon { B.vinit = Some { e = B.Lambda b1; _ }; _ } ->
+      if_config
+        (fun x -> x.arrow_is_function)
+        ~then_:(m_function_definition a1 b1)
+        ~else_:(fail ())
```
dimitris-m, line 3105: Why so? This is not so obvious.

src/tainting/Dataflow_tainting.ml
```diff
@@ -618,105 +625,280 @@ let effects_of_tainted_return env taints shape return_tok : Effect.t list =
...
+  else
+    match fun_shape with
+    | S.Arg fun_arg ->
+        [ Effect.ToSinkInCall { callee = fun_exp; arg = fun_arg; args_taints } ]
+    | __else__ ->
+        Log.debug (fun m ->
+            m "Function (?) %s has shape %s"
+              (Display_IL.string_of_exp fun_exp)
+              (S.show_shape fun_shape));
+        []
```
dimitris-m: Question: this was not always [] when not intrafile, how come this works?

src/matching/Generic_vs_generic.ml
corneliuhoffman: This is for js things like {a: (params) -> ….} to be recognised as fun a

src/matching/Generic_vs_generic.ml
corneliuhoffman: Oh indeed, great catch will change

src/analyzing/AST_to_IL.ml
```diff
@@ -880,16 +880,66 @@ and expr_aux env ?(void = false) g_expr =
...
+  | G.Lambda fdef -> (
+      (* Elixir implicit parameter lambda conversion:
+       * Elixir lambdas like `fn x -> sink(x) end` are parsed with an implicit parameter
+       * and a Switch body. Convert them to explicit parameter lambdas. *)
+      let converted_fdef =
+        match (env.lang, Tok.unbracket fdef.fparams, fdef.fbody) with
+        | ( Lang.Elixir,
+            [ G.Param { pname = Some (param_name, _); _ } ],
+            G.FBStmt { s = G.Switch (_, Some (G.Cond _), cases); _ } )
+          when param_name =*= "!_implicit_param!" && List.length cases =*= 1 -> (
```
dimitris-m: I recommend to not hardcode `!_implicit_param!`, this will change soon.

Is there a way to avoid depending on this string? All Elixir `fn` have this encoding anyway! Maybe `List.length` is enough.

What I mean by that is that whenever the `fdef.body` is a Switch (and it always is for `Lambda` in Elixir, right?) then `param_name` is `!_implicit_param!`.

So the whole test can be simplified.

src/analyzing/AST_to_IL.ml
```diff
@@ -1009,6 +1059,66 @@ and expr_aux env ?(void = false) g_expr =
...
+              match arg with
+              | G.Arg e | G.ArgKwd (_, e) -> max acc (find_max_placeholder e)
+              | _ -> acc
+            ) 0 in
+            max callee_max args_max
+        | _ -> 0
+      in
+      let max_placeholder = find_max_placeholder body_expr in
+      if max_placeholder =*= 0 then
+        (* No placeholders, treat as fixme *)
```
dimitris-m: What is the semantics of no placeholders? Is it like a thunk or?
If so why fixme?

src/analyzing/AST_to_IL.ml
dimitris-m: By the way, I wonder if these language-specific Generic AST => Generic AST mappings, meaning this and also the case below for `G.OtherExpr (("ShortLambda", tok), [ G.E body_expr ])` and any other such cases, really belong in a separate module.

The logic would remain the same but maybe it would be cleaner....

src/matching/Generic_vs_generic.ml
dimitris-m: So again we need also the case `G.FieldDefColon`, `B.FuncDef` I think.

benchmark_taint.sh
dimitris-m, line 1: I think @maciejpirog 's bench scripts have a lot of error checking that we worked on. Also dealing with submodules which can change between revisions! 

Btw the script file location is top-level?

So I recommend to improve the existing bench scripts as needed and then delete this file once done.

src/matching/Generic_vs_generic.ml
corneliuhoffman: indeed. my (hurried) point wad to match fucntion (... $S, ....) kindo of things. but indeed arrows might also be caught

src/tainting/Dataflow_tainting.ml
corneliuhoffman: not sure I understand this

src/analyzing/AST_to_IL.ml
dimitris-m: As a side comment, knowing of course you have to do this here to catch taints, this kind of thing introduces tech debt....

Whenever the elixir translation changes, this also has to change! And it might be forgotten....

Not saying not to keep it, but it's one of the reasons I'd rather see that stuff in a separated module that we can "remember" to update somehow.

src/tainting/Dataflow_tainting.ml
dimitris-m: The original code was the `else` branch which would record the effect for `S.Arg` even when not in intrafile mode.

Or am I missing something?

src/analyzing/AST_modifications.ml
```diff
@@ -0,0 +1,145 @@
...
+  let body_stmt = G.ExprStmt (replaced_body, G.sc) |> G.s in
+  let fdef =
+    {
+      G.fparams = Tok.unsafe_fake_bracket params;
+      frettype = None;
+      fkind = (G.LambdaKind, tok);
+      fbody = G.FBStmt body_stmt;
+    }
+  in
+  G.Lambda fdef |> G.e
```
dimitris-m: I don't think this works as intended, compositionally.

What if the placeholder appears nested deep inside an expression?

Can we just declare parameters with existing "&1" etc? Avoiding replacement? 
Don't see the reason for replacement.

dimitris-m: For example something along these lines would not translate proper: `&(if cond then foo(&1, &3) else E)`

src/analyzing/AST_modifications.ml
```diff
@@ -0,0 +1,145 @@
...
+        args
+        |> List.fold_left
+             (fun acc arg ->
+               match arg with
+               | G.Arg e | G.ArgKwd (_, e) -> max acc (find_max_placeholder e)
+               | _ -> acc)
+             0
+      in
+      max callee_max args_max
+  | _ -> 0
```
dimitris-m: As you pointed out offline, this requires a visitor,

src/analyzing/AST_modifications.ml
dimitris-m: Another possible issue: naming. Adding parameters won't identify them with instances of `&n` in the body perhaps?

So the `&n` in the body must be related to these new parameters.

PS Now that you pointed out to me that `&n` is not an Id in the body, there is also this issue...

dimitris-m: Stuff in `AST_modifications` might need another naming pass plus other modifications.

corneliuhoffman: done now

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -177,6 +177,66 @@ let args_of_do_block_opt (blopt : do_block_generic option) : G.argument list =
...
+            max_found := max !max_found (Int64.to_int n)
+        | _ -> ());
+        super#visit_expr env expr
+    end
+  in
+  visitor#visit_expr () e;
+  !max_found
+
+(** Replace placeholder references with parameter names.
+    E.g., &1 becomes &1, &2 becomes &2, etc. Recursively replaces in all subexpressions. *)
```
dimitris-m: The E.g. is not needed.

src/matching/Generic_vs_generic.ml
```diff
@@ -3076,6 +3076,20 @@ and m_definition_kind a b =
...
+  (* iso: FuncDef pattern can match VarDef with Lambda (arrow function) *)
+  | G.FuncDef a1, B.VarDef { B.vinit = Some { e = B.Lambda b1; _ }; _ }
+  | B.VarDef { B.vinit = Some { e = B.Lambda b1; _ }; _ }, G.FuncDef a1  ->
+      if_config
+        (fun x -> x.arrow_is_function)
+        ~then_:(m_function_definition a1 b1)
+        ~else_:(fail ())
+  (* iso: FuncDef pattern can match FieldDefColon with Lambda (arrow function in object literal) *)
+  | G.FuncDef a1, B.FieldDefColon { B.vinit = Some { e = B.Lambda b1; _ }; _ }
+  |B.FieldDefColon { B.vinit = Some { e = B.Lambda b1; _ }; _ }, G.FuncDef a1 ->
```
dimitris-m: missing space after |

tests/rules/cross_function_tainting/test_hof_comprehensive_apex.cls
```diff
@@ -0,0 +1,27 @@
+// HOF test for Apex: Custom higher-order functions using simple loops
+// This tests taint flow through array iteration patterns
```
dimitris-m: Where is the HOF here?

src/analyzing/AST_modifications.ml
```diff
@@ -0,0 +1,94 @@
...
+      ] ->
+          (* Convert to lambda with explicit parameter from the pattern *)
+          let new_param = G.Param (G.param_of_id id) in
+          {
+            fdef with
+            fparams = Tok.unsafe_fake_bracket [ new_param ];
+            fbody = G.FBStmt body_stmt;
+          }
+      | _ -> fdef)
+  | _ -> fdef
```
dimitris-m: No longer needed, since https://github.com/opengrep/opengrep/pull/483.

dimitris-m: In fact the encoding of lambdas changed and now used normal Pat*.

src/analyzing/AST_to_IL.ml
```diff
@@ -880,16 +880,31 @@ and expr_aux env ?(void = false) g_expr =
...
   | G.Comprehension _ -> todo env.stmts (G.E g_expr)
+  | G.Lambda fdef when env.lang =*= Lang.Elixir ->
+      (* Elixir implicit parameter lambda conversion *)
+      let converted_fdef = AST_modifications.convert_elixir_implicit_lambda fdef in
+      let lval = fresh_lval (snd converted_fdef.fkind) in
+      let _, final_fdef =
+        function_definition { env with stmts = [] } converted_fdef
+      in
+      let env = add_instr env (mk_i (AssignAnon (lval, Lambda final_fdef)) eorig) in
+      (env, mk_e (Fetch lval) eorig)
```
dimitris-m: No longer needed.

languages/typescript/tree-sitter/Parse_typescript_tree_sitter.ml
```diff
@@ -343,23 +343,19 @@ let anon_choice_COMMA_5194cb4 (env : env) (x : CST.anon_choice_COMMA_5194cb4) =
...
-  match type_or_typeof with
-  | Some _ -> (* TODO: 'type foo', 'typeof foo' *) None
-  | None ->
-      let expr_id = identifier env v2 in
-      Some (expr_id, opt_as_id)
+  (* Note: We now include type-only imports (e.g., `import { type Foo }`)
+     in the AST so that pattern matching with ellipsis works correctly.
+     Previously these were skipped with a TODO comment. *)
+  let expr_id = identifier env v2 in
+  Some (expr_id, opt_as_id)
```
dimitris-m, line 359: Two questions: 
- Why return option, now you always have a value.
- What is an example of what we can do now with this change.

languages/elixir/tree-sitter/Parse_elixir_tree_sitter.ml
```diff
@@ -657,7 +657,36 @@ and map_capture_expression (env : env) (x : CST.capture_expression) =
...
+                in
+                (* Create the call: fun_name(&1, &2, ...) *)
+                let call_expr = Call (actual_fun_name, (tand, (placeholder_args, []), tand), None) in
+                (* Convert to ShortLambda: &(fun(&1, &2, ...)) *)
+                ShortLambda (tand, (tand, call_expr, tand))
+            | _ ->
+                (* Fallback for other captures like &(&1/2) *)
+                Capture (tand, e))
+        | _ ->
+            (* For other capture forms (e.g., &(&1 + &2)), keep as Capture *)
```
dimitris-m, line 688: Here what is meant is `&1` etc as `Capture...` right?

Also what is `&(&1/2)`

src/core_scan/Core_scan.ml
```diff
@@ -561,14 +567,17 @@ let iter_targets_and_get_matches_and_exn_to_errors
...
                    as exn ->
                      log_critical_exn_and_last_rule ();
                      let errors = errors_of_timeout_or_memory_exn exn target in
                      (* we got an exn on the target so definitely we tried to
                       * process the target
                       *)
                      let scanned = true in
+                     Logs.warn (fun m ->
+                         m "File dropped %s: %s"
+                           !!internal_path (Printexc.to_string exn));
```
dimitris-m: Is that left intentionally or for debugging?

languages/elixir/tree-sitter/Parse_elixir_tree_sitter.ml
corneliuhoffman: parses ok as:

```ocaml
 Pr(
       [ExprStmt(
          OtherExpr(("ShortLambda", ()),
            [Params(
               [Param(
                  {pname=Some(("&1", ())); pdefault=None; ptype=None; pattrs=[
                   ];
                   pinfo={id_flags=Ref(0); id_flags=Ref(0);
                          id_resolved_alternative=Ref([]);
                          id_resolved=Ref(Some((Parameter, 1)));
                          id_type=Ref(None); id_svalue=Ref(None); };
                   })]);
             S(
               ExprStmt(
                 Call(IdSpecial((Op(Div), ())),
                   [Arg(
                      N(
                        Id(("&1", ()),
                          {id_flags=Ref(0); id_flags=Ref(0);
                           id_resolved_alternative=Ref([]);
                           id_resolved=Ref(Some((Parameter, 1)));
                           id_type=Ref(None); id_svalue=Ref(None); })));
                    Arg(L(Int((Some(2), ()))))]), ()))]), ())])
```

languages/typescript/tree-sitter/Parse_typescript_tree_sitter.ml
corneliuhoffman: I kept the type from before, added a little explanation on what we can do now

src/engine/Match_tainting_mode.ml
```diff
@@ -472,56 +474,141 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
 
+          (* Use shared call graph if provided, otherwise compute it *)
           let call_graph =
-            Function_call_graph.build_call_graph ~lang ~object_mappings ast
+            match shared_call_graph with
+            | Some (graph, shared_mappings) ->
+                (* Verify that object_mappings match to ensure consistency *)
+                if
+                  List.length shared_mappings
+                  <> List.length all_object_mappings
```
dimitris-m: or make the match case :

```
| ..... when Int.equal (List.length shared_mappings) (List.length all_object_mappings)
```

But why could this happen? 
Under which conditions?

dimitris-m: Also is length enough to discriminate?
@maciejpirog comment.

src/engine/Match_tainting_mode.ml
```diff
@@ -724,19 +800,59 @@ let check_rules ~match_hook
...
-                 ^ (rule.id |> fst |> Rule_ID.to_string))
-               (fun () ->
-                 check_rule per_file_formula_cache rule match_hook xconf xtarget)))
+  (* Pre-compute call graph if taint_intrafile is enabled for any rule.
+     The call graph only depends on the AST structure, not on individual rules,
+     so we can compute it once and share it across all rules for this file. *)
+  let shared_call_graph_opt =
+    match rules with
+    | [] -> None
+    | rule :: _ ->
```
dimitris-m: What if taint-intrafile is only set for some rules (and it may be not the first rule)? 
This is the case if the flag `--taint-intrafile` is not passed.

perf/opengrep-scripts/bench.py
```diff
@@ -16,28 +17,110 @@
...
+    "rules": "rules",
+    "target": None,
+    "with_semgrep": False,
+    "taint_intrafile": False,
+    "reps": None,  # Override default reps if set
+    "timeout": 9,
+}
+
+def scan_cmd(repo, intrafile=False):
+    """OpenGrep scan command for repo scanning."""
```
dimitris-m: OpenGrep => Opengrep (everywhere)

src/analyzing/Visit_function_defs.ml
```diff
@@ -130,3 +130,170 @@ let fold_with_class_context
...
+    method! visit_definition f ((ent, def_kind) as def) =
+      match def_kind with
+      | G.ClassDef _cdef ->
+          let old_class = !current_class in
+          (current_class :=
+             match ent.name with
+             | EN name -> Some name
+             | _ -> None);
+          super#visit_definition f def;
+          current_class := old_class
```
dimitris-m: There's `Common.save_excursion_unsafe` for this type of pattern. (And `Common.save_excursion` for concurrent use when the ref is in DLS).

src/analyzing/Visit_function_defs.ml
```diff
@@ -130,3 +130,170 @@ let fold_with_class_context
...
+    inherit [_] G.iter_no_id_info as super
+    val current_class : G.name option ref = ref None
+    val parent_path : IL.name option list ref = ref []
+
+    (* Helper to convert G.name to IL.name *)
+    method private g_name_to_il_name (g_name : G.name) : IL.name option =
+      match g_name with
+      | G.Id ((str, tok), id_info) ->
+          let id_info = { id_info with G.id_resolved = ref None } in
+          Some IL.{ ident = (str, tok); sid = G.SId.unsafe_default; id_info }
```
dimitris-m: Why do we need to ignore id_resolved and use sid = unsafe_default? 
Not immediately obvious.

src/analyzing/Visit_function_defs.ml
```diff
@@ -130,3 +130,170 @@ let fold_with_class_context
...
+            else
+              (* Nested function - append to parent path *)
+              !parent_path @ [func_il]
+          in
+
+          (* Call the visitor function with parent path (without current function) *)
+          let visitor_parent_path =
+            if !parent_path = [] then [class_il] else !parent_path
+          in
+          f (Some ent) visitor_parent_path fdef;
```
dimitris-m: Isn't that equivalent?

```ocaml
          let visitor_parent_path =
            if !parent_path = [] then [class_il] else !parent_path
          in

         let current_fn_id = visitor_parent_path @ [func_il]
         in

          (* Call the visitor function with parent path (without current function) *)
          f (Some ent) visitor_parent_path fdef;
```

src/analyzing/Visit_function_defs.ml
```diff
@@ -130,3 +130,170 @@ let fold_with_class_context
...
+            if !parent_path = [] then
+              [class_il; func_il]
+            else
+              !parent_path @ [func_il]
+          in
+
+          let visitor_parent_path =
+            if !parent_path = [] then [class_il] else !parent_path
+          in
+          f (Some ent) visitor_parent_path fdef;
```
dimitris-m: As above

src/analyzing/Visit_function_defs.ml
```diff
@@ -130,3 +130,170 @@ let fold_with_class_context
...
+                if !parent_path = [] then
+                  [class_il; func_il]
+                else
+                  !parent_path @ [func_il]
+              in
+
+              let visitor_parent_path =
+                if !parent_path = [] then [class_il] else !parent_path
+              in
+              f (Some ent) visitor_parent_path fdef;
```
dimitris-m: As above can be simplified?

src/analyzing/Visit_function_defs.ml
```diff
@@ -130,3 +130,170 @@ let fold_with_class_context
...
+
+              let old_path = !parent_path in
+              parent_path := current_fn_id;
+              let body = H.funcbody_to_stmt fdef.G.fbody in
+              self#visit_stmt f body;
+              parent_path := old_path
```
dimitris-m: Is there something in common with `visit_definition` that we could factor out?

src/engine/Match_tainting_mode.ml
```diff
@@ -335,19 +345,18 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
       record_matches glob_effects;
 
       (* Only use signature database if cross-function taint analysis is enabled *)
-      let final_signature_db =
+      let final_signature_db, relevant_graph =
         if taint_inst.options.taint_intrafile then (
+          Log.debug (fun m -> m "INTRAFILE_MODE: Entering taint_intrafile code path");
```
dimitris-m: Is that intentionally left in the logs?

src/engine/Match_tainting_mode.ml
```diff
@@ -472,56 +480,133 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
+            in
+            Log.debug (fun m -> m "SUBGRAPH: source_function[%d] = %s" i name))
+            source_functions;
+          List.iteri (fun i fn_id ->
+            let name = match Shape_and_sig.get_fn_name fn_id with
+              | Some n -> fst n.IL.ident
+              | None -> "<no-name>"
+            in
+            Log.debug (fun m -> m "SUBGRAPH: sink_function[%d] = %s" i name))
+            sink_functions;
```
dimitris-m: For performance reasons, it's best to only make these calculations inside the `fun m -> ...` since they are not needed outside of the debug context.

src/analyzing/Visit_function_defs.ml
corneliuhoffman: refactored

src/engine/Match_tainting_mode.ml
```diff
@@ -472,56 +480,133 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
+             * f(a, b) vs f(a) { b } *)
+            let updated_db =
+              if Lang.equal lang Lang.Kotlin && arity >= 2 then
+                let last_param_is_lambda =
+                  match List.rev params with
+                  | G.Param { G.ptype = Some ptype; _ } :: _ -> (
+                      match ptype.G.t with
+                      | G.TyFun _ -> true
+                      | _ -> false)
+                  | _ -> false
```
dimitris-m: Why not just: 

```ocaml
                 | G.Param { G.ptype = Some {t = G.TyFun _; _}; _ } :: _ -> true
                 | _ -> false
```

src/engine/Match_tainting_mode.ml
```diff
@@ -472,56 +480,133 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
+                ~method_properties:info.method_properties
+                ~call_graph:(Some relevant_graph)
+                info.cfg ast
             in
-            if info.is_lambda_assignment then updated_db
+            (* For Kotlin, if the last parameter is a lambda (function type),
+             * also extract signature with arity-1 to handle trailing lambda syntax:
+             * f(a, b) vs f(a) { b } *)
+            let updated_db =
+              if Lang.equal lang Lang.Kotlin && arity >= 2 then
```
dimitris-m: Why not `&& arity >= 1` ?

Is it not the case that `f(a)` where `a` is a function is also `f  { a }` ?

corneliuhoffman: done

src/matching/Generic_vs_generic.ml
```diff
@@ -3089,6 +3089,20 @@ and m_definition_kind a b =
...
+  (* iso: FuncDef pattern can match VarDef with Lambda (arrow function) *)
+  | G.FuncDef a1, B.VarDef { B.vinit = Some { e = B.Lambda b1; _ }; _ }
+  | B.VarDef { B.vinit = Some { e = B.Lambda b1; _ }; _ }, G.FuncDef a1  ->
+      if_config
+        (fun x -> x.arrow_is_function)
+        ~then_:(m_function_definition a1 b1)
+        ~else_:(fail ())
+  (* iso: FuncDef pattern can match FieldDefColon with Lambda (arrow function in object literal) *)
+  | G.FuncDef a1, B.FieldDefColon { B.vinit = Some { e = B.Lambda b1; _ }; _ }
+  | B.FieldDefColon { B.vinit = Some { e = B.Lambda b1; _ }; _ }, G.FuncDef a1 ->
```
dimitris-m: I think this is wrong: we are inverting the order of arguments, left is Pattern (G) and right is Source code (B).
But from the cases we should not assume that they are interchangeable, ie `m_expr a b` is not necessarily the same as `m_expr b a`.

We need 2 separate cases or maybe (if possible) rename the variables on the rhs of `| ...` to have `a1` pattern and `b1` code in both cases.

src/matching/Generic_vs_generic.ml
```diff
@@ -3089,6 +3089,20 @@ and m_definition_kind a b =
   (* boilerplate *)
   | G.EnumEntryDef a1, B.EnumEntryDef b1 -> m_enum_entry_definition a1 b1
   | G.FuncDef a1, B.FuncDef b1 -> m_function_definition a1 b1
+  (* iso: FuncDef pattern can match VarDef with Lambda (arrow function) *)
+  | G.FuncDef a1, B.VarDef { B.vinit = Some { e = B.Lambda b1; _ }; _ }
+  | B.VarDef { B.vinit = Some { e = B.Lambda b1; _ }; _ }, G.FuncDef a1  ->
```
dimitris-m: Same comment as below.

src/matching/Generic_vs_generic.ml
corneliuhoffman: indeed, in reading this I realised that I quite dislike the idea that someone looking for a lambda will find a functuon no?

src/matching/Generic_vs_generic.ml
corneliuhoffman: resolved

src/matching/Generic_vs_generic.ml
```diff
@@ -3089,6 +3089,20 @@ and m_definition_kind a b =
...
+        (fun x -> x.arrow_is_function)
+        ~then_:(m_function_definition a1 b1)
+        ~else_:(fail ())
+  (* iso: FuncDef pattern can match FieldDefColon with Lambda (arrow function in object literal) *)
+  | G.FuncDef a1, B.FieldDefColon { B.vinit = Some { e = B.Lambda b1; _ }; _ }
+  | B.FieldDefColon { B.vinit = Some { e = B.Lambda a1; _ }; _ }, G.FuncDef b1 ->  
+      if_config
+        (fun x -> x.arrow_is_function)
+        ~then_:(m_function_definition a1 b1)
+        ~else_:(fail ())
```
dimitris-m, line 3105: I am starting to doubt we need any of that.

What is the reason to have these identifications of different constructs? Won't it return weird results in search mode? Won't it enlarge the set of taint results in perhaps unintentional ways (if one has pattern-inside for example)?

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -193,6 +193,66 @@ let args_of_do_block_opt (blopt : do_block_generic option) : G.argument list =
...
+    object
+      inherit [_] AST_generic.iter_no_id_info as super
+
+      method! visit_expr env expr =
+        (match expr.G.e with
+        | G.OtherExpr
+            (("PlaceHolder", _), [ G.E { e = G.L (G.Int (Some n, _)); _ } ]) ->
+            max_found := max !max_found (Int64.to_int n)
+        | _ -> ());
+        super#visit_expr env expr
```
dimitris-m: I think we can do:

```ocaml
| _ -> super#visit_expr env expr);
```

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -193,6 +193,66 @@ let args_of_do_block_opt (blopt : do_block_generic option) : G.argument list =
...
+    end
+  in
+  mapper#visit_expr () e
+
+(** Convert a ShortLambda/Capture body to OtherExpr with params and body.
+    Creates params &1, &2, ... based on placeholders.
+    Structure: OtherExpr("ShortLambda", [Params [...]; S body_stmt])
+    This allows Naming_AST to create proper scope for the params. *)
+let convert_short_lambda (tok : Tok.t) (body_expr : G.expr) : G.expr =
+  let max_placeholder = find_max_placeholder body_expr in
```
dimitris-m, line 243: When you match a **pattern** such as `&(foo(... &3 ...))` how can you depend on the max placeholder? 
It will only match short lambdas with exactly 3 parameters.
How about `&(foo(...))`? No parameters at all, and no match in target code.

I think that if we are to be able to match using such patterns, we need to treat patterns differently. 

What I did in my current lang is to add an ellipsis after the max index (extra ellipsis parameter) only when parsing a pattern, so you can simply match any number of parameters with such patterns.

For the empty parameters case (no `&n` appears at all in the pattern), one should still add that ellipsis and it will match short lambdas of any parameter count. Better than no matches at all.

For Elixir the mechanics are there in `languages/elixir/generic/Elixir_to_generic.ml`: `type env = Program | Pattern`.

I recommend to add some tests about this material under `tests/patterns/elixir/`.

Also, tainting should work for a rule  with: 
```
- pattern-inside: |
       &($_(... $P ...))
- focus-metavariable: $P
```
and a sink like `sink(...)`, for example: 

```
&(sink &3)
```

should match I think.

src/tainting/Object_initialization.ml
```diff
@@ -371,6 +371,9 @@ let apex_constructor_pattern : constructor_pattern =
         | G.New (_, class_type, _, _) -> (
             match class_type.G.t with
             | G.TyN name when is_known_class name class_names -> Some name
+            | G.TyN name ->
+                (* For anonymous classes, accept the interface name even if not in class_names *)
+                Some name
```
dimitris-m: In that we could just remove then `when` from the case above.

perf/opengrep-scripts/bench.py
dimitris-m, line 1: @maciejpirog we have agreed to simplify this, leaving here as todo.

src/matching/Generic_vs_generic.ml
dimitris-m: Leaving a TODO for Part II: ensure this is what we want to do. 

Hopefully this only happens when we have already matched the entity and we should be comparing two `DefStmt definition`.

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -193,6 +193,65 @@ let args_of_do_block_opt (blopt : do_block_generic option) : G.argument list =
...
+      method! visit_expr env expr =
+        (match expr.G.e with
+        | G.OtherExpr
+            (("PlaceHolder", _), [ G.E { e = G.L (G.Int (Some n, _)); _ } ]) ->
+            max_found := max !max_found (Int64.to_int n)
+        | _ -> super#visit_expr env expr)
+    end
+  in
+  visitor#visit_expr () e;
+  !max_found
```
dimitris-m, line 217: TODO for Part II: 

- make class;
- place max_found in env so the instance can be reused.

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -193,6 +193,65 @@ let args_of_do_block_opt (blopt : do_block_generic option) : G.argument list =
...
+
+      method! visit_expr env expr =
+        match expr.G.e with
+        | G.OtherExpr
+            (("PlaceHolder", _), [ G.E { e = G.L (G.Int (Some n, tk)); _ } ]) ->
+            let param_name = Printf.sprintf "&%Ld" n in
+            let param_id = (param_name, tk) in
+            G.N (G.Id (param_id, G.empty_id_info ())) |> G.e
+        | _ -> super#visit_expr env expr
+    end
```
dimitris-m, line 234: TODO for Part II: make class and reuse instance.

languages/elixir/generic/Elixir_to_generic.ml
dimitris-m: TODO for Part II: changes as discussed offline.

src/analyzing/Visit_function_defs.ml
```diff
@@ -130,3 +131,124 @@ let fold_with_class_context
...
+  (visitor_parent_path, current_fn_id)
+
+class ['self] visitor_with_parent_path =
+  object (self : 'self)
+    inherit [_] G.iter_no_id_info as super
+    val current_class : G.name option ref = ref None
+    val parent_path : IL.name option list ref = ref []
+
+    method! visit_definition f ((ent, def_kind) as def) =
+      match def_kind with
```
dimitris-m, line 156: TODO for Part II: 
- make current_class and parent_path part of `env` (together with `f`)
- make stateless and reuse 1 instance of the class.

src/engine/Match_tainting_mode.ml
```diff
@@ -472,55 +487,142 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
+          (* Use shared call graph if provided, otherwise compute it *)
           let call_graph =
-            Function_call_graph.build_call_graph ~lang ~object_mappings ast
+            match shared_call_graph with
+            | Some (graph, _shared_mappings) -> graph
+            | None ->
+                (* Compute call graph as before *)
+                Function_call_graph.build_call_graph ~lang
+                  ~object_mappings:all_object_mappings ast
+          in
```
dimitris-m, line 504: Can this happen? Where is it stored after it's calculated?

dimitris-m: TODO for Part II: Check if this can really happen, at least debug log if it does.

src/tainting/Dataflow_tainting.ml
```diff
@@ -630,93 +637,264 @@ let effects_of_call_func_arg fun_exp fun_shape args_taints =
...
+      | Fetch { base = Var obj; rev_offset = [ { o = Dot method_name; _ } ] } -> (
+          (* Try object mapping first (for OOP languages) *)
+          match get_signature_for_object env.call_graph env.func.fn_id db method_name object_mappings obj arity with
+          | Some _ as result -> result
+          | None ->
+              (* Fallback: try qualified function name (Module.function for Elixir, etc.) *)
+              let qualified_name =
                 {
-                  ident = (class_str, snd method_name.ident);
+                  ident = (fst obj.ident ^ "." ^ fst method_name.ident, snd method_name.ident);
```
dimitris-m, line 880: TODO for Part II: Can we use `IdQualified` here? We might have to extend things a bit.

src/tainting/Dataflow_tainting.ml
```diff
@@ -1591,15 +1897,206 @@ let call_with_intrafile lval_opt e env args instr =
...
+              (match Lval_env.find_lval env.lval_env lval with
+              | Some (S.Cell (_, shape)) ->
+                  (match shape with
+                  | S.Fun _fun_sig ->
+                      (* It's a function/lambda! *)
+                      Some (e, lambda_exp)
+                  | _ -> None)
+              | None -> None)
+          | _ -> None)
+      | _ -> None)
```
dimitris-m, line 1919: TODO for Part II: Simplify this with a more refined pattern (else None)

src/tainting/Dataflow_tainting.ml
```diff
@@ -2270,14 +2871,14 @@ and fixpoint_aux taint_inst func ?(needed_vars = IL.NameSet.empty)
       ||| Limits_semgrep.taint_FIXPOINT_TIMEOUT)
   in
   let timeout =
-    if taint_inst.options.taint_intrafile then base_timeout *. 10.0
+    if taint_inst.options.taint_intrafile then base_timeout *. 20.0
```
dimitris-m, line 2874: TODO for Part II: timeout is not used any more (see `Dataflow_core.ml`). Remove.

src/tainting/Dataflow_tainting.ml
```diff
@@ -2328,6 +2930,116 @@ and (fixpoint :
...
+          IL.NameMap.fold
+            (fun lambda_name lambda_cfg acc_db ->
+              try
+                   Log.debug (fun m ->
+                       m "Extracting signature for lambda %s"
+                         (IL.str_of_name lambda_name));
+                   let lambda_fn_id = Shape_and_sig.make_fn_id_no_class lambda_name in
+                   let params = Signature.of_IL_params lambda_cfg.params in
+                   (* Create assumptions for lambda parameters *)
+                   (* Create assumptions for lambda parameters using Fold_IL_params *)
```
dimitris-m, line 2947: TODO for Part II: Unify comments.

src/tainting/Function_call_graph.ml
dimitris-m, line 1: TODO: Check for Part II

src/tainting/Taint_signature_extractor.ml
```diff
@@ -161,7 +166,8 @@ let mk_param_assumptions ?taint_inst (params : IL.param list) : Taint_lval_env.t
                  | None -> Taint.Taint_set.empty
                in
                let taint_set = Taint.Taint_set.union (Taint.Taint_set.singleton generic_taint) source_taints in
-               let new_env = Taint_lval_env.add_lval il_lval taint_set env in
+               (* Give the parameter an Arg shape so it can be used in HOF *)
+               let new_env = add_param_to_env il_lval taint_set taint_arg env in
```
dimitris-m, line 170: TODO for Part II: 
See below case of IL.PatternParm.
More cases needed probably.

PR #483, opened 2025-12-08
Improve handling of patterns and AST_to_IL translation

dimitris-m:
Fixes an number of issues, and improves matching and tainting: 

- In switch / match statements containing a list pattern of the shape `[..., $P, ...]`  matching did not work as expected, specifically `...` was ignored and it would only match lists of length 3. The same applies for tuples.

- Patterns in switches and lambdas with destructuring (example: Elixir) did not introduce proper variables, so e.g. the Elixir lambda `fn x -> sink(x)` which is internally translated into a switch with cases did not report taint. This is now fixed in the IL translation. Almost all kinds of patterns are now translated and taint propagates through nested destructuring patterns. See https://github.com/opengrep/opengrep/pull/483/commits/085093650b8d4dc62fb929f691424ccccb21c088.

- Fixed a bug that caused stuff like Elixir lambdas and OCaml `function`s to not match at all for some rules. See https://github.com/opengrep/opengrep/pull/483/commits/e520986578f97cbeb4b5ece1c8c7c565031c3458.

- Improved matching so that e.g. we can create a search rule that matches the identity function or more generally where e.g. a parameter is returned. See https://github.com/opengrep/opengrep/pull/483/commits/cf44d63a9fdfb20fc577ecdf4ad0dd10097a0a40.

- Changed the translation of Elixir lambdas so it uses better supported features, resulting in more fine-grained control of taint in guard conditions (previous translation was failing to distinguish that the last lambda in our test is actually ok, now this is captured).

See new and adapted files under `tests/`

### DIFF
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -418,39 +418,6 @@ and pattern env pat =
        * this kind of pattern. But I wonder if it would be pointless
        * to bother with it. *)
       (env, lval, pat_stmts @ guard_stmts)
-  (* XXX: This is for Elixir lambdas, and it's incomplete...
-   * TODO: Deal with other kinds of Arg and when expression which is a singleton
-   * rest in G.Args args :: rest.
-   * TODO: This interpretation is very specific to Elixir, not ideal.
-   * Maybe change the language encoding? We have G.PatWhen in AST_generic. *)
-  | G.OtherPat (("ArgsAndWhenOpt", tok), G.Args args :: maybe_when_expr)
-    when env.lang =*= Lang.Elixir ->
-    let expr_args =
-      List.filter_map (function G.Arg arg -> Some arg | _ -> None) args
-    in
-    let pats =
-      (* XXX: Now Related will not match a real part of the AST of the source,
-       * but tokens should be ok since they are preserved in translation? *)
-      List_.map AST_generic_helpers.expr_to_pattern expr_args
-    in
-    let env, lval, pat_stmts =
-      match pats with
-        | [] ->
-          let lval = fresh_lval (Tok.unsafe_fake_tok "_nopats") in
-          (env, lval, [])
-        | [pat] -> pattern env pat
-        | _ ->
-          pattern env (G.PatTuple (G.fake "(", pats, tok)) in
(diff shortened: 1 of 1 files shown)


libs/ast_generic/AST_generic.ml
```diff
@@ -2368,8 +2368,12 @@ let param_of_type ?(pattrs = []) ?pdefault ?pname typ =
...
   }
 
-(* for 'function 0 -> 1 ...' in OCaml or 'do 0 -> 1 ...' in Elixir *)
-let implicit_param_id tk = ("!_implicit_param!", tk)
+(* For 'function 0 -> 1 ...' in OCaml or 'do 0 -> 1 ...' in Elixir.
+ * Coupling: src/optimizing/Analyze_rule.ml where such idents are now
+ * ignored, since they are not expected to be found in the source file. *)
+let implicit_param = "!!_implicit_param!"
+let implicit_param_id tk = (implicit_param, tk)
+let is_implicit_param s = String.starts_with ~prefix:"!!_implicit_param!" s
```
maciejpirog, line 2377: Why can't we use an ID flag `hidden` for this? The comment in https://github.com/opengrep/opengrep/blob/main/libs/ast_generic/IdFlags.mli#L10 suggests this is exactly its purpose. The way it is done now, we have two mechanisms for the same thing, or am I confusing the use-cases for the `hidden` flag?

src/analyzing/AST_to_IL.ml
```diff
@@ -418,39 +418,6 @@ and pattern env pat =
        * this kind of pattern. But I wonder if it would be pointless
        * to bother with it. *)
       (env, lval, pat_stmts @ guard_stmts)
-  (* XXX: This is for Elixir lambdas, and it's incomplete...
```
maciejpirog: 🤣

libs/ast_generic/AST_generic.ml
dimitris-m: I see the TODO in Pattern.ml... maybe we can do this on a next PR?

src/analyzing/AST_to_IL.ml
dimitris-m: This was my previous attempt :)

PR #485, opened 2025-12-09
Improve rust tainting

dimitris-m:
## We now correctly propagate taint when using `if let`

```rust
if let P = E {...} else {...}
```

For example this target:

``` rust
fn f(some_x: Option<String>) {
    // ok
    let z = sink(x) 
    if let Some(x) = some_x {
        // ruleid: taint
        sink(x)
    } else {
        // ok
        sink(x)
        // should not hit since x is not defined here!
    }
}
```

will test as expected with this rule: 

``` 
rules:
- id: taint
  mode: taint
  languages: [rust]
  message: "tainted"
  severity: INFO
  pattern-sources:
    - patterns:
        - pattern-inside: |
            fn $F(..., $P: $_, ...){...}
        - focus-metavariable: $P
  pattern-sinks:
    - pattern: sink($R)
```

## Other fixes

- The tests in `tainting_tests/` were not running.
- The macos-13 image is now deprecated in GHA and CI failed on main, this is now fixed by bumping to macos-14.

### DIFF
--- a/src/osemgrep/cli_test/Test_subcommand.ml
+++ b/src/osemgrep/cli_test/Test_subcommand.ml
@@ -202,7 +202,7 @@ let rules_and_targets (kind : Test_CLI.target_kind) (errors : error list ref) :
            *)
           failwith
             "TODO: the split of tests/ and rules/ is not supported yet in \
-             semgrep test"
+             opengrep test"
       | File _
       | URL _
       | R _


src/engine/tests/Test_engine.ml
```diff
@@ -223,7 +223,7 @@ let make_test_rule_file ?(fail_callback = fun _i m -> Alcotest.fail m)
     | None -> ()
     | Some (rules, target, xlang) -> (
         (* expected *)
-        (* not tororuleid! not ok:! not todook:
+        (* not todoruleid! not ok:! not todook:
```
corneliuhoffman, line 226: :)))

src/analyzing/AST_to_IL.ml
```diff
@@ -1873,6 +1873,28 @@ and stmt_aux env st =
...
       let env, list_of_lists = List.fold_left_map stmt env xs in
       (env, List.concat list_of_lists)
+  (* Rust: if let Some(x) = some_x { ... } etc. *)
+  (* TODO: Handle LetChain too, see Parse_rust_tree_sitter. *)
+  | G.If (tok, G.OtherCond (("LetCond", _tk), [G.P pat; G.E e]), st1, st2)
+    when env.lang =*= Lang.Rust  ->
+    (* Convert to switch(e) { pat -> if_branch, _ -> else_branch }. *)
+    let cond_opt = Some (G.Cond e) in
+    let if_case_and_body =
+      G.CasesAndBody ([ G.Case (G.fake "some", pat) ], st1)
```
corneliuhoffman: seems to me this is specially for a Some pattern no? what if it is a different pattern?

corneliuhoffman: In fact I just realised that these are fine bc of Naming ... I only object to the "some" here :)

dimitris-m: I just call it "case" now, it does not matter what name it has, and there is a function in Generic AST that uses that name.

Issue #492, opened 2025-12-16
Release tag for v1.13.1 missing v prefix


lae:
`v` seems to have been accidentally dropped from the release tag for 1.13.1.

i.e. https://github.com/opengrep/opengrep/releases/tag/1.13.1 was created, but doc updates reference https://github.com/opengrep/opengrep/releases/tag/v1.13.1 and my attempt to update opengrep somewhere by just bumping the version number ran into a 404 as well.

PR #493, opened 2025-12-16
Validate tag input on release

lae:
This adds a step to the release workflow to help prevent releases being created without the `v` prefix (or failing semver in general), like in #492.

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -12,22 +12,37 @@ permissions:
   id-token: write  # for Cosign keyless
 
 jobs:
+  validate-inputs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate tag input
+        run: |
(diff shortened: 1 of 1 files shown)


dimitris-m:
> This adds a step to the release workflow to help prevent releases being created without the `v` prefix (or failing semver in general), like in #492.
> 

I'm afraid we can't always force that, we frequently run this to test things and in such cases we intentionally put a dummy tag like DO-NOT-RELEASE.

dimitris-m:
> Ah okay, in that case would it still be valid to just check for `TEST-\s+` as well? (I popped into the Actions history and I guess I see a few that are `TEST-DO-NOT-RELEASE`)

How about something much more liberal:
- it does not start with a digit
- if it starts with 'v', then it respects the format.

This way we don't have any accidental omission of the 'v' without disruption to our workflows.

It's not worth to put very strict conditions especially for something that is fixable easily if it happens (and only happened once).

lae:
That works. Updated the commit.

dimitris-m:
I did it again yesterday it seems, only realised now!

PR #498, opened 2025-12-19
Add dump-il to the show command

maciejpirog:
Adjusted from core. For example:

```
$ cat ../test/test.py
def foo():
 while True:
  x = 3

def bar():
  x = 2

$ opengrep show dump-il ../test/test.py
=== Toplevel ===
{ s =
  (MiscStmt
     (DefStmt
        ({ name =
           (EN
              (Id (("foo", ()),
                 { id_resolved = ref (None);
                   id_resolved_alternatives = ref ([]); id_type = ref (None);
                   id_svalue = ref (None); id_flags = ref (0) }
                 )));
           attrs = []; tparams = None },
         (FuncDef
            { fkind = (Function, ()); fparams = ((), [], ());
              frettype = None;
              fbody =
              (FBStmt
                 { s =
                   (Block
                      ((),
                       [{ s =
                          (While ((),
                             (Cond
                                { e = (L (Bool (true, ()))); e_id = 0;
                                  e_range = None; is_implicit_return = false;
                                  facts = <opaque> }),
                             { s =
                   ...
```

### DIFF
--- a/src/osemgrep/cli_show/Show_subcommand.ml
+++ b/src/osemgrep/cli_show/Show_subcommand.ml
@@ -1,5 +1,6 @@
 open Common
 module J = JSON
+module G = AST_generic
 
 (*****************************************************************************)
 (* Prelude *)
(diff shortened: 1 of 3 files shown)


src/osemgrep/cli_show/Show_subcommand.ml
```diff
@@ -135,6 +136,43 @@ let run_conf (caps : < caps ; .. >) (conf : Show_CLI.conf) : Exit_code.t =
...
+            let _ = CFG_build.cfg_of_gfdef lang fdef in
+
+            (* This round, the IL stmts will show return nodes when
+            * they were implicit before.
+            *)
+            let IL.{ fbody = xs; _ } = AST_to_IL.function_definition lang fdef in
+            let s = IL.show_any (IL.Ss xs) in
+            print s
+        in
+        Visit_function_defs.visit report_func_def_with_name ast;
```
dimitris-m, line 174: But now we have code duplication in src/core_cli/Core_actions.ml and here.

Anyway not important, seems to be the case for other things too in this file.

PR #508, opened 2025-12-25
Dart: fixes

maciejpirog:
Dart things fixed in this PR:
- file extensions were not properly recognised
- in some cases, method bodies were dropped in the to-ast-generic step
- incorrect AST for interpolated strings

### DIFF
--- a/src/engine/tests/Unit_engine.ml
+++ b/src/engine/tests/Unit_engine.ml
@@ -619,6 +619,7 @@ let lang_tainting_tests () =
       (* lang, dir, ext *)
       (Lang.Apex, "apex", ".trigger"); (* TODO: Use Lang.* functions to derive .2, .3 *)
       (Lang.Csharp, "csharp", ".cs");
+      (Lang.Dart, "dart", ".dart");
       (Lang.Elixir, "elixir", ".ex");
       (Lang.Go, "go", ".go");
       (Lang.Java, "java", ".java");


src/engine/tests/Unit_engine.ml
```diff
@@ -619,6 +619,7 @@ let lang_tainting_tests () =
       (* lang, dir, ext *)
       (Lang.Apex, "apex", ".trigger"); (* TODO: Use Lang.* functions to derive .2, .3 *)
       (Lang.Csharp, "csharp", ".cs");
+      (Lang.Dart, "dart", ".dart");
```
dimitris-m, line 622: Added this, but really we should auto-discover.
I also forgot to do this in my clojure PR.

languages/dart/generic/Parse_dart_tree_sitter.ml
```diff
@@ -2297,23 +2302,21 @@ and map_switch_label (env : env) ((v1, v2) : CST.switch_label) : case =
...
-        in
-        Left3 (s2, t2)
-  in
-  v2
+      in
+      (* TODO: this is copy-paste from map_sub_string_test, *)
+      (* something weird is going on in the grammar *)
+      match env.extra with
+      | Program -> Either_.Middle3 (N (H2.name_of_id (s2, t2)) |> G.e)
+      | Pattern -> Either_.Left3 (s1 ^ s2, Tok.combine_toks t1 [ t2 ])
```
dimitris-m, line 2319: nit: this is pervasive but I personally find the use o Either bad practice in such contexts. Better to introduce new types so it's clear what happens.

dimitris-m (approved):

src/engine/tests/Unit_engine.ml
maciejpirog: Thanks! Indeed, easy to forget about these

languages/dart/generic/Parse_dart_tree_sitter.ml
maciejpirog: I definitely agree. Shouldn't be difficult to refactor (added to my TODO list)

PR #513, opened 2025-12-30
Fixed top level HOFS and inlined lambda

corneliuhoffman:
Two issues with HOF taint analysis were not working:

  1. Inline lambda calls: Calling a lambda immediately after definition wasn't tracked
 ```python
  (lambda x: sink(x))(source())  # Not detected
  ```
  2. Top-level HOF callbacks: Lambdas/functions defined at module level and passed to HOFs weren't being analyzed
  ```ruby
  toplevel_sink = ->(x) { sink(x) }
  toplevel_items.each(&toplevel_sink)  # Not detected
```
  Solution

  Inline lambda calls (Dataflow_tainting.ml)
  - Detect lambda expressions in callee position and analyze them inline with tainted arguments

  Top-level HOF callbacks (Function_call_graph.ml, Visit_function_defs.ml)
  - Add visit_with_parent_path to track parent function context during AST traversal
  - Extract HOF callbacks from top-level code (outside function bodies)
  - Add edges from top-level lambdas to <top_level> node in call graph

  Ruby/PHP block parameters (AST_to_IL.ml)
  - Handle &callback syntax (OtherParam("Ref", ...)) as proper parameters for taint tracking

  Builtin models (Builtin_models.ml)
  - Add MethodHOF for Ruby to handle .each(&callback) pattern

  Code cleanup

  - Simplified constructor implicit edge logic (removed unnecessary exclusion)
  - Removed unsafe fallback matching ANY class method by name
  - Commented out debug dot file generation

  Test plan

  - All HOF comprehensive tests pass (Ruby, Scala, Python, JS, etc.)
  - New test: tests/tainting/lambda_inline_call.py
  - New test: tests/tainting/toplevel_hof_callback.py

### DIFF
--- a/src/analyzing/Visit_function_defs.ml
+++ b/src/analyzing/Visit_function_defs.ml
@@ -164,6 +164,22 @@ let fold_with_class_context
    - Nested function: [None; Some parent_func; Some nested_func] (excluding current function)
 *)
 
+(* Convert G.name to IL.name for fn_id path construction.
+   Uses unsafe_default sid and clears id_resolved to ensure consistent
+   comparison in FunctionMap (which compares by string name + position). *)
+let g_name_to_il_name (g_name : G.name) : IL.name option =
+  match g_name with
+  | G.Id ((str, tok), id_info) ->
+      let id_info = { id_info with G.id_resolved = ref None } in
+      Some IL.{ ident = (str, tok); sid = G.SId.unsafe_default; id_info }
+  | _ -> None
+
+(* Convert G.entity to IL.name for fn_id path construction. *)
+let entity_to_il_name (ent : G.entity) : IL.name option =
+  match ent.G.name with
+  | G.EN name -> g_name_to_il_name name
+  | _ -> None
+
 let append_to_parrent_path parent_path class_il func_il =
   let visitor_parent_path =
     if parent_path = [] then [ class_il ] else parent_path
@@ -189,8 +205,8 @@ class ['self] visitor_with_parent_path =
               super#visit_definition f def)
       | G.FuncDef fdef ->
           (* Build fn_id path: [class_option; ...parent_path...; current_func] *)
-          let class_il = Option.map AST_to_IL.var_of_name !current_class in
-          let func_il = AST_to_IL.name_of_entity ent in
+          let class_il = Option.bind !current_class g_name_to_il_name in
+          let func_il = entity_to_il_name ent in
 
           (* Call the visitor function with parent path (without current function) *)
           let visitor_parent_path, current_fn_id =
@@ -204,8 +220,8 @@ class ['self] visitor_with_parent_path =
               super#visit_stmt f body)
       | G.VarDef { vinit = Some { e = G.Lambda fdef; _ }; _ } ->
           (* Handle lambda assignments like: const f = () => {...} *)
-          let class_il = Option.map AST_to_IL.var_of_name !current_class in
-          let func_il = AST_to_IL.name_of_entity ent in
+          let class_il = Option.bind !current_class g_name_to_il_name in
+          let func_il = entity_to_il_name ent in
           let visitor_parent_path, current_fn_id =
             append_to_parrent_path !parent_path class_il func_il
           in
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/analyzing/Visit_function_defs.ml [hunk 1]
```diff
@@ -164,8 +205,8 @@ class ['self] visitor_with_parent_path =
               super#visit_definition f def)
       | G.FuncDef fdef ->
           (* Build fn_id path: [class_option; ...parent_path...; current_func] *)
-          let class_il = Option.map AST_to_IL.var_of_name !current_class in
-          let func_il = AST_to_IL.name_of_entity ent in
+          let class_il = Option.bind !current_class g_name_to_il_name in
+          let func_il = entity_to_il_name ent in
```
dimitris-m, line 209: But `AST_to_IL.var_of_name` also deals with `IdQualified`.

Is there no reason to deal with that in `g_name_to_il_name`?

src/analyzing/Visit_function_defs.ml
```diff
@@ -252,3 +307,30 @@ let fold_with_parent_path
...
+      (match e.G.e with
+      | G.Call (callee, args) -> f e callee args
+      | _ -> ());
+      super#visit_expr f e
+  end
+
+(* Visit all Call expressions at top level (outside function bodies).
+   Callback receives: full call expr, callee expr, and arguments *)
+let visit_toplevel_calls (f : G.expr -> G.expr -> G.arguments -> unit) (ast : G.program) : unit =
+  let v = new toplevel_call_visitor in
```
dimitris-m, line 327: Is there a reason to create a new instance on each call?

Why not share the visitor (pull out of function)?

src/analyzing/Visit_function_defs.ml
```diff
@@ -252,3 +307,30 @@ let fold_with_parent_path
...
+  v#visit_program f ast
+
+(* Fold over all Call expressions at top level *)
+let fold_toplevel_calls (f : 'acc -> G.expr -> G.expr -> G.arguments -> 'acc)
+    (init_acc : 'acc) (ast : G.program) : 'acc =
+  let acc_ref = ref init_acc in
+  let v = new toplevel_call_visitor in
```
dimitris-m, line 334: As above, visitor instance can be shared.

src/analyzing/Visit_function_defs.ml
corneliuhoffman: oh ... that answers my question in your PR :)))
My point is that I wanted the function to be kindda pure ... despirte the visitor mess inside. And that way I don't have to worry too much about threards

src/analyzing/Visit_function_defs.ml
corneliuhoffman: I ampretty sure not (we only use it for classes and methods)

src/analyzing/Visit_function_defs.ml
corneliuhoffman: as above, can change but did not want to think about threads

tests/rules/cross_function_tainting/test_hof_comprehensive_c.c
```diff
@@ -65,9 +65,26 @@ char* source() {
...
     test_custom_foreach();
     test_direct_call();
     test_original_example();
+
+    // Top-level user-defined HOF with named callback
+    char* toplevel_items[] = {source()};
+    customForEach(toplevel_items, 1, &toplevel_handler);
+
+    // Top-level direct call with function pointer
+    directCall(&toplevel_handler);
```
dimitris-m, line 87: Don't you need some `ruleid: test-hof-taint` here?

src/tainting/Function_call_graph.ml
```diff
@@ -577,160 +579,110 @@ let identify_callback ?(all_funcs = []) ?(caller_parent_path = [])
...
-        | _ -> super#visit_expr env e
+        (match e.G.e with
+        | G.Call (callee, args) ->
+            let found = extract_hof_callbacks_from_call
+              ~method_hofs ~function_hofs ~user_hofs ~all_funcs ~caller_parent_path
+              callee args
+            in
+            callbacks := found @ !callbacks
+        | _ -> ());
+        super#visit_expr env e
```
dimitris-m, line 679: Leaving as TODO to extract this into a class, there are some potential concurrency issues with objects + domains.
It's one of the reasons I converted most (all?) such objects into classes and had the instances stateless & shared.

PR #514, opened 2025-12-30
C#: Better handing of extension methods and extension blocks

maciejpirog:
## Extension methods

In C# some function parameters are not real parameters, in particular `this` parameters in extension methods. Such a method can be defined as:

```C#
void foo(this string s, int a) { }
```

It is, in fact, a single-argument method on the `string` class, which can later be called as `val.foo(123)`. To better handle definitions of such extension methods and `extension` blocks, we:

- make it match function definition with arguments without the artificial `this` arguments, e.g. the definition above will be matched by patterns:
   * `$T $F($PT $PN, int a) { ... }`
   * `$T $F(this $PT $PN, int a) { ... }`
   * `$T $F(int a) { ... }`

## Extension blocks

There is a syntactic sugar for extension methods: extension blocks. The code

```C#
public static class C
{
  extension(string s)
  {
    public string InQuotes()
    {
      return "\"" + s + "\"";
    }
  }
}
```

is equivalent to

```C#
public static class C
{
  public string InQuotes(this string s)
  {
    return "\"" + s + "\"";
  }
}
```

There is no good way to express such a block in Generic AST and we want to be able to do taint analysis with the `string s` above treated as source. Hence, in this PR, we translate extension blocks to a bunch of extension methods. This way, we can write a rule such as:

```yaml
rules:
  - id: extension_taint
    message: "extension taint test"
    languages:
      - csharp
    severity: ERROR
    mode: taint
    pattern-sources:
      - patterns:
        - pattern-inside: |
            $RET $FUNC(this string $PARAM, ...) { ... }
        - focus-metavariable: $PARAM
    pattern-sinks:
      - patterns:
          - pattern: |
              sink(s)
```

And it will match the following:

```C#
public static class C
{
    // in an extension block:
    
    // in the rule, we assume that the extended string is tainted
    extension(string s)
    {
        void foo1(int aaaa){
            // ruleid: extension_taint
            sink(s);
        }
        
        void foo2(){
            // ok:
            sink(t);
        }
        
        void foo3(){
            // ruleid: extension_taint
            sink(s);
        }
        
        // ruleid: extension_taint
        public string Prop2 => sink(s);

        public string Prop
        {
            // ruleid: extension_taint
            get => sink(s);
            // ruleid: extension_taint
            set => sink(s);
        }
    }

    // standalone:

    public void foo4(this string s)
    {
      // ruleid: extension_taint
      sink(s);
    }
    
}
```

## Future work

- At the moment we cannot match `extensions` block as separate units, we have to write `class $C { ... extension ($T $V) { ... } ... }` which might be imprecise for more complicated matching rules.

- For extension blocks, if we want to say the entire object is tainted, we show the source of taint in the dataflow trace to be the method itself, not the `string s` in the line that contains `extension(string s)`.


languages/csharp/generic/Parse_csharp_tree_sitter.ml
```diff
@@ -3075,7 +3076,27 @@ and record_struct_declaration env (_, _, v3, _, _, _, _, _, _, _, _) =
...
+          (* don't insert the param: it would be shadowed anyway *)
+          s
+      | _ ->
+          let params = tp :: params in
+          let fdef = { fdef with fparams = (l, params, r) } in
+          let res = G.DefStmt (ent, G.FuncDef fdef) |> G.s in
+          let orig_range = H2.range_of_any_opt (G.S s) in
+          (match orig_range with
+          | Some _ -> res.s_range <- orig_range; res
+          | None -> res))
```
dimitris-m, line 3096: Here you could have just done `{s with s = G.DefStmt (ent, G.FuncDef fdef)}` instead of creating new statement -- would that make the range stuff redundant?

dimitris-m: Maybe not, now you get extended range is that it? Because of injected param.

languages/csharp/generic/Parse_csharp_tree_sitter.ml
```diff
@@ -3395,13 +3418,15 @@ and declaration (env : env) (x : CST.declaration) : stmt =
                   let has_params = iname <> "get" in
                   let has_return = iname = "get" in
                   let ent = basic_entity (iname ^ "_" ^ fname, itok) ~attrs in
+                  let this_param_list = (* Option.to_list (Option.map (fun f -> f ()) this_param) *) [] in
```
dimitris-m: Is this supposed to always be [] ?

languages/csharp/generic/Parse_csharp_tree_sitter.ml
maciejpirog: in this case, `{s with s = G.DefStmt (ent, G.FuncDef fdef)}` and `G.DefStmt (ent, G.FuncDef fdef) |> G.s` do the same thing, as the range in the argument `s` is not set anyway (it is a freshly created statement coming from the `declaration` function).

This means I create a stmt (in `declaration`) only to create a different statement (in `add_this_param`), but it is intentional: the former statement is used to calculate its range, which is used as the range of the latter statement.

languages/csharp/generic/Parse_csharp_tree_sitter.ml
maciejpirog: Nope, forgot to revert some experiments. Fixed now.

PR #516, opened 2026-01-01
C#: Allow implicit variables in properties to be taint sources

maciejpirog:
The implicit variables `field` and `value` in property definitions can be considered a source of taint. In this PR we introduce a special syntax to write patterns that do that. It is possible by treating `get` and `set` as functions with `field` and `value` as arguments. The syntax is enabled by giving the `[get]` or `[set]` attribute to the function definition in the pattern:

For example, the rules:

```yaml
rules:
- id: taint_prop_field
  languages: [csharp]
  message: "a function is called"
  severity: WARNING
  mode: taint
  pattern-sources:
    - patterns:
      - pattern-inside: |
          [get] $T $F($FIELD) { ... }
      - focus-metavariable: $FIELD
    - patterns:
      - pattern-inside: |
          [set] $T $F($VALUE, $FIELD) { ... }
      - focus-metavariable: $FIELD
  pattern-sinks:
    - pattern: |
        sink(...)
- id: taint_prop_value
  languages: [csharp]
  message: "a function is called"
  severity: WARNING
  mode: taint
  pattern-sources:
    - patterns:
      - pattern-inside: |
          [set] $T $F($VALUE, $FIELD) { ... }
      - focus-metavariable: $VALUE
  pattern-sinks:
    - pattern: |
        sink(...)
```

can be used to match

```C#
class C
{
    public string Message
    {
        // ruleid: taint_prop_field
        get => sink(field);
        //ruleid: taint_prop_field
        set => sink(field);
    }
    public string Message2
    {
        // ruleid: taint_prop_value
        set => sink(value);
    }
}
```

## TODO

- Sources have to have real locations. The location of a parameter cannot overlap with the location of another parameter, because otherwise, when one is a source of taint, so is the other. The solution for now is to anchor the source to particular characters in the `get` and `set` keywords. This makes `--dataflow-traces` highlighting a bit weird, but a deeper refactoring is needed to solve this. :(

### DIFF
--- a/src/reporting/Core_json_output.ml
+++ b/src/reporting/Core_json_output.ml
@@ -285,7 +285,7 @@ let token_to_intermediate_var token : Out.match_intermediate_var option =
    * a specific job for us.
    * TODO: At least make the condition str =*= AST_generic.fake_ident and then
    * other places can use it. *)
-  | Tok.OriginTok {str;_ } when str =*= "G__1111" -> None
+  | Tok.OriginTok {str;_ } when str =*= "G__1111" || str =*= "" -> None
   | _ ->
     let* location = OutUtils.tokens_to_single_loc [ token ] in
     Some


maciejpirog:
> Something along these lines:

Nope, this won't work because you will match any instance of `value` or `field`, even after you sanitize it, or ones that are shadowing variables, e.g.

```
get => value => sink(value)
```

which introduces a fresh var called `value`, which is inside the pattern

src/osemgrep/cli_show/Show_CLI.ml
```diff
@@ -47,7 +47,9 @@ and show_kind =
   | DumpAST of Fpath.t * Lang.t
   | DumpIL  of Fpath.t * Lang.t
   | DumpConfig of Rules_config.config_string
+  | DumpRule of Fpath.t
   | DumpRuleV2 of Fpath.t
+  | DumpPatternsOfRule of Fpath.t
```
dimitris-m, line 52: Is that not from #519 ?

maciejpirog: Must have cherry-picked on a wrong branch somehow. Fixed

PR #529, opened 2026-01-08
Php modernisation

### LINKED ISSUE #297: PHP 5->Php 8.4 parser

corneliuhoffman:
PHP 7.1 - 8.4 Syntax Support

  Adds comprehensive PHP syntax support to the Menhir parser. Closes #297.

  ---
  b0529bca0f - added tests

  Initial test files for PHP 8.x features.

  ---
  d3da6612c4 - readonly fix

  PHP 8.2: Readonly Classes
```php
  readonly class ImmutablePoint {
      public int $x;
      public int $y;
  }
```
  ---
  18aa6ce570 - Union and Intersection DNF

  PHP 8.2: DNF Types (Disjunctive Normal Form)
```php
  function process((Countable&Iterator)|null $input): (A&B)|C {
      // ...
  }
```
  ---
  b59447d2ff - property hooks

  PHP 8.4: Property Hooks
```php
  class User {
      // Block syntax
      public string $name {
          get { return $this->name; }
          set { $this->name = strtoupper($value); }
      }

      // Arrow syntax
      public int $age {
          get => $this->age;
          set => max(0, $value);
      }
  }
```
  ---
  aaed1fb1ad - asymmetric visibility private(set) and protected(set)

  PHP 8.4: Asymmetric Visibility
```php
  class Counter {
      public private(set) int $count = 0;
      public protected(set) string $name = "";
  }
```
  ---
  19e3d19d50 - few others

  PHP 7.1: Class Constant Visibility
```php
  class Config {
      public const PUBLIC_CONST = 1;
      protected const PROTECTED_CONST = 2;
      private const PRIVATE_CONST = 3;
  }
```
  PHP 7.1: Multi-Catch Exception Handling
```php
  try {
      riskyOperation();
  } catch (InvalidArgumentException | TypeError | ValueError $e) {
      handleError($e);
  }
```
  PHP 8.3: Dynamic Class Constant Fetch
```php
  $name = 'VERSION';
  echo Config::{$name};
```
  ---
e09be5428b27 - first-class callable syntax

  PHP 8.1: First-Class Callable Syntax
```php
  $fn = strlen(...);           // → Closure::fromCallable('strlen')
  $fn = $obj->method(...);     // → Closure::fromCallable([$obj, 'method'])
  $fn = Foo::bar(...);         // → Closure::fromCallable([Foo::class, 'bar'])
```  In semgrep patterns, ... retains its ellipsis meaning.

### DIFF
--- a/languages/php/generic/php_to_generic.ml
+++ b/languages/php/generic/php_to_generic.ml
@@ -366,13 +366,12 @@ and expr e : G.expr =
         match callable_expr with
         (* strlen(...) → Closure::fromCallable('strlen') *)
         | Id [ name ] ->
-            let str = fst name in
-            G.Arg (G.L (G.String (fb (str, snd name))) |> G.e)
+            G.Arg (G.L (G.String (fb name)) |> G.e)
         (* $obj->method(...) → Closure::fromCallable([$obj, 'method']) *)
         | Obj_get (obj, _arrow, Id [ method_name ]) ->
             let obj_expr = expr obj in
             let method_str =
-              G.L (G.String (fb (fst method_name, snd method_name))) |> G.e
+              G.L (G.String (fb method_name)) |> G.e
             in
             G.Arg
               (G.Container (G.Array, fb [ obj_expr; method_str ]) |> G.e)
@@ -380,7 +379,7 @@ and expr e : G.expr =
         | Class_get (class_ref, _colons, Id [ method_name ]) ->
             let class_expr = expr class_ref in
             let method_str =
-              G.L (G.String (fb (fst method_name, snd method_name))) |> G.e
+              G.L (G.String (fb  method_name)) |> G.e
             in
             G.Arg (G.Container (G.Array, fb [ class_expr; method_str ]) |> G.e)
         (* For other forms, pass as-is *)
--- a/tests/parsing/php/php8_readonly.php
+++ b/tests/parsing/php/php8_readonly.php
@@ -1,5 +1,5 @@
 <?php
-// PHP 8.1: Readonly properties - WORKS
+// PHP 8.1: Readonly properties 
 class ReadonlyProperty {
     public readonly int $id;
 
@@ -8,7 +8,7 @@ class ReadonlyProperty {
     }
 }
 
-// PHP 8.2: Readonly classes - WORKS
+// PHP 8.2: Readonly classes 
 readonly class ReadonlyClass {
     public int $value;
 }
--- a/tests/parsing/php/php8_static_return.php
+++ b/tests/parsing/php/php8_static_return.php
@@ -1,5 +1,5 @@
 <?php
-// PHP 8.0: Static return type - WORKS
+// PHP 8.0: Static return type 
 class StaticReturn {
     public function create(): static {
         return new static();

### CODE REVIEW

File: languages/php/ast/ast_php.ml [hunk 1]
```diff
@@ -269,7 +272,11 @@ and hint_type =
   | HintArray of tok
   | HintQuestion of tok * hint_type
   | HintTuple of hint_type list bracket
-  | HintUnion of hint_type list
+  (* first type, then list of (| tok, type) pairs *)
+  | HintUnion of hint_type * (tok * hint_type) list
+  (* PHP 8.1 intersection types, PHP 8.2 DNF types *)
+  (* first type, then list of (& tok, type) pairs, then parens *)
+  | HintIntersection of hint_type * (tok * hint_type) list * unit bracket
```
dimitris-m: What is `unit bracket` here?

dimitris-m: Maybe this is supposed to be: `(hint_type * (tok * hint_type) list * unit) bracket` ?

dimitris-m: And indeed why not have `HitUnion` also be bracketed? It can be fake brackets when missing.

corneliuhoffman: You are right (without the unit). I was just moving the bracket out of the way but there is no needs. changed

corneliuhoffman: I thought about it, the point is that it seems that HintIntersection has mandatory brackets ... will add for thwe others as well

corneliuhoffman: in fact on second thought this is DNF so the brackets on Union never happen https://wiki.php.net/rfc/dnf_types

languages/php/menhir/ast_php_build.ml
```diff
@@ -822,19 +856,40 @@ and class_variables env ?(add_dollar = true) st acc =
...
+  let kind = match fst hook.ph_kind with
+    | PhGet -> A.PhGet (snd hook.ph_kind)
+    | PhSet -> A.PhSet (snd hook.ph_kind)
+  in
+  let params = match hook.ph_params with
+    | None -> []
+    | Some (_, ps, _) -> List_.map (fun p -> A.ParamClassic (parameter env p)) (comma_list_dots ps)
+  in
+  let body = match hook.ph_body with
+    | PHExpr (_, e) -> Some (A.Expr (expr env e, Tok.unsafe_fake_tok ";"))
```
dimitris-m: Is the `;` really there? Else we could use `AST_generic.sc`.

tests/parsing/php/php81_first_class_callable.php
```diff
@@ -0,0 +1,42 @@
...
+// Closures and invokables
+$fn11 = $closure(...);
+$fn12 = $invokableObject(...);
+
+// Array callables
+$fn13 = [$obj, 'method'](...);
+$fn14 = [Foo::class, 'method'](...);
+
+// Complex expressions
+$fn15 = self::{$complex . $expression}(...);
```
dimitris-m, line 30: Is this also `Closure::fromCallable` ?

dimitris-m: Seems so.

tests/parsing/php/php8_property_hooks.php
```diff
@@ -0,0 +1,27 @@
...
+// PHP 8.4: Property hooks
+class PropertyHooks {
+    // Block syntax
+    public string $name {
+        get {
+            return $this->name;
+        }
+        set {
+            $this->name = strtoupper($value);
+        }
```
dimitris-m, line 27: I recommend to add a tainting test also for this material. Where `$value` is tainted for example.

dimitris-m: In particular I'm wondering how it relates to #516 and #525.

dimitris-m: cc @maciejpirog

tests/rules/php_first_class_callable.yaml
```diff
@@ -0,0 +1,6 @@
+rules:
+- id: first-class-callable
+  message: Detected first-class callable (desugared to Closure::fromCallable)
+  languages: [php]
+  severity: INFO
+  pattern: Closure::fromCallable(...)
```
dimitris-m, line 6: We should document that this is the only way to match such constructs.

dimitris-m: Maybe in Wiki?

tests/rules/php_ellipsis_args.php
```diff
@@ -0,0 +1,23 @@
+<?php
+// Test that ... in patterns matches any arguments
+// This should NOT be confused with PHP 8.1 first-class callables
```
dimitris-m, line 3: Maybe we can add an example of this, with `// ok` ?

tests/rules/php_ellipsis_args.php
```diff
@@ -0,0 +1,23 @@
...
+
+function test() {
+    # ruleid:ellipsis-args
```
dimitris-m: This may work but it's not valid PHP comment right? Should it not be `// ruleid: ellipsis-args` ?

Does it parse?

Same for other comments below.

dimitris-m: Oh, it seems valid... ok.

Missing space between : and rule name.

corneliuhoffman:
> Property hooks don't produce correct AST. Example from https://www.php.net/manual/en/language.oop5.property-hooks.php
  whoa, thanks

languages/php/menhir/ast_php_build.ml
corneliuhoffman: cleaned up to include that in the PHExpr

tests/rules/php_first_class_callable.yaml
corneliuhoffman: will add a php upgrade in wiki

tests/rules/php_ellipsis_args.php
corneliuhoffman: modified

tests/rules/php_ellipsis_args.php
corneliuhoffman: added

tests/parsing/php/php8_property_hooks.php
corneliuhoffman: added

languages/php/generic/php_to_generic.ml
```diff
@@ -356,6 +360,45 @@ and expr e : G.expr =
...
       let v1 = expr v1 and v2 = bracket (list argument) v2 in
       G.Call (v1, v2) |> G.e
+  (* PHP 8.1: first-class callable - desugar to Closure::fromCallable(...) *)
+  | FirstClassCallable (callable_expr, tok) ->
+      let callable_arg =
+        match callable_expr with
+        (* strlen(...) → Closure::fromCallable('strlen') *)
+        | Id [ name ] ->
+            let str = fst name in
+            G.Arg (G.L (G.String (fb (str, snd name))) |> G.e)
```
dimitris-m: So why not just: 

`G.Arg (G.L (G.String (fb name)) |> G.e)`

languages/php/generic/php_to_generic.ml
```diff
@@ -356,6 +360,45 @@ and expr e : G.expr =
...
+        (* $obj->method(...) → Closure::fromCallable([$obj, 'method']) *)
+        | Obj_get (obj, _arrow, Id [ method_name ]) ->
+            let obj_expr = expr obj in
+            let method_str =
+              G.L (G.String (fb (fst method_name, snd method_name))) |> G.e
```
dimitris-m: `G.L (G.String (fb method_name)) |> G.e` ?

languages/php/generic/php_to_generic.ml
```diff
@@ -356,6 +360,45 @@ and expr e : G.expr =
...
+            in
+            G.Arg
+              (G.Container (G.Array, fb [ obj_expr; method_str ]) |> G.e)
+        (* Foo::bar(...) → Closure::fromCallable([Foo::class, 'bar']) *)
+        | Class_get (class_ref, _colons, Id [ method_name ]) ->
+            let class_expr = expr class_ref in
+            let method_str =
+              G.L (G.String (fb (fst method_name, snd method_name))) |> G.e
```
dimitris-m: `G.L (G.String (fb method_name)) |> G.e` as above?

languages/php/generic/php_to_generic.ml
corneliuhoffman: :)) fixed

languages/php/generic/php_to_generic.ml
corneliuhoffman: ditto

languages/php/generic/php_to_generic.ml
corneliuhoffman: ditto

tests/parsing/php/php71_multi_catch.php
```diff
@@ -0,0 +1,21 @@
...
+
+class Exception1 extends Exception {}
+class Exception2 extends Exception {}
+class Exception3 extends Exception {}
+
+try {
+    throw new Exception1("error");
+} catch (Exception1 | Exception2 $e) {
+    echo "Caught exception 1 or 2: " . $e->getMessage();
+} catch (Exception3 | \RuntimeException $e) {
```
dimitris-m, line 12: Is `\RuntimeException` special (what is the slash)?

corneliuhoffman: I think "\" means "global". The others are in the local file.

tests/parsing/php/php8_readonly.php
```diff
@@ -0,0 +1,14 @@
+<?php
+// PHP 8.1: Readonly properties - WORKS
```
dimitris-m: Remove "- WORKS"

tests/parsing/php/php8_static_return.php
```diff
@@ -0,0 +1,7 @@
+<?php
+// PHP 8.0: Static return type - WORKS
```
dimitris-m: Remove "- WORKS"

tests/parsing/php/php8_readonly.php
```diff
@@ -0,0 +1,14 @@
...
+class ReadonlyProperty {
+    public readonly int $id;
+
+    public function __construct(int $id) {
+        $this->id = $id;
+    }
+}
+
+// PHP 8.2: Readonly classes - WORKS
```
dimitris-m: Remove "- WORKS"

dimitris-m (approved):
LGTM - modulo small comments

PR #530, opened 2026-01-09
Refactor Sig_inst.find_pos_in_actual_args

maciejpirog:
Refactor `Sig_inst.find_pos_in_actual_args` to make it shorter, more readable, and more performant (hopefully)

### DIFF
--- a/src/tainting/Sig_inst.ml
+++ b/src/tainting/Sig_inst.ml
@@ -349,86 +349,96 @@ let instantiate_shape inst_var inst_trace shape =
   in
   inst_shape shape
 
-let find_pos_in_actual_args ?(err_ctx = "???") args_taints fparams :
-    T.arg -> _ option =
+let find_pos_in_actual_args ?(err_ctx = "???") (args : 'a IL.argument list)
+    (fparams : Signature.params) : T.arg -> 'a option =
   Log.debug (fun m ->
       m "FIND_POS_IN_ACTUAL_ARGS: err_ctx=%s, num_args=%d, num_fparams=%d, fparams=%s"
         err_ctx
-        (List.length args_taints)
+        (List.length args)
         (List.length fparams)
         (fparams |> List.map (function Signature.P s -> "P(" ^ s ^ ")" | Signature.Other -> "Other") |> String.concat ", "));
-  let pos_args_taints, named_args_taints =
+  (* We go left-to-right through formal params. If a param is named and there
+   * is an actual named arg, use it; if not, take the first non-named actual
+   * arg available. NOTE that it is the Python semantics, and can potentially lead 
+   * to problems with languages like OCaml, where there is a clear distinction
+   * between named and non-named arguments. *)
+  let pos_args, named_args =
     List.partition_map
-      IL.(
-        function
-        | Unnamed taints -> Left taints
-        | Named (id, taints) -> Right (id, taints))
-      args_taints
+      (function
+      | IL.Unnamed v -> Left v
+      | IL.Named ((name, _token), v) -> Right (name, v))
+      args
   in
-  let named_arg_map =
-    named_args_taints
-    |> List.fold_left
-         (fun xmap ((s, _), taint) -> SMap.add s taint xmap)
-         SMap.empty
+  let name_vals =
+    fparams |> List.map (function
+       | Signature.P name -> Some (name, List.assoc_opt name named_args)
+       | _ -> None)
   in
-  let name_to_taints = Hashtbl.create 10 in
-  let idx_to_taints = Hashtbl.create 10 in
-  (* We first process the named arguments, and then positional arguments.
-   *)
-  let remaining_params =
-    (* Here, we take all the named arguments and remove them from the list of parameters.
-     *)
-    List_.fold_right
-      (fun (param : Signature.param) acc ->
-        match param with
-        | P s' -> (
-            match SMap.find_opt s' named_arg_map with
-            | Some taints ->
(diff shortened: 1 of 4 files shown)


src/tainting/Sig_inst.ml
```diff
@@ -349,86 +349,66 @@ let instantiate_shape inst_var inst_trace shape =
...
+       | _ -> None, None
+     in
+     match fparams, name, named_val with
+     | [], _, _ when List.is_empty pos_args ->
+        []
+     | [], _, _ ->
+        Log.err (fun m ->
+          m "function applied to more arguments than expected by the signature (%s)" err_ctx);
+          []
+     | _ :: fparams, name, Some v ->
```
dimitris-m: Would be easier to read if we used `Some name` here (which seems to be what `name` is).

src/tainting/Sig_inst.ml
```diff
@@ -349,86 +349,66 @@ let instantiate_shape inst_var inst_trace shape =
...
+        (name, v) :: merge fparams pos_args named_args
+     | _ :: _fparams , _, _ when List.is_empty pos_args ->
+        Log.err (fun m ->
+          m "function applied to fewer arguments than expected by the signature (%s)" err_ctx);
+          []
+     | _ :: fparams, name, _ ->
+        (name, List.hd pos_args) :: merge fparams (List.tl pos_args) named_args
```
dimitris-m: But when a named arg is not provided, should we not assume a `Null` value for it?

src/tainting/Sig_inst.ml
```diff
@@ -349,86 +349,66 @@ let instantiate_shape inst_var inst_trace shape =
...
-               Hashtbl.add idx_to_taints i taints;
-               (i + 1, rest))
-         (0, remaining_params)
+  let name_opt_value_list = merge fparams pos_args named_args in
+  let param_index_array = Array.of_list (List.map snd name_opt_value_list) in
+  let param_name_map =
+    name_opt_value_list
+    |> List.filter_map
+         (function (None, _) -> None | (Some n, v) -> Some (n, v))
+    |> SMap.of_list (* TODO: An assoc list? *)
```
dimitris-m: Suggestion: add some inline tests.

```
let%test _ =
  merge ... ... ... EQ expected_value
```

and so on.

dimitris-m: Serves well as documentation too.

tests/rules/cross_function_tainting/test_python_args.py
```diff
@@ -0,0 +1,141 @@
...
+
+def foo4(danger, ok1, ok2):
+    # ruleid: taint
+    sink(danger)
+    # ok:
+    sink(ok1)
+    # ok:
+    sink(ok2)
+
+foo4(0, 1, danger=source())
```
dimitris-m, line 45: But this is python-specific semantics no? What about other languages?

src/tainting/Sig_inst.ml
maciejpirog: Nope. When a named argument is not provided it is the following situation:

```python
foo(x, y):
  return x + y
  
foo(1, x=2)
```

which means the named arg `x` is provided with value `2`, while the named argument `y` is provided value `2`, although not using the name. Every argument always has a value, unless the arity doesn't match.

maciejpirog: i.e. the semantics in tainting is that of Python (except in Python you have to have named args at the end in application), not that of OCaml.

src/tainting/Sig_inst.ml
maciejpirog: Why? `name` could very well be a `None`

src/tainting/Sig_inst.ml
maciejpirog: 👍 good idea! I'll add a few!

tests/rules/cross_function_tainting/test_python_args.py
maciejpirog: As above. When we have a function call with some named arguments in AST, the semantics (given in this function) is that of Python. This means that if we have a naive translation to AST of languages with a different semantics, we can get wrong arguments. I don't think the original code paid much attention to such details, I'll look at some other languages to see what happens there.

maciejpirog: I mean this PR only offers a refactoring, with no change of behaviour. Not sure what the right place for a more language-specific semantics would be. Without static type info, we can know only when applying the function, which means we would have to make the behaviour of `find_pos_in_actual_args` language-dependent (which we can do, why not)

src/tainting/Sig_inst.ml
maciejpirog: ...because not all formal arguments have names. They are given by the type `Shape_and_sig.Signature.param`:

```ocaml
type param = P of string | Other [@@deriving eq, ord]
```

I guess `Other` now is used to hold the type of params that we don't support in tainting (like `ParamRest` or `ParamPattern`)

dimitris-m: But from : 

```
let name, named_val =
       match fparams with
       | Signature.P name :: _ -> Some name, List.assoc_opt name named_args
       | _ -> None, None
     in
```

I see that we cannot have None, Some _, am I missing something?

maciejpirog: Oh! I see what you mean. Yes, when we have a value for a name, we definitely have a name. I can refactor to make the code even more readable

PR #538, opened 2026-01-13
Intrafile tainting with variadic functions

maciejpirog:
We add tainting via "rest" params, as in:

```python
def test1(*args):
  for x in args.items():
      sink(x)

test1("abc", source(), "xyz")
```

This is done by adding a new kind of param (`ParamRest` in IL and `PRest` in function tainting signature) that corresponds to `ParamRest` in AST.

**NOTE**: In tainting signature db, the arity of a variadic function is still counted as if the "rest" param was a single param, so there might be problems when a variadic function is only one of possible overloads. But fixing that will be part of subsequent effort to upgrade naming and overloading resolution.

### DIFF
--- a/src/tainting/Shape_and_sig.ml
+++ b/src/tainting/Shape_and_sig.ml
@@ -599,10 +599,14 @@ end
  *)
 and Signature : sig
   (** A simplified version of 'AST_generic.parameter', we use 'Other' to
-      represent * parameter kinds that we do not support yet. We don't want to
-      just remove * those unsupported parameters because we rely on the position
-      of a parameter * to represent taint variables, see 'Taint.arg'. *)
-  type param = P of string | Other [@@deriving eq, ord]
+      represent parameter kinds that we do not support yet. We don't want to
+      just remove those unsupported parameters because we rely on the position
+      of a parameter to represent taint variables, see 'Taint.arg'. *)
+  type param =
+    | P of string
+    | PRest of string
+    | Other
+  [@@deriving eq, ord, show]
 
   type params = param list [@@deriving eq, ord]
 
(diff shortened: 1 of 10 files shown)


src/tainting/Shape_and_sig.ml
```diff
@@ -821,7 +816,7 @@ let lookup_signature (db : signature_database) (func_name : fn_id) (arity : int)
           let all_sigs = SignatureSet.elements sigs in
           let sig_details = all_sigs |> List.map (fun s ->
             Printf.sprintf "(arity=%d, params=%s)" s.arity
-              (s.sig_.params |> List.map (function Signature.P s -> "P(" ^ s ^ ")" | Signature.Other -> "Other") |> String.concat ",")
+              (s.sig_.params |> List.map Signature.show_param |> String.concat ",")
```
dimitris-m, line 819: I think we still need the `"P("` ... `)` part to remain compatible in terms of output?

dimitris-m: Same for `Other` vs `_?`

src/tainting/Shape_and_sig.ml
```diff
@@ -852,7 +847,7 @@ let add_signature (db : signature_database) (func_name : fn_id)
     (signature : extended_sig) : signature_database =
   (* Debug logging for ALL signatures being added *)
   let fn_name_str = show_fn_id func_name in
-  let params_str = signature.sig_.params |> List.map (function Signature.P s -> "P(" ^ s ^ ")" | Signature.Other -> "Other") |> String.concat "," in
+  let params_str = signature.sig_.params |> List.map Signature.show_param |> String.concat "," in
```
dimitris-m, line 850: Ok, maybe better, not sure. Maybe print `P(...)` and `PRest(...)` ?

This could be done by adapting `show_`, not necessarily here.

maciejpirog: It is for logging only, I believe

dimitris-m (approved):

src/tainting/Shape_and_sig.ml
dimitris-m: Up to you, lgtm for me.

PR #539, opened 2026-01-14
C#: Collection expressions vs attributes with targets (parser fix)

maciejpirog:
The parser confused collection expression that started with an identifier similar to an attribute target, e.g., `[type, ...]` etc with attributes with targets `[type: ...]`.


languages/csharp/generic/Parse_csharp_tree_sitter.ml
```diff
@@ -989,7 +984,7 @@ and element_binding_expression (env : env) (x : CST.element_binding_expression)
   let exprs = List_.map H2.argument_to_expr args in
   (open_br, exprs, close_br)
 
-and maybe_element_binding_expression (env : env) (x : CST.maybe_element_binding_expression)
+and maybe_element_binding_expression (env : env) (x : CST.maybe_bracketed_argument_list)
```
dimitris-m, line 987: This change is not strictly necessary given that we have: 

```
and maybe_element_binding_expression = maybe_bracketed_argument_list
```

right?

maciejpirog: yes, but the definition

```
and maybe_element_binding_expression = maybe_bracketed_argument_list
```

was in CST, which is auto-generated. I must have added it by accident in my last PR, and after I regenerated the parser, it disappeared, so I'm using the original type from CST

tests/parsing/csharp/collection-vs-attribute.cs
```diff
@@ -0,0 +1,10 @@
+namespace N
+{
+    public class C
+    {
+        void foo()
+        {
+            x = [type];
```
dimitris-m, line 7: aren't words like `type` reserved keywords?

maciejpirog: nope, `type` has no special meaning in C# except as a target in an attribute, as in the following:

```C#
[type: Obsolete("Use NewClass instead")]
public class OldClass { }
```

PR #544, opened 2026-01-17
ci: add options to skip Windows and OSX builds in rolling-release.yaml

### LINKED ISSUE #495: CI: Allow to skip Windows and maybe OSX using workflow parameters in rolling-release.yaml
We frequently produce binaries for testing using the release workflow, and in many cases these are tested in Linux so there is no need to wait for the slow parts (Windows).

Related #454.

WaterWhisperer:
Fixes: #495

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -6,6 +6,16 @@ on:
         description: the tag to use
         required: true
         type: string
+      skip_windows:
+        description: Skip Windows build (faster for Linux-only testing)
+        required: false
+        type: boolean
+        default: false
+      skip_osx:
+        description: Skip OSX build (faster for Linux-only testing)
+        required: false
+        type: boolean
+        default: false
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/rolling-release.yml
dimitris-m, line 113: Notice that this step will not run, since some of the pre-requisites won't happen when `skip_osx` or `skip_windows` is true.

dimitris-m, line 139: Even if the release does run, here it will fail when the osx artifact is not present.

dimitris-m (commented):
I'm a little divided about this... 

- some artifacts will be created, 
- the release itself will not happen (but it should not really happen without OSX and Win).

Need to think about it a bit.

.github/workflows/rolling-release.yml
WaterWhisperer: Thanks for the review!

I initially considered adding logic to handle partial releases as well, but wasn't sure what the expected behavior should be, so I went with a conservative approach: just providing the option to skip builds for now.

Let me know how you'd like the release step to behave, and I'll update accordingly!

WaterWhisperer: And as you said, 

> it should not really happen without OSX and Win

skipping Windows and OSX was initially intended to avoid waiting for the slower parts in testing, but I'm not quite sure how to handle them.

dimitris-m:
closing as we won't proceed with that issue any more, thanks in any case!

PR #552, opened 2026-01-21
add dump-intrafile-graph and  dump-taint-signatures 

corneliuhoffman:
To use
```:sh
bin/opengrep show dump-intrafile-graph tests/rules/cross_function_tainting/interprocedural.py
```

respectively
```sh
bin/opengrep show dump-taint-signatures tests/rules/cross_function_tainting/interprocedural.yaml tests/rules/cross_function_tainting/interprocedural.py
```

### DIFF
--- a/src/osemgrep/cli_show/Show_subcommand.ml
+++ b/src/osemgrep/cli_show/Show_subcommand.ml
@@ -227,43 +227,64 @@ let run_conf (caps : < caps ; .. >) (conf : Show_CLI.conf) : Exit_code.t =
                  (Fpath.to_string rule_file)
                  (Rule_error.string_of_error e))
       in
-      let rules =
+      let has_disabled_intrafile (r: Rule.t) =
+        match r.Rule.options with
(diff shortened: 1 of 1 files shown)


src/osemgrep/cli_show/Show_subcommand.ml
```diff
@@ -211,6 +211,59 @@ let run_conf (caps : < caps ; .. >) (conf : Show_CLI.conf) : Exit_code.t =
...
+               | _ -> false)
+      in
+      let _search_rules, taint_rules, _extract_rules, _join_rules =
+        Rule.partition_rules rules
+      in
+      (match taint_rules with
+      | [] ->
+          print "No taint rules found for this language";
+          Exit_code.ok ~__LOC__
+      | rule :: _ ->
```
dimitris-m: But in theory there could be > 1 applicable taint rules.

dimitris-m: I'll push an edit now

PR #553, opened 2026-01-22
Graph refactoring

corneliuhoffman:
## Summary
This si a refactoring consolidating PR. The main conceptual change is that in the intrafile graph/signature we have no need for the type of nodes/keys to be a name list but just a name. This cleans up the graph and simplifies the procedures.
  - Introduce `libs/graph_lib` library with shared types for call graph construction abstracting the graphs a bit
  - Refactor intrafile taint analysis to use unified graph types (`Call_graph_types.node`, `Call_graph_types.edge`, `Call_graph_types.G`)
  - Consolidate language-specific configurations into `Lang_config.ml` (HOF patterns, constructor models)
  - Simplify `Builtin_models.ml` and `Object_initialization.ml` by extracting shared logic
  
### Additional chages by @maciejpirog
- Move graph-related thing to `src/call_graph`. Rationale: it depends on AST and IL, so should not be in libs. Move some helper functions from `tainting/Function_call_graph` to the `call_graph/Call_graph` (in future, I want to remove the Function_call_graph module). Make Reachable module a module (not a functor). Rename it to `Graph_reachability`.
- Remove `callee_fn_id` from the type of edge labels. This is because it can be always recovered form the "src" of the edge.
- Remove some unused types.
- Introduce `Function_id`, which serves as an abstract type for identifying functions globally. It is used as the type of nodes in function call graph and as the type of keys in the signature db.
- Remove `fn_id` (the list of names) from Shape_and_sig. The `fn_id` type will be an internal type used when building a call graph from AST. Maybe removed completely when function name resolution is moved to naming.
- Move some things out of `tainting/Function_call_graph` and rename it to `tainting/Graph_from_AST`. Ideally, it would be moved to `src/call_graph`, but it depends on things from `src/tainting`, so for now it stays where it is.
- Add .mli files to `Graph_reachability` and `Graph_from_AST`. These modules now export only 1 and 3 functions respectively, which should clarify the code.

### DIFF
--- a/src/call_graph/Call_graph.ml
+++ b/src/call_graph/Call_graph.ml
@@ -3,10 +3,6 @@ module Log = Log_call_graph.Log
 (** Call graph node - just a function identifier (IL.name) *)
 type node = Function_id.t
 
-(* Helper to extract normalized key for node comparison.
-   Resolves file paths to canonical form to handle different representations
-   of the same file (e.g., /foo/bar vs /foo/baz/../bar). *)
-
 let hash_node = Function_id.hash
 
 let compare_node = Function_id.compare
--- a/src/call_graph/Graph_serialization.ml
+++ b/src/call_graph/Graph_serialization.ml
@@ -25,7 +25,7 @@ let export_dot (graph : G.t) (path : string) =
   close_out oc
 
 (* Export graph to JSON format (Graphology format for Cytoscape.js viewer) *)
-let graph_to_json (graph : G.t) : Yojson.Basic.t =
+let graph_to_graphology (graph : G.t) : Yojson.Basic.t =
   (* Generate unique node ID from name, file, line, col - avoids hash collisions *)
   let node_unique_id v =
     let name = Function_id.show v in 
--- a/src/call_graph/Function_id.ml
+++ b/src/call_graph/Function_id.ml
@@ -12,7 +12,10 @@
    speed up comparison. *)
 
 type t = IL.ident
-
+ 
+(* Helper to extract normalized key for node comparison.
+   Resolves file paths to canonical form to handle different representations
+   of the same file (e.g., /foo/bar vs /foo/baz/../bar). *)
 let normalize_file (file : Fpath.t) : string =
   try Unix.realpath (Fpath.to_string file)
   with Unix.Unix_error _ -> Fpath.to_string (Fpath.normalize file)


src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
+(* Call Graph Types - core types for call graph construction and analysis *)
+
+(** Position in a file (line/column) *)
+type position = { line : int; column : int [@key "character"] }
```
maciejpirog: Why a new type? Can't we simply use `Pos.t`?

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
...
+[@@deriving yojson, eq, show, ord]
+
+(** Range in a file (start/end positions) *)
+type range = { start : position; end_ : position [@key "end"] }
```
maciejpirog: As above, maybe reuse somethign we already have? (`Tok_range.t` maybe?)

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
...
+[@@deriving yojson, eq, show, ord]
+
+(** Function identifier - IL.name contains name string + token with file/position *)
+type function_id = IL.name option
```
maciejpirog: `IL.name` seems to have some irrelevant information, like `sid`, which is not stable across runs. Maybe use `Tok.location`?

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
...
+        Hashtbl.hash (fst n.IL.ident, filename, line, col)
+
+(** Call graph node - just a function identifier (IL.name) *)
+type node = IL.name
+[@@deriving show]
+
+(* Helper to extract normalized key for node comparison.
+   Resolves file paths to canonical form to handle different representations
+   of the same file (e.g., /foo/bar vs /foo/baz/../bar). *)
+let node_compare_key (v : node) =
```
maciejpirog: Code duplication: `hash_function_id` vs `node_compare_key`

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
...
+        Hashtbl.hash (fst n.IL.ident)
+      else
+        let file = Tok.file_of_tok tok in
+        let filename = Fpath.basename file in
+        let line = Tok.line_of_tok tok in
+        let col = Tok.col_of_tok tok in
+        Hashtbl.hash (fst n.IL.ident, filename, line, col)
+
+(** Call graph node - just a function identifier (IL.name) *)
+type node = IL.name
```
maciejpirog: `type node = function_id`?
I guess most of the code is oblivious to what the type `function_id` actually is, so let's not break the abstraction

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
...
+
+let compare_node n1 n2 =
+  compare (node_compare_key n1) (node_compare_key n2)
+
+let equal_node n1 n2 =
+  (node_compare_key n1) = (node_compare_key n2)
+
+(** Call graph edge - represents a call from one function to another *)
+type edge = {
+  callee_fn_id : IL.name;
```
maciejpirog: `callee_fn_id : function_id;` ?

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
...
+  let vertex_attributes _ = []
+  let default_edge_attributes _ = []
+  let edge_attributes _ = []
+  let get_subgraph _ = None
+end
+
+module Dot = Graph.Graphviz.Dot (Display)
+
+(** Node tracking for incremental updates *)
+let removed_node_keys : (string, unit) Hashtbl.t = Hashtbl.create 1000
```
maciejpirog: Hashtable on the toplevel. Not thread-safe

src/call_graph/Graph_serialization.ml
```diff
@@ -0,0 +1,112 @@
...
+let fpath_to_uri path = "file://" ^ Fpath.to_string path
+
+(* Save graph to marshal file *)
+let save_graph (graph : G.t) (path : string) =
+  let oc = open_out_bin path in
+  Marshal.to_channel oc graph [];
+  close_out oc
+
+(* Load graph from marshal file *)
+let load_graph (path : string) : G.t =
```
maciejpirog, line 15: Don't we need user-friendly error messages in case of read/write error?

src/call_graph/Call_graph.ml
dimitris-m: Also a starting size of 1000 is maybe not an optimal choice... start smaller?

dimitris-m: Ideal solution: no top level hashtable, can this be embedded where it's used?
Second-best: DLS assuming of course that our concurrency architecture has not changed in any way.

src/engine/Match_tainting_mode.ml
```diff
@@ -42,8 +42,7 @@ module Effects = Shape_and_sig.Effects
 module Signature = Shape_and_sig.Signature
 
 type fun_info = {
-  fn_id : Shape_and_sig.fn_id;
-  opt_name : IL.name option;
+  name : IL.name;
```
maciejpirog, line 45: Don't we prefer `name : Call_grap_types.function_id`?
As in the comments in Call_graph_types.ml: I think we should keep `name` an abstract type, as in most of the codebase we don't care about what it is exactly

src/engine/Match_tainting_mode.ml
```diff
@@ -242,7 +241,7 @@ let pms_of_effect ~match_on (effect_ : Effect.t) =
 (* Main entry points *)
 (*****************************************************************************)
 
-let check_fundef (taint_inst : Taint_rule_inst.t) name ctx ?glob_env ?class_name
+let check_fundef (taint_inst : Taint_rule_inst.t) (name : IL.name) ctx ?glob_env ?class_name
```
maciejpirog, line 244: Again: `function_id`?

src/engine/Match_tainting_mode.ml
maciejpirog, line 45: Maybe not relevant to this commit, but what is the exact purpose of `class_name_str` and `method_properties` in this type?

src/engine/Match_tainting_mode.mli
```diff
@@ -16,7 +16,7 @@ val hook_setup_hook_function_taint_signature :
 
 val check_fundef :
   Taint_rule_inst.t ->
-  Shape_and_sig.fn_id (** entity being analyzed *) ->
+  IL.name (** entity being analyzed *) ->
```
maciejpirog, line 19: Again: `IL.name` or `func_id`?

src/call_graph/Call_graph.ml
maciejpirog: Also, if you want to move parts of the call graph to `lib`, I don't think it should depend on anything in `IL`

maciejpirog: Also, if we want to have a single type we use to identify functions (at the moment the use of `IL.name` is quite confusing for me), which serves as the type of keys in the signature db and can be obtained by querying the LSP, we need to keep it in a more profound and specific place. Especially that it is a cross-file thing, so I would stress it strongly in the code that it as a different kind of being that leave outside the AST/IL of the current target. I propose that we add a separate module for this

maciejpirog: By "this" I mean a type that identifies a function within the entire project (not a single target). This module would also be a good place to keep functions that query LSP to convert a given location in the current target to the global location, etc.

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,128 @@
...
+
+(** Position in a file (line/column) *)
+type position = { line : int; column : int [@key "character"] }
+[@@deriving yojson, eq, show, ord]
+
+(** Range in a file (start/end positions) *)
+type range = { start : position; end_ : position [@key "end"] }
+[@@deriving yojson, eq, show, ord]
+
+(** Function identifier - IL.name contains name string + token with file/position *)
```
maciejpirog: I'm pretty sure I saw a comment that something file names in tokens are not 100% trustworthy, but I can't find it now. I'm also not sure if it was a global thing, or only at that given place. We need to make sure token is enough to identify a file

src/tainting/Shape_and_sig.ml
```diff
@@ -704,25 +704,33 @@ let make_fn_id_with_class class_name method_name =
 (** Helper to create fn_id with just a function name (no class) *)
 let make_fn_id_no_class name = [None; Some name]
 
+(* Function key for the signature database - uses just the function name (last element of fn_id).
+   This matches the graph vertex type in Function_call_graph.ml. *)
+type func_key = IL.name
```
maciejpirog: This is another place where we would benefit from having a proper type for global function identifiers instead of `IL.name`

maciejpirog (changes requested):
I think it is time to do some major clean up about types, identifiers etc we use in the codebase. I left some comments in particular files, but my major concern is the growing complexity and size of the code without introducing any abstraction over what we're doing. I propose that:

- We are more precise about how functions are identified. At the moment we have types like `Shape_and_sig.fn_id`, `Shape_and_sig.func_key`, `Call_grap_types.function_id`. and on top of that in most places of the code, we use `IL.name`. I propose we make a separate module, say `src/tainting/Signature_db`, in which we:

  - Explicitly define the one and only type to identify functions, as they are used cross-file. If possible, I would like to keep this type abstract,

  - Move the signature db code here,

  - Add code to deal with LSPs here as well,

  - Reason for the two above: I guess these are the only two things that need to know the internal structure of these identifiers.

  - Problems with the above: Does naming want to know the structure of these (also: we cannot keep two separate mechanism for naming and we need to kill off one of them)

- We don't put Call_graph_types.ml in lib, since it depends on things in src. Maybe we can make it a functor and parameterize with things from src inside src?

src/call_graph/Call_graph.ml
```diff
@@ -0,0 +1,152 @@
+module Log = Log_call_graph.Log
+
+(** Call graph node - just a function identifier (IL.name) *)
+type node = Function_id.t
+
+(* Helper to extract normalized key for node comparison.
```
maciejpirog: Remove this

maciejpirog: Put it in Function_id

src/call_graph/Graph_serialization.ml
```diff
@@ -0,0 +1,106 @@
...
+  graph
+
+(* Export graph to DOT format *)
+let export_dot (graph : G.t) (path : string) =
+  let oc = open_out path in
+  Dot.output_graph oc graph;
+  close_out oc
+
+(* Export graph to JSON format (Graphology format for Cytoscape.js viewer) *)
+let graph_to_json (graph : G.t) : Yojson.Basic.t =
```
maciejpirog: rename to graph_to_graphology_format

src/call_graph/Graph_serialization.ml
```diff
@@ -0,0 +1,106 @@
+(* Graph serialization - save/load/export functions for call graphs *)
+
+open Call_graph
+
+(* Convert path to file:// URI *)
+let fpath_to_uri path = "file://" ^ Fpath.to_string path
```
maciejpirog, line 6: sanitize this path

src/engine/Match_tainting_mode.ml
```diff
@@ -377,113 +376,79 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
                 | Arrow -> (
                     match opt_ent with
                     | None -> (infos, info_map)
-                    | Some _ ->
-                        let opt_name =
-                          let* ent = opt_ent in
-                          AST_to_IL.name_of_entity ent
+                    | Some ent ->
```
maciejpirog, line 379: What does this do?

src/call_graph/Graph_serialization.ml
maciejpirog: I don't think there is a need to sanitize this path: it is only used as a label in the exported json

maciejpirog (approved):

PR #555, opened 2026-01-22
Dart: Use case of identifier to guess Call vs New

maciejpirog:
In Dart, the `new` keyword is optional and the convention is to omit it. So, in theory, there is no difference between the syntax of a function call and a creation of a new object via a call to constructor (it is determined by the type checker in the compiler, and so we were translating both cases to `Call`). However, a strictly followed convention in Dart is that names of types begin with uppercase characters, while names of functions begin with with lowercase characters. We use that as a heuristic, translating to `Call` or `New` based on the case of the identifier.

This way, for the rule:

```yaml
  pattern-sources:
    - pattern: |
        (C $FOO).getData()
  pattern-sinks:
    - pattern: |
        sink(...)
```

We have:

```dart
  var bad = C();
  // ruleid: taint
  sink(bad.getData());

  var good = D();
  // ok:
  sink(good.getData());

  var good = c();
  // ok:
  sink(good.getData());
```

### DIFF
--- a/languages/dart/generic/Parse_dart_tree_sitter.ml
+++ b/languages/dart/generic/Parse_dart_tree_sitter.ml
@@ -1926,13 +1926,13 @@ and map_selector (env : env) (x : CST.selector) : expr -> expr =
   | `Assi_sele x -> map_assignable_selector env x expr
   | `Arg_part x ->
       let _tyargs_TODO, args = map_argument_part env x in
-      (* Just by looking at the syntax, we cannot know when a(...) is *)
-      (* a function call or a "new" expression, because the "new" keyword *)
-      (* is optional. However, there is almost universal convention in *)
-      (* Dart that types begin with an uppercase, while everything else *)
-      (* with a lowercase character. We use this convention as a heuristic: *)
-      (* if an identifier starts with an uppercase char, we convert it to *)
-      (* G.New; othwerwise, it is a call. *)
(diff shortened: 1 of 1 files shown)


languages/dart/generic/Parse_dart_tree_sitter.ml
```diff
@@ -1926,7 +1926,19 @@ and map_selector (env : env) (x : CST.selector) : expr -> expr =
...
   | `Arg_part x ->
       let _tyargs_TODO, args = map_argument_part env x in
-      Call (expr, args) |> G.e
+      (* Just by looking at the syntax, we cannot know when a(...) is *)
+      (* a function call or a "new" expression, because the "new" keyword *)
+      (* is optional. However, there is almost universal convention in *)
+      (* Dart that types begin with an uppercase, while everything else *)
+      (* with a lowercase character. We use this convention as a heuristic: *)
+      (* if an identifier starts with an uppercase char, we convert it to *)
+      (* G.New; othwerwise, it is a call. *)
```
dimitris-m: NIT: This comment style (independent lines) is a little weird imo. Hard to edit comments.

dimitris-m: Also a bit inconsistent with the rest of the codebase.

maciejpirog: Will fix

PR #556, opened 2026-01-29
Fixed a perf regression by removing system calls and improving the reachability graph and the callee lookup

corneliuhoffman:
This PR introduces 4 changes

 - removes the Unix.realpath that slowed everything down
 - Simplifies the reachability graph to compute in batch using OcamlGraph primitives, faster computation.
 - Improves the lookuop_callee_from_graph by not searching over all edges but only those going int the caller
 - adds a fix for elixir shortLambdas

### DIFF
--- a/src/call_graph/Graph_reachability.ml
+++ b/src/call_graph/Graph_reachability.ml
@@ -1,7 +1,5 @@
 module G = Call_graph.G
 module Bfs = Graph.Traverse.Bfs (G)
-module Oper = Graph.Oper.I (G)
-module Comp = Graph.Components.Make (G)
 
 type graph = G.t
 type vertex = G.V.t
@@ -38,17 +36,43 @@ let reachable_vertices_batch (g : graph) (starts : vertex list) : VSet.t =
     VSet.empty starts
 
 (* Compute the subgraph containing only functions relevant for taint flow
-   from sources to sinks. Optimized batched version using sets. *)
+   from sources to sinks. Excludes dead-end nodes that have no independent
+   source/sink connections. *)
 let compute_relevant_subgraph (graph : Call_graph.G.t)
     ~(sources : Function_id.t list) ~(sinks : Function_id.t list) : Call_graph.G.t =
   match (sources, sinks) with
   | [], _ | _, [] ->
       Call_graph.G.create ()
   | _ :: _, _ :: _ ->
-      (* Batch: compute reachable vertex SETS (no intermediate graphs) *)
+      let source_set = VSet.of_list sources in
+      let sink_set = VSet.of_list sinks in
+      let is_source_or_sink v = VSet.mem v source_set || VSet.mem v sink_set in
+
+      (* Batch: compute reachable vertex SETS *)
       let from_sources = reachable_vertices_batch graph sources in
       let from_sinks = reachable_vertices_batch graph sinks in
       (* Fast set intersection *)
       let common = VSet.inter from_sources from_sinks in
-      (* Reverse BFS from all common descendants to get ancestor edges *)
-      reverse_reachable_subgraph graph (VSet.elements common)
+
+      (* A node is relevant if:
+         1. It's a source or sink, OR
+         2. It has a predecessor that is source/sink or in XOR (entry point), OR
+         3. It has multiple predecessors in common (bridge between groups) *)
+      let is_relevant v =
+        is_source_or_sink v ||
+        (try
+          let preds = G.fold_pred (fun p acc -> p :: acc) graph v [] in
+          (* Entry point: has pred that is source/sink or in XOR *)
+          let is_entry = List.exists (fun pred ->
+            is_source_or_sink pred ||
+            (VSet.mem pred from_sources <> VSet.mem pred from_sinks)
+          ) preds in
+          (* Bridge: has multiple predecessors in common *)
+          let preds_in_common = List.filter (fun p -> VSet.mem p common) preds in
+          is_entry || List.length preds_in_common > 1
+        with _ -> false)
+      in
+      let relevant = VSet.filter is_relevant common in
+
+      (* Reverse BFS from relevant nodes on original graph to get ancestor edges *)
+      reverse_reachable_subgraph graph (VSet.elements relevant)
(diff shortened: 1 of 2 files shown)


src/tainting/Graph_from_AST.ml
```diff
@@ -443,25 +443,43 @@ let detect_user_hof (fdef : G.function_definition) : (string * int) list =
...
+      Some (callback_name, snd id, None)
   (* DotAccess: module.func or obj.method - common in Python/JS *)
   | G.DotAccess (_, _, G.FN (G.Id (id, id_info))) ->
       let callback_name = AST_to_IL.var_of_id_info id id_info in
-      Some (callback_name, snd id)
+      Some (callback_name, snd id, None)
+  (* Elixir: &func/n - ShortLambda wrapping a call to the named function.
+     Structure: OtherExpr("ShortLambda", [Params[&1,...]; S(ExprStmt(Call(func, args)))])
+     Create a _tmp node to match what AST_to_IL creates for the anonymous wrapper. *)
+  | G.OtherExpr (("ShortLambda", shortlambda_tok), [G.Params _; G.S { G.s = G.ExprStmt (inner_e, _); _ }]) ->
```
dimitris-m, line 471: In places like AST_to_IL we have a way to specify the language, for example: 
``` 
G.OtherExpr (("ShortLambda", _), _) when env.lang =*= Lang.Elixir
```

We also have short lambdas in Clojure, are these covered?

src/call_graph/Call_graph.ml
```diff
@@ -84,65 +84,43 @@ let add_edge (graph : G.t) ~(src : node) ~(dst : node) ~(call_tok : Tok.t) =
...
+        match exact_match with
+        | Some edge ->
+            Some (G.E.src edge)
+        | None ->
+            (* Fallback: check for implicit/HOF edges by matching line 0 *)
+            incoming_edges
+            |> List.find_opt (fun edge ->
+                let label = G.E.label edge in
+                Int.equal label.call_site.Pos.line 0)
+            |> Option.map G.E.src
```
dimitris-m, line 126: In fact one can do find_opt only once right? 

And return a type Explicit _ | Fallback _ or whatever.

Not so important for now.

src/tainting/Graph_from_AST.ml
```diff
@@ -592,12 +612,14 @@ let extract_hof_callbacks_from_call ~method_hofs ~function_hofs ~user_hofs ~all_
           | None -> []))
   | _ -> []
   in
-  (* Combine and deduplicate *)
-  all_callback_args @ configured_callbacks |> dedup_fn_ids
+  (* Combine - dedup would need to handle the tmp_opt too, skip for now *)
```
dimitris-m, line 616: Why do we skip dedup? Because _tmp are identified?

src/tainting/Graph_from_AST.ml
```diff
@@ -443,25 +443,43 @@ let detect_user_hof (fdef : G.function_definition) : (string * int) list =
...
+     Create a _tmp node to match what AST_to_IL creates for the anonymous wrapper. *)
+  | G.OtherExpr (("ShortLambda", shortlambda_tok), [G.Params _; G.S { G.s = G.ExprStmt (inner_e, _); _ }]) ->
+      (match inner_e.G.e with
+      | G.Call ({ e = G.N (G.Id (id, id_info)); _ }, _) ->
+          let callback_name = AST_to_IL.var_of_id_info id id_info in
+          (* Create _tmp IL.name using Tok.fake_tok like AST_to_IL.fresh_var does *)
+          let tmp_tok = Tok.fake_tok shortlambda_tok "_tmp" in
+          let tmp_name = IL.{
+            ident = ("_tmp", tmp_tok);
+            sid = G.SId.unsafe_default;
```
dimitris-m, line 479: But `AST_to_IL.fresh_var` has a proper sid: 
```
  let i = G.SId.mk () in
```
which I guess does not matter here?

src/call_graph/Graph_reachability.ml
```diff
@@ -29,103 +16,39 @@ end
...
-              Call_graph.G.iter_vertex (Call_graph.G.add_vertex result) subgraph;
-              Call_graph.G.iter_edges_e (Call_graph.G.add_edge_e result) subgraph)
-            sinks)
-        sources;
-      result
+      (* Batch: compute reachable vertex SETS (no intermediate graphs) *)
+      let from_sources = reachable_vertices_batch graph sources in
+      let from_sinks = reachable_vertices_batch graph sinks in
+      (* Fast set intersection *)
+      let common = VSet.inter from_sources from_sinks in
```
maciejpirog, line 55: How about the situation when the input graph is:

```
src -> foo <- snk
        |
        v
       bar
```

My understanding is that it will add `bar` to the relevant graph, although it is not really necessary

src/tainting/Graph_from_AST.ml
corneliuhoffman: will write a clojure test

src/tainting/Graph_from_AST.ml
corneliuhoffman: bc the dedup is already done by the graph

src/tainting/Graph_from_AST.ml
corneliuhoffman: yes the sids are irrelevant here

src/call_graph/Graph_reachability.ml
corneliuhoffman: true, this my plan for today, the graph lib has some trick for that, you need to look at SCC's that have no predecesors.

src/call_graph/Call_graph.ml
corneliuhoffman: yes but then you need to worry about whioch came first I agreee it might be mildly cheaper though I still have to convince myself wether the fallback ever happens

src/call_graph/Graph_reachability.ml
corneliuhoffman: we agreed we will keep the filtering for now, In principle things like this will be a problem but our HOF technology is not yet sophisticated enough to catch them anyway now:
```python
def f():
    return source()

def g(x):
    # ruleid: test-tuple-unpacking-taint
    sink(x)
    return x

def h(x):
    return (f(), lambda y: g(y))

def k(x):
    a, b = h(x)
    z, w = b(a)
```

src/tainting/Graph_from_AST.ml
corneliuhoffman: I think Clojure is not yet supported

PR #557, opened 2026-01-30
New benchmarking using hyperfine

dimitris-m:
## Changes

Must merge https://github.com/opengrep/semgrep-interfaces/pull/4 first -- this is now DONE. 

See `perf/opengrep-scripts/README.md` for details.

### Benchmarking 

  - Replace Python bench scripts with shell-based scripts using [hyperfine](https://github.com/sharkdp/hyperfine]).
  - Add bench.sh, bench-against-sha.sh, compare-outputs.sh, and supporting utilities.
  - New `opengrep-diff` tool for comparing scan outputs between versions.

### Performance improvements

  - Remove unnecessary system calls.  In some cases `--time` was making scans > 2 times slower... now the performance is almost the same. 
  - Make `--time` output more compact and space-efficient.  In large repos like grafana the output json with `--time` was around 20M and now it's around 3M (close to the size without `--time`).  

### Small fixes and technical debt

  - Fix Python CLI `--time` output.
  - Remove unused Common.jsoo (JS target dropped).
  - Remove unused and commented out CLI arguments.
  - Fix opengrep email in cli/setup.py.
  - Add missing opam/parser_vbnet.opam file.


perf/opengrep-scripts/.semgrepignore
```diff
@@ -0,0 +1,32 @@
...
+_build/
+_cargo/
+# note that PHP composer uses vendor/ and C++ conan uses build/
+# .venv is used both by Go and Python
+
+# Common test paths
+test/
+tests/
+testsuite/
+*_test.go
```
corneliuhoffman, line 32: for future maybe add .opengrep

dimitris-m: but I have not seen .opengep in the wild, but yes we can have a conversation about the default ignore in general: this is a copy of the standard .semgrepignore in our repo, I added it to ensure it's used consistently in benchmarks but now we need to see what happens with opengrep vs semgrep because of v2.

PR #577, opened 2026-02-09
In `--experimental`, don't report git untracked files as skipped with `--use-git-ignore`

### LINKED ISSUE #575: Every git scan with --experimental reports skipped files
## Problem

It happens even if the root is not a git repo. For example:

```
$ ls -a code
.        ..       hello.py
$ cat code/hello.py
print('hello')
$ opengrep scan --experimental -c rules.yaml code
...
┌──────────────┐
│ Scan Summary │
└──────────────┘
Some files were skipped or only partially analyzed.
  Scan was limited to files tracked by git.

Ran 1 rule on 1 file: 1 finding.
```

The `--verbose` flag does not reveal any concrete files:

```
$ opengrep scan --experimental -c rules.yaml code --verbose
...
========================================
Files skipped:
========================================

  Always skipped by Opengrep:

   • <none>

  Skipped by .gitignore:
  (Disable by passing --no-git-ignore)

   • <all files not listed by `git ls-files` were skipped>

  Skipped by .semgrepignore:
  (See: https://semgrep.dev/docs/ignoring-files-folders-code/#understand-semgrep-defaults)

   • <none>

  Skipped by --include patterns:

   • <none>

  Skipped by --exclude patterns:

   • <none>

  Skipped by limiting to files smaller than 1000000 bytes:
  (Adjust with the --max-target-bytes flag)

   • <none>

  Partially analyzed due to parsing or internal Opengrep errors

   • <none>
```

The problem disappears when we don't use gitignore:

```
$ opengrep scan --experimental -c rules.yaml code --no-git-ignore
...
┌──────────────┐
│ Scan Summary │
└──────────────┘

Ran 1 rule on 1 file: 1 finding.
```

## Reason

In `src/osemgrep/cli_scan/Scan_subcommand.ml` we call `Summary_report.pp_summary` given it simply the flag from the config:

```ocaml
          Logs.app (fun m ->
              m "%a"
                (Summary_report.pp_summary
                   ~respect_gitignore:conf.targeting_conf.respect_gitignore
                   ~maturity:conf.common.maturity
                   ~max_target_bytes:conf.targeting_conf.max_target_bytes
                   ~skipped_groups)
                ());
```

while inn `src/osemgrep/reporting/Summary_report.ml` we first check if the argument was set to `true`:

```ocaml
  let out_limited =
    if respect_gitignore then
      Some "Scan was limited to files tracked by git."
    else None
```

and then print out the message if it was, regardless if any files were actually ignored

```ocaml
  match (out_skipped, out_partial, out_limited, skipped_groups.ignored) with
  | [], None, None, [] -> ()
  | xs, parts, limited, _ignored -> (
      Fmt.pf ppf "Some files were skipped or only partially analyzed.@\n";
```

## Expected behaviour:

Either:
1. If no other skipped files detected, only print the message "Scan was limited to files tracked by git", but don't print the message "Some files were skipped or only partially analyzed".
2. As above, but print the message only if the root is a git repo
3. Check if any files were actually skipped (which is the case in practice, because in most cases we have things like build artifacts in the scanned repo)

**EDIT**: A scan can have multiple roots, so point 2 needs to take it into account

maciejpirog:
Closes #575

Before, when run `--experimental` with `--use-git-ignore` (default), we **always** reported that some files were skipped. This was counterintuitive, because it didn't mean that any files were actually skipped.

Now, if the flag is set **and** the scanned root (more precisely, at least one of the scanned roots) is a git repo, we print the message `Scan was limited to files tracked by git` separate from the `Some files were skipped or only partially analyzed` message. This way, when the user sees the latter, it means that there were actually files skipped for a reason (different than not being tracked by git).

### Example when both messages appear:

```
┌──────────────┐
│ Scan Summary │
└──────────────┘
Scan was limited to files tracked by git.
Some files were skipped or only partially analyzed.
  Scan skipped: 76 files matching .semgrepignore patterns.
  For a full list of skipped files, run opengrep with the --verbose flag.

Ran 2 rules on 460 files: 8038 findings.
```

### Example of a scan of a git repo when no files tracked by git are actually skipped:

```
┌──────────────┐
│ Scan Summary │
└──────────────┘
Scan was limited to files tracked by git.

Ran 1 rules on 2 files: 2 findings.
```

### Example of a scan of a root that is not a git repo, but there are files skipped:

```
┌──────────────┐
│ Scan Summary │
└──────────────┘
Some files were skipped or only partially analyzed.
  Scan skipped: 1 files larger than 1 MB, 1 files matching .semgrepignore patterns.
  For a full list of skipped files, run opengrep with the --verbose flag.

Ran 2 rules on 5 files: 315 findings.
```

### DIFF
--- a/src/targeting/Find_targets.ml
+++ b/src/targeting/Find_targets.ml
@@ -126,16 +126,6 @@ type project_root =
 
 and git_remote = { url : Uri.t } [@@deriving show]
 
-module Fppath_set = struct
-  module Self = Set.Make (Fppath)
-  include Self
-
-  (* This is for occasional debugging *)
-  let[@warning "-unused-value-declaration"] show set =
-    spf "[%s]"
-      (set |> Self.elements |> List_.map Fppath.show |> String.concat ", ")
-end
-
 (* Yet another file path related type ...
 
    This module is a bit fragile as it assumes that target file paths found in
@@ -251,6 +241,7 @@ let default_conf : conf =
 (* The actual returned type *)
 (*************************************************************************)
 
+(* 'a is either Fpath or Fppath *)
 type 'a targets = {
   selected : 'a list;
   skipped : Semgrep_output_v1_t.skipped_target list;
@@ -359,7 +350,7 @@ let filter_path (ign : Gitignore.filter)
 *)
 let filter_paths
     ((ign, include_filter) : Gitignore.filter * Include_filter.t option)
-    (target_files : Fppath.t list) : Fppath_set.t * Out.skipped_target list =
+    (target_files : Fppath.t list) : Fppath.t list * Out.skipped_target list =
   let (selected_paths : Fppath.t list ref) = ref [] in
   let (skipped : Out.skipped_target list ref) = ref [] in
   let add path = Stack_.push path selected_paths in
@@ -380,8 +371,11 @@ let filter_paths
          | Skip x -> skip x
          | Ignore_silently ->
              Log.debug (fun m -> m "ignore silently: %s" !!(fppath.fpath)));
-  (Fppath_set.of_list !selected_paths, !skipped)
+  (!selected_paths, !skipped)
 
+(* Note: throughout this file we use List.rev_append instead of (@) for
+ * concatenating file lists, since it is tail-recursive. The order does not
+ * matter because we sort and deduplicate at the end in get_targets. *)
 let filter_size_and_minified max_target_bytes exclude_minified_files paths =
   let selected_fppaths, skipped_size =
     Result_.partition
@@ -401,7 +395,7 @@ let filter_size_and_minified max_target_bytes exclude_minified_files paths =
   in
   Log.debug (fun m -> m "skipped_size: %d" (List.length skipped_size));
   Log.debug (fun m -> m "skipped_minified: %d" (List.length skipped_minified));
-  (selected_fppaths, skipped_size @ skipped_minified)
+  (selected_fppaths, List.rev_append skipped_size skipped_minified)
 
 (*************************************************************************)
 (* Finding by walking *)
@@ -488,14 +482,14 @@ let walk_skip_and_collect (ign : Gitignore.filter)
 *)
 let git_list_files ~exclude_standard
     (file_kinds : Git_wrapper.ls_files_kind list)
-    (project_roots : Project.roots) : Fppath_set.t option =
+    (project_roots : Project.roots) : Fppath.t list option =
   Log.debug (fun m ->
       m "Find_targets.git_list_files for project %s"
         (Project.show project_roots.project));
   let project = project_roots.project in
   (* TODO: we should not call git_list_files when the project
    * is not a Git_project. We should assert it and not return
-   * an option type but an Fppath_set.t instead.
+   * an option type but an Fppath.t list instead.
    *)
   match project.kind with
   | Git_project ->
@@ -592,8 +586,7 @@ let git_list_files ~exclude_standard
                  (* scanning root is neither a file nor a folder *)
                  Log.warn (fun m ->
                      m "invalid scanning root %s" !!(sc_root.fpath));
-                 []))
-        |> Fppath_set.of_list)
+                 [])))
   | _ -> None
 
 (*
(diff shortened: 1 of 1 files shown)


src/osemgrep/cli_scan/Diff_scan.ml
```diff
@@ -175,7 +175,7 @@ let scan_baseline_and_remove_duplicates (caps : < Cap.chdir ; Cap.tmp >)
                 | PRO Engine_type.{ analysis = Interprocedural; _ } ->
                     (* Keeping all the PRO stuff since we will probably need
                      * these code paths. *)
-                    let all_in_baseline, _ =
+                    let { Find_targets.targets = all_in_baseline; _ } =
```
dimitris-m: This is PRO engine path

dimitris-m: ok type change

src/osemgrep/cli_scan/Ls_subcommand.ml
```diff
@@ -14,16 +14,16 @@ type format = Paths_only | Long [@@deriving show]
 let default_format = Paths_only
 
 let run ~target_roots ~targeting_conf:conf ~format () =
-  let selected, skipped = Find_targets.get_target_fpaths conf target_roots in
-  selected |> List.sort Fpath.compare
+  let targets = Find_targets.get_target_fpaths conf target_roots in
+  targets.targets |> List.sort Fpath.compare
```
dimitris-m: maybe `targets.selected` would be more clear?

dimitris-m: and would match `targets.skipped`

maciejpirog: I think you are right. I will change it to selected

maciejpirog: Fixed

src/targeting/Find_targets.ml
```diff
@@ -898,34 +910,33 @@ let clone_if_remote_project_root conf =
 (* Entry point *)
 (*************************************************************************)
 
-let get_targets conf scanning_roots : Fppath.t list * Out.skipped_target list =
+(* TODO: The 'git_repo' fields is needed to print out a warning to the *)
+(* user, because some files in a git repo can be ignored. When multiple *)
+(* roots are specified, we display this warning to the user when *)
+(* at least one root is a git repo. Should we be more precise and *)
+(* display which roots are git repos? *)
```
dimitris-m: Comments are like this everywhere:

```ocaml
(*
 *
 *
*)
```

But here each line is a comment.

dimitris-m: `git_repo` field (not fields)

src/targeting/Find_targets.ml
```diff
@@ -898,34 +910,33 @@ let clone_if_remote_project_root conf =
...
+let get_targets conf scanning_roots : Fppath.t targets =
   clone_if_remote_project_root conf;
-  let grouped_scanning_roots =
-    scanning_roots |> group_scanning_roots_by_project conf
+  let path_set, raw_skipped, git_repo =
+    List.fold_left
+      (fun (t, s, gr) root ->
+        let t', s', gr' = get_targets_for_project conf root in
+        Fppath_set.union t t', s' @ s, gr || gr')
```
dimitris-m: I would rename these variables `(t, s, gr)` to something that's immediately clear.

src/targeting/Find_targets.ml
```diff
@@ -866,14 +877,15 @@ let get_targets_for_project conf (project_roots : Project.roots) =
...
+  let is_git_repo = Option.is_some git_tracked in 
   let selected_targets, skipped_targets =
     force_select_scanning_roots
       ~apply_includes_excludes_to_files:conf.apply_includes_excludes_to_file_targets
       project_roots
       selected_targets
       skipped_targets
   in
-  (selected_targets, skipped_targets)
+  (selected_targets, skipped_targets, is_git_repo)
```
dimitris-m: But we have a record type now, can we not reuse?

src/osemgrep/reporting/Summary_report.ml
```diff
@@ -52,9 +51,7 @@ let pp_summary ~respect_gitignore ~(maturity : Maturity.t) ~max_target_bytes
                      targets_not_in_git += 1
                      continue
          if targets_not_in_git != dir_targets: *)
-      Some "Scan was limited to files tracked by git."
-    else None
-  in
+    Fmt.pf ppf "Scan was limited to files tracked by git.@\n";
```
dimitris-m, line 54: Indentation of nested comment block.

src/targeting/Find_targets.ml
maciejpirog: > Comments are like this everywhere

Yeah, a setting somehow changed in my editor, and I keep forgetting to fix it....

dimitris-m (approved):
lgtm

PR #582, opened 2026-02-11
Tainting: per-arity signature extraction for Clojure and Elixir

### LINKED ISSUE #561: Improved clojure HOFs parsing
Tried the clojure version of the HOFs test and failed. Attaching an automatically generated version of the test (added txt bc github)

[test_hof_comprehensive_clojure.yaml](https://github.com/user-attachments/files/25016097/test_hof_comprehensive_clojure.yaml)

[test_hof_comprehensive_clojure.clj.txt](https://github.com/user-attachments/files/25016088/test_hof_comprehensive_clojure.clj.txt)

dimitris-m:
## Summary

- Extract **N separate taint signatures** (one per arity branch) for Clojure multi-arity functions and Elixir multi-clause lambdas, instead of a single merged signature with `params = [PRest !!_implicit_param!]`
- Add `sig_arity` type (`Arity_exact | Arity_at_least`) for rest-param support, with fallback lookup semantics
- Add built-in HOF configs for Clojure (`map`, `filter`, `keep`, `reduce`, `mapcat`, etc.)
- Unwrap Clojure CList arguments so individual args are visible to taint analysis
- Enable **implicit return** for Clojure (`lang_supports_implicit_return`), enabling cross-function return-value taint propagation (e.g. `(defn get-data [x] (source))` now produces a `ToReturn` effect)
- Bridge the `OtherExpr("ExprBlock", ...)` wrapper gap in `AST_to_IL.stmt_aux`: Clojure's AST wraps function bodies in `ExprBlock`, so `mark_first_instr_ancestor` sets `is_implicit_return` on the inner expression but `stmt_aux` checked the outer wrapper. Now propagates through by checking the last expression in the block.
- Remove redundant `detect_user_hof` from call graph construction: `all_callback_args` already identifies every function reference passed as an argument, making the `user_hofs` path purely duplicative (confirmed by DOT graph comparison and full test suite)

Closes #561

## Checks

- [x] `test_hof_comprehensive_clojure`: custom HOF with named callback, inline lambda, multi-arity delegation, built-in `map`/`filter`/`reduce`/`keep`/`mapcat`/etc.
- [x] `test_hof_comprehensive_clojure`: cross-function return taint via implicit return (`get-history`/`mapcat`/`process-history` from issue #561)
- [x] `test_hof_comprehensive_clojure`: variadic arity (`& rest`) delegation and rest-param sink
- [x] `test_hof_multi_clause_elixir`: multi-clause `fn` lambda as HOF callback, guarded clauses (`PatWhen`)
- [x] All 72 `cross_function_tainting` tests pass after `detect_user_hof` removal

## TODO: Destructuring

Currently HOFs are not identified when they are inside destructuring patterns, we need more work for this and I left it for a future PR.

### DIFF
--- a/src/tainting/Shape_and_sig.ml
+++ b/src/tainting/Shape_and_sig.ml
@@ -644,13 +644,7 @@ end = struct
   let of_IL_params il_params =
     il_params
     |> List_.map (function
-         | IL.Param { pname = { ident = s, _; _ }; _ } ->
-             (* Clojure's !!_implicit_param! is a synthetic wrapper for all actual
-              * arguments. Mark it as PRest so that find_pos_in_actual_args gathers
-              * the unwrapped CList arguments into a combined indexed shape, enabling
-              * correct resolution of per-element offsets (Oint i) during signature
-              * instantiation. *)
-             if AST_generic.is_implicit_param s then PRest s else P s
+         | IL.Param { pname = { ident = s, _; _ }; _ } -> P s
          (* functions signatures don't look into the shape of the argument. *)
          | IL.ParamRest { pname = { ident = s, _; _ }; _ } -> PRest s
          | IL.ParamPattern pat -> (
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -1792,7 +1792,13 @@ and parameters params : param list =
   params |> Tok.unbracket
   |> List_.map (function
        | G.Param { pname = Some i; pinfo; pdefault; _ } ->
-           Param { pname = var_of_id_info i pinfo; pdefault }
+           let pname = var_of_id_info i pinfo in
+           (* Clojure/Elixir/OCaml encode multi-clause functions with a
+              single synthetic !!_implicit_param! that wraps all actual
+              arguments. Translate it as ParamRest so the taint signature
+              layer treats it as a rest param without a special-case check. *)
+           if G.is_implicit_param (fst i) then ParamRest { pname; pdefault }
+           else Param { pname; pdefault }
        | G.ParamRest (_, { pname = Some i; pinfo; pdefault; _ }) ->
            ParamRest { pname = var_of_id_info i pinfo; pdefault }
        | G.ParamPattern pat -> ParamPattern pat


src/engine/Match_tainting_mode.ml
```diff
@@ -534,48 +534,199 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
-              let _flow, fdef_effects, _mapping =
-                check_fundef taint_inst info.name !ctx ~glob_env
-                  ?class_name:info.class_name_str ~signature_db:updated_db
-                  ?builtin_signature_db ?call_graph:(Some relevant_graph)
-                  info.fdef
-              in
-              record_matches fdef_effects;
-              updated_db
+              match fdef.G.fbody with
+              | G.FBStmt { G.s = G.Switch (_, _, cases); _ } ->
```
maciejpirog: Don't we want to check if we switch on the implicit arg, not to confuse with legitimate switch in a function with an implicit arg? Or are implicit args used only to simulate multi-arity functions?

src/engine/Match_tainting_mode.ml
```diff
@@ -534,48 +534,199 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
-            let params = Tok.unbracket info.fdef.fparams in
-            let arity = get_arity params info lang in
-            let updated_db, _signature =
-              Taint_signature_extractor.extract_signature_with_file_context
-                ~arity ~db ?builtin_signature_db taint_inst ~name:info.name
-                ~method_properties:info.method_properties
-                ~call_graph:(Some relevant_graph) info.cfg ast
+          (* Convert a Case pattern back into a [G.parameter list] for per-arity
+           * signature extraction (Clojure multi-arity / Elixir multi-clause). *)
+          let params_of_case_pattern (pat : G.pattern) : G.parameter list =
```
maciejpirog: Would it be much work to extract `params_of_case_pattern` and `process_fun_info` out of `check_rule`? At the moment, `check_rule` has 600 LOC and is totally human-unreadable.

src/tainting/Shape_and_sig.ml
```diff
@@ -765,18 +804,41 @@ let lookup_signature (db : signature_database) (name : Function_id.t) (arity : i
...
-      let total_sigs_card = SignatureSet.cardinal sigs in      
+      let total_sigs_card = SignatureSet.cardinal sigs in
       (* If there's only one signature for this function name, return it regardless of arity *)
       if total_sigs_card =*= 1 then
         Some (SignatureSet.choose sigs).sig_
       else
-        let filtered_sigs =
-          SignatureSet.filter (fun x -> Int.equal arity x.arity) sigs
+        (* Try exact match first *)
+        let exact =
```
maciejpirog: Code duplication (vs line 763)

src/engine/Match_tainting_mode.ml
maciejpirog: Disregard, I'm reading commit-by-commit and it was indeed fixed in the next commit

src/engine/Match_tainting_mode.ml
dimitris-m: Implicit args are only used to simulate, there is no legitimate function using them and I don't expect this to be an issue because the implicit arg is not a valid clojure identifier.

src/engine/Match_tainting_mode.ml
dimitris-m: no, I agree, will refactor

src/tainting/Shape_and_sig.ml
maciejpirog: Disregard. Fixed in the next commit

src/engine/Match_tainting_mode.ml
dimitris-m: but I did do a refactor step

src/engine/Match_tainting_mode.ml
maciejpirog: Maybe just in case we should restrict this to Clojure and Elixir? Unless I missed something, this transformation is not guarded by a `Lang`

maciejpirog:
lgtm

src/engine/Match_tainting_mode.ml
dimitris-m: Good point, will do

dimitris-m: I think maybe this is not needed because we may have more such cases, for example OCaml `function`.

src/engine/Match_tainting_mode.ml
```diff
@@ -276,6 +276,102 @@ let get_arity params info lang =
   in
   List.length filtered_params
 
+(** Convert a Case pattern back into a [G.parameter list] for per-arity
+    signature extraction (Clojure multi-arity / Elixir multi-clause). *)
```
corneliuhoffman, line 280: should we guard for clojure/elixir nly ... not knowledgable enough to see if this can happen "by mistake" in other langs?

src/tainting/Shape_and_sig.ml
```diff
@@ -644,7 +644,13 @@ end = struct
   let of_IL_params il_params =
     il_params
     |> List_.map (function
-         | IL.Param { pname = { ident = s, _; _ }; _ } -> P s
+         | IL.Param { pname = { ident = s, _; _ }; _ } ->
```
corneliuhoffman: is this not a job of AST_to_IL? looks wrong to me that that is a Param?

src/engine/Match_tainting_mode.ml
dimitris-m: The encoding is detected quite precisely, in fact OCaml's `function` also uses it -- not sure it is worth to gate this.

dimitris-m: That is if you look at `extract_multi_arity_cases` where this is used.

maciejpirog (approved):

src/tainting/Shape_and_sig.ml
dimitris-m: Maybe it could be PRest yes

dimitris-m: ok you are right, I will change

dimitris-m: fixed

Issue #99, opened 2025-02-20
Release Opengrep container


notdurson:
**Describe the solution you'd like**

As a corporate SAST engineer, I'd like to be able to pull a pre-built Opengrep container into my private Amazon ECR repository for internal use. Doing so will enable us to run internal scans of our source code using Opengrep, instead of our existing solution.

We have considered building our own container; however, maintenance of same will take work, and we'd prefer not to do that.

notdurson:
I am happy to work on this.

willem-delbare:
we'd love this contribution @notdurson

notdurson:
@dimitris-m do you care what form the container takes? I looked at the existing container build workflow and I have some concerns:

1. It uses Depot. AFAIK, the Opengrep project does not, so the workflow will have to be rewritten to use GHA. 
2. It's complex; it builds Opengrep from source instead of importing a pre-built binary. 
3. It tries to publish the image to DockerHub.

How much of this would you like to keep? I can very easily create a minimal dockerfile which builds Opengrep's latest manylinux86 binary into a minimal [Wolfi](https://cgr.dev/chainguard/wolfi-base) container, but I'm happy to make it more nuanced.

dimitris-m:
> @dimitris-m do you care what form the container takes? I looked at the existing container build workflow and I have some concerns:
> 
> 1. It uses Depot. AFAIK, the Opengrep project does not, so the workflow will have to be rewritten to use GHA. 
> 2. It's complex; it builds Opengrep from source instead of importing a pre-built binary. 
> 3. It tries to publish the image to DockerHub.
> 
> How much of this would you like to keep? I can very easily create a minimal dockerfile which builds Opengrep's latest manylinux86 binary into a minimal [Wolfi](https://cgr.dev/chainguard/wolfi-base) container, but I'm happy to make it more nuanced.

Sorry, I forgot to mention. The existing dockerfile of Semgrep is to be discarded indeed.

If it is to be hosted here we should probably build from source as we do in the workflow.

Else we can build a minimal dockerfile that pulls in the Linux binary..

dimitris-m:
And if we are to include a Dockerfile in this repo, we might as well publish to dockerhub.

notdurson:
I don't know if it's necessary to include a fresh source build in the Docker container workflow; it can certainly call the _existing_ workflow, though. No issues there. Will prototype and present shortly.

notdurson:
[...]

PR #583, opened 2026-02-11
Fix Dockerfile to work with current opengrep codebase

backkem:
## Summary

The existing Dockerfile has not been updated since the fork and does not build.
This PR fixes it based on what we learned building opengrep in our own CI pipeline.

- Use `ocamlpro/ocaml:5.3.0` as builder base (same as `.github/workflows/build-test-core-x86.yml`)
- Fix `semgrep.opam` path → `opam/semgrep.opam`
- Pin `memprof-limits` dev before opam dependency resolution
- Add static link deps (`libidn2`, `libunistring`, `libpsl`, `zstd`) required by `src/main/flags.sh`
- Install binary as `opengrep-core` (not `Main.exe` / `semgrep-core`)
- Drop semgrep-pro, wheel, and perf-test stages

Tested: builds and runs successfully (`opengrep scan --config auto .`).

Relates to #99 having a working Dockerfile in-repo would also make it straightforward to publish to Docker Hub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,259 +1,152 @@
-# TODO: should adapt: semgrep => opengrep
-
 ###############################################################################
 # Overview
 ###############################################################################
-# Dockerfile to build the semgrep/semgrep:canary docker image
-# (see .github/workflows/{build-test-docker,push-docker}.jsonnet).
+# Dockerfile to build the opengrep Docker image.
 #
-# First, we build a fully *static* 'semgrep-core' binary on Alpine. This
+# First, we build a fully *static* 'opengrep-core' binary on Alpine. This
 # binary does not even depend on Glibc because Alpine uses Musl instead
 # which can be statically linked.
 #
-# Then 'semgrep-core' alone is copied to another Alpine-based container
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,377 +1,158 @@
...
-# installed them under /usr/local/lib/python3.xx/site-packages/semgrep/
-RUN ln -s semgrep-core /usr/local/bin/osemgrep && rm -rf /pysemgrep
-
+# Get opengrep-core from step1
+COPY --from=core-builder /src/opengrep/_build/install/default/bin/opengrep-core /usr/local/bin/opengrep-core
 
+# The CLI looks for opengrep-core in PATH (see entrypoint.py).
+# Keep semgrep-core and osemgrep as compatibility symlinks.
+RUN ln -s opengrep-core /usr/local/bin/semgrep-core && \
+    ln -s opengrep-core /usr/local/bin/osemgrep && \
```
dimitris-m, line 135: What would the point of these be exactly?

backkem: You're right, dropped. The binary is already built as opengrep-core by the Makefile and installed directly as  /usr/local/bin/opengrep-core.

Issue #528, opened 2026-01-08
Opengrep version v1.11.5 is taking a huge performance hit when it comes to excluding directories


somakdutta:
Open grep Version : v1.11.5

When scanning a directory which has massive number of files , deeply nested, which needs to be excluded is taking a far longer time than expected .

Eg 
a single repo with a simple Helloworld.kt file, open grep takes ~2mins to complete scanning.

You can reproduce the same with the script attached.( it takes 2mins for a simple 5line file scan)

Please help. Happy to hear suggestions to improve this further

Please note this is important : for many maven pipelines, CI devs prefer to retain .m2 directory within the workspace and use --exclude to exclude the same from opengrep.

[benchmark.sh](https://github.com/user-attachments/files/24496187/benchmark.sh)

Regards,
Somak

somakdutta:
> What happens if you add the m2 directory to . semgrepignore?

I dont see any difference in performance with .semgrepignore . This is what i added in the script

```
cat > "$BENCHMARK_DIR/.semgrepignore" << 'EOF'
# Semgrep ignore file

# Ignore Maven local repository
.m2/
.m2/**

# Alternative patterns (for completeness)
**/.m2/
EOF
```

somakdutta:
> > > What happens if you add the m2 directory to . semgrepignore?
> > 
> > 
> > I dont see any difference in performance with .semgrepignore . This is what i added in the script
> > ```
> > cat > "$BENCHMARK_DIR/.semgrepignore" << 'EOF'
> > # Semgrep ignore file
> > 
> > # Ignore Maven local repository
> > .m2/
> > .m2/**
> > 
> > # Alternative patterns (for completeness)
> > **/.m2/
> > EOF
> > ```
> 
> Still, 2 minutes is not a long time in absolute terms, not sure if this is high priority now.
> 
> But let's leave the issue open. Maybe related to [#2](https://github.com/opengrep/opengrep/issues/2).

Actually 2 mins is the time taken to scan just a single .kt file with just 10lines of code. If this can be looked at on priority would be appreciated.

TreetopTechie:
Hi, it doesn't look like this is resolved. Would it be possible to look at it again please?

Old version (1.11.5):

```
➜  opengrep-issue git:(main) ✗ opengrep --version
1.11.5
➜  opengrep-issue git:(main) ✗ time 1>/dev/null 2>/dev/null opengrep scan -c $RULES HelloWorld.kt
[...]

PR #596, opened 2026-02-18
Python CLI: Optimize `--exclude` and .semgrepignore to ignore directories without listing their content

maciejpirog:
## Overview

Relevant issue: #528

**Current status:**

- In python, when you exclude a dir, it still needs to list all files in that dir
- This is optimized with `--experimental` and here we try to achieve similar behaviour

## Tests

Tested with

```
pytest tests/default/unit/targeting/test_target_manager.py tests/default/e2e/test_target_selection.py -v
```

which seem sufficiently comprehensive

## Evaluation of performance

The env for testing is set up using the script atttached to #528, in particular:

```bash
$ ls -a
.             ..            .m2           HelloWorld.kt

$ find .m2 | wc -l
  280004
```

Comapring the result. All commands are `opengrep scan -c $RULES ...`, where `$RULES` point to a local copy of the semgrep rule repo.

| Test | Before Fix | After Fix |
| --- | --- | --- |
| Baseline (`HelloWorld.kt`) | 17.912 | 16.534 |
| With flag (`--exclude .m2 .`) | 1:52.67 | 20.306 |
| With `.semgrepignore` file (`.`) | 3:30.92 | 22.228 |

### Problem

  When excluding large directories (e.g. `node_modules`, `build`), the Python frontend was slow and still reported tens/hundreds of thousands of skipped files in the scan summary. The root cause was that
  `Target.files_from_filesystem()` uses `path.glob("**/*")`, which expands the **entire** directory tree before any filtering occurs. `filter_excludes` and `FileIgnore.filter_paths` then pattern-match against every
   single collected file — O(N) in the number of files inside excluded directories.

  By contrast, `osemgrep` (`src/targeting/Find_targets.ml`) checks each directory against exclusion patterns **before descending**, making excluded directories O(1) regardless of their size.

  ### Solution

  Add directory-level early pruning to the Python filesystem walker, matching osemgrep's behavior, with identical scan semantics.

  #### `Target.files_from_filesystem_with_dir_pruning` (new method)

  Replaces `path.glob("**/*")` with `os.walk(topdown=True, followlinks=False)`. Before descending into each subdirectory, two checks are applied in-place on `dirnames`:

  1. **`--exclude` patterns** via `wcglob.globfilter` — patterns are passed in already preprocessed (e.g. `node_modules` → `**/node_modules`, `**/node_modules/**`) so the preprocessing cost is paid once per scan.
  Paths are expressed relative to the scan root (not absolute) so that `**/`-prefixed wcmatch patterns match correctly.

  2. **`.semgrepignore` patterns** via `FileIgnore._survives(dir / ".__check__")` — checking a virtual sentinel file *inside* the directory (rather than the directory path itself) correctly matches both `dir`
  patterns (via the generated `dir/**` fnmatch pattern) and `dir/` folder patterns. `_survives` is pure pattern-matching with no filesystem access, so this is safe with non-existent paths.

  #### `TargetManager.get_all_files_with_dir_pruning` (new method)

  Preprocesses exclude patterns once, then iterates over targets mirroring the **exact fallback chain** of `Target.files()` — substituting `files_from_filesystem_with_dir_pruning` wherever `files_from_filesystem`
  would have been called:

  | Scenario | Behaviour |
  |---|---|
  | Explicit file target | Delegates to `target.files()` unchanged |
  | Diff/baseline mode | Tries `files_from_git_diff()`; on failure falls through |
  | `ignore_baseline_handler=True` | Uses pruned walk directly |
  | Git-tracked-only (the default, since `respect_git_ignore = not no_git_ignore` and `no_git_ignore` defaults to `False`) | Tries `files_from_git_ls()`; on success uses that result; on failure falls through to
  pruned walk |
  | `--no-git-ignore` (pure filesystem) | Uses pruned walk directly |

  The critical fix for the default scan mode: previously when `files_from_git_ls()` failed (non-git directory or git unavailable), it silently fell back to the original `files_from_filesystem()` with no pruning.
  Now it falls back to `files_from_filesystem_with_dir_pruning`.

  Decorated with `@lru_cache` so the filesystem is walked at most once per unique `(exclude_patterns, ignore_baseline_handler, file_ignore)` combination.

  #### `TargetManager.get_files_for_language` (modified)

  Replaces `self.get_all_files(ignore_baseline_handler)` with a call to `get_all_files_with_dir_pruning`, passing:
  - `--exclude` patterns for the current product + `PATHS_ALWAYS_SKIPPED` (`.git`)
  - The product's `FileIgnore` object (if `respect_semgrepignore` is set)

  The existing `filter_excludes` and `filter_paths` calls remain as safety nets for file-level patterns (e.g. `--exclude "*.pyc"`) and for results from git-based traversals.

### DIFF
--- a/cli/src/semgrep/target_manager.py
+++ b/cli/src/semgrep/target_manager.py
@@ -504,6 +504,52 @@ class Target:
             if match.is_file() and not match.is_symlink()
         )
 
+    def files_from_filesystem_with_dir_pruning(
+        self,
+        preprocessed_patterns: Tuple[str, ...],
+        file_ignore: Optional[FileIgnore] = None,
+    ) -> FrozenSet[Path]:
+        """
+        Like files_from_filesystem but uses os.walk with directory-level pruning.
+        Receives already-preprocessed wcmatch patterns (via preprocess_path_patterns).
+        Optionally also accepts a FileIgnore to prune .semgrepignore-excluded dirs.
+        Directories matching either source of patterns are removed from dirnames
+        in-place so os.walk never descends into them.
+        """
+        if not preprocessed_patterns and file_ignore is None:
+            return self.files_from_filesystem()
+
+        result: Set[Path] = set()
+        for dirpath_str, dirnames, filenames in os.walk(
+            str(self.path), topdown=True, followlinks=False
+        ):
+            dirpath = Path(dirpath_str)
+            # Prune directories in-place — os.walk won't descend into removed entries.
+            # Use paths relative to self.path so that patterns like **/node_modules
+            # (which have no leading /) match correctly.
+            dirnames[:] = [
+                d
+                for d in dirnames
+                if not wcglob.globfilter(
+                    [str((dirpath / d).relative_to(self.path))],
+                    list(preprocessed_patterns),
+                    flags=wcglob.GLOBSTAR | wcglob.DOTGLOB,
+                )
+                # For .semgrepignore: check a virtual sentinel file inside the
+                # directory. _survives() is pure pattern-matching (no filesystem
+                # access) and checking `dir/__check__` correctly handles both `dir`
+                # patterns (matched via `dir/**`) and `dir/` folder patterns.
+                and (
+                    file_ignore is None
+                    or file_ignore._survives((dirpath / d / ".__check__").absolute())
+                )
+            ]
+            for filename in filenames:
+                filepath = dirpath / filename
+                if filepath.is_file() and not filepath.is_symlink():
+                    result.add(filepath)
+        return frozenset(result)
+
     @lru_cache(maxsize=None)
     def files(self, ignore_baseline_handler: bool = False) -> FrozenSet[Path]:
         """
@@ -750,6 +796,61 @@ class TargetManager:
         )
 
     @lru_cache(maxsize=None)
+    def get_all_files_with_dir_pruning(
+        self,
+        exclude_patterns: Tuple[str, ...],
+        ignore_baseline_handler: bool = False,
+        file_ignore: Optional[FileIgnore] = None,
+    ) -> FrozenSet[Path]:
+        """
+        Like get_all_files but performs directory-level pruning during filesystem
+        traversal for directory targets that use the filesystem scanner.
+
+        Mirrors Target.files() logic exactly, substituting
+        files_from_filesystem_with_dir_pruning wherever files_from_filesystem
+        would be called so that excluded directories are pruned during the walk
+        regardless of which fallback path leads to the filesystem scan.
+        """
+        preprocessed = tuple(
+            TargetManager.preprocess_path_patterns(list(exclude_patterns))
+        )
+        result: Set[Path] = set()
+        for target in self.targets:
+            if not target.path.is_dir():
+                # Explicit file target (or non-directory path) — no pruning needed.
+                result |= target.files(ignore_baseline_handler)
+                continue
+
+            # --- Directory target: replicate Target.files() fallback chain,
+            #     but replace files_from_filesystem() with the pruned variant. ---
+
+            if target.baseline_handler is not None:
+                if ignore_baseline_handler:
+                    # Scan all files ignoring the baseline (e.g. SCA lockfile scan).
+                    result |= target.files_from_filesystem_with_dir_pruning(preprocessed, file_ignore)
+                    continue
+                try:
+                    result |= target.files_from_git_diff()
+                    continue
+                except (subprocess.CalledProcessError, FileNotFoundError):
+                    logger.verbose(
+                        f"Unable to target only the changed files since baseline commit. Running on all git tracked files instead..."
+                    )
+
+            if target.git_tracked_only:
+                try:
+                    result |= target.files_from_git_ls()
+                    continue
+                except (subprocess.CalledProcessError, FileNotFoundError):
+                    logger.verbose(
+                        f"Unable to ignore files ignored by git ({target.path} is not a git directory or git is not installed). Running on all files instead..."
+                    )
+
+            result |= target.files_from_filesystem_with_dir_pruning(preprocessed, file_ignore)
+
+        return frozenset(result)
+
+    @lru_cache(maxsize=None)
     def get_files_for_language(
         self,
         lang: Union[None, Language, Literal["dependency_source_files"]],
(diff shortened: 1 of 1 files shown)


cli/src/semgrep/target_manager.py
```diff
@@ -749,6 +795,57 @@ def get_all_files(self, ignore_baseline_handler: bool = False) -> FrozenSet[Path
...
+            if target.baseline_handler is not None:
+                if ignore_baseline_handler:
+                    # Scan all files ignoring the baseline (e.g. SCA lockfile scan).
+                    result |= target.files_from_filesystem_with_dir_pruning(preprocessed, file_ignore)
+                    continue
+                try:
+                    result |= target.files_from_git_diff()
+                    continue
+                except (subprocess.CalledProcessError, FileNotFoundError):
+                    pass  # fall through to git_tracked_only / filesystem
```
dimitris-m: the original code had some logs here...

cli/src/semgrep/target_manager.py
```diff
@@ -749,6 +795,57 @@ def get_all_files(self, ignore_baseline_handler: bool = False) -> FrozenSet[Path
...
+
+            if target.git_tracked_only:
+                try:
+                    result |= target.files_from_git_ls()
+                    continue
+                except (subprocess.CalledProcessError, FileNotFoundError):
+                    pass  # fall through to pruned filesystem walk
```
dimitris-m: as before, original code has logs, why swallow exceptions?

cli/src/semgrep/target_manager.py
```diff
@@ -504,6 +504,52 @@ def files_from_filesystem(self) -> FrozenSet[Path]:
...
+        Optionally also accepts a FileIgnore to prune .semgrepignore-excluded dirs.
+        Directories matching either source of patterns are removed from dirnames
+        in-place so os.walk never descends into them.
+        """
+        if not preprocessed_patterns and file_ignore is None:
+            return self.files_from_filesystem()
+
+        result: Set[Path] = set()
+        for dirpath_str, dirnames, filenames in os.walk(
+            str(self.path), topdown=True, followlinks=False
```
dimitris-m, line 524: why not follow symlinks? 
what is the previous behaviour?

cli/src/semgrep/target_manager.py
```diff
@@ -504,6 +504,52 @@ def files_from_filesystem(self) -> FrozenSet[Path]:
...
+                # access) and checking `dir/__check__` correctly handles both `dir`
+                # patterns (matched via `dir/**`) and `dir/` folder patterns.
+                and (
+                    file_ignore is None
+                    or file_ignore._survives((dirpath / d / ".__check__").absolute())
+                )
+            ]
+            for filename in filenames:
+                filepath = dirpath / filename
+                if self._is_valid_file(filepath):
```
dimitris-m: why does it begin with _ ?

cli/src/semgrep/target_manager.py
```diff
@@ -504,6 +504,52 @@ def files_from_filesystem(self) -> FrozenSet[Path]:
...
+                    list(preprocessed_patterns),
+                    flags=wcglob.GLOBSTAR | wcglob.DOTGLOB,
+                )
+                # For .semgrepignore: check a virtual sentinel file inside the
+                # directory. _survives() is pure pattern-matching (no filesystem
+                # access) and checking `dir/__check__` correctly handles both `dir`
+                # patterns (matched via `dir/**`) and `dir/` folder patterns.
+                and (
+                    file_ignore is None
+                    or file_ignore._survives((dirpath / d / ".__check__").absolute())
```
dimitris-m, line 544: why does _survive start with _ ?

cli/src/semgrep/target_manager.py
maciejpirog: The previous behaviour is as defined in:

```python
    def files_from_filesystem(self) -> FrozenSet[Path]:
        return frozenset(
            match
            for match in self.path.glob("**/*")
            if match.is_file() and not match.is_symlink()
        )
```

it means that:
- symlinks, when we encounter them, are not returned as files.
- `Path.glob` does not follow simlinks by default.

The def of `path.glob` is:

```python
Path.glob(pattern, *, case_sensitive=None, recurse_symlinks=False)
```

(https://docs.python.org/3.13/library/pathlib.html#pathlib.Path.glob).

maciejpirog: Wait a minute. The same doc says:

> By default, or when the `recurse_symlinks` keyword-only argument is set to `False`, this method follows symlinks except when expanding `**` wildcards. Set `recurse_symlinks` to `True` to always follow symlinks.

It is not intuitive: we follow symlinks when `recurse_symlink` is set to `False`?? I will double-check this just in case

maciejpirog: Even more weird, in this discussion https://discuss.python.org/t/treating-symlinks-consistently-in-pathlib-path-glob/49233 the argument is called `follow_symlinks` + there is this `False`/`None` discrepancy.

But... This discussion https://stackoverflow.com/questions/46529760/getting-glob-to-follow-symlinks-in-python says that the default behaviour of `Path.glob` before adding the `follow_symlink` arguments was to follow symlinks **except for** when expanding `**` (which is our case). This explains the weird comment in the docs: when the new argument is not specified, use the weird old behaviour.

In other words, there is only one way to be 100% sure: manually check what the old behaviour is

maciejpirog: Test result: **We don't follow symlinks**. Example:

```bash
$ tree
.
├── a
│   └── a.py -> ../b/b.py
├── b
│   └── b.py
└── rule.yaml

$ opengrep scan -c rule.yaml b 2>/dev/null


┌────────────────┐
│ 1 Code Finding │
└────────────────┘

    b/b.py
    ❯❱ function_is_called
          a function is called

            1┆ foo()

$ opengrep scan -c rule.yaml a 2>/dev/null

$ opengrep scan -c rule.yaml . 2>/dev/null


┌────────────────┐
│ 1 Code Finding │
└────────────────┘

    b/b.py
    ❯❱ function_is_called
          a function is called

            1┆ foo()

Comment (maciejpirog) [reply cli/src/semgrep/target_manager.py:524]: ➕  the tricky case

```bash
opengrep scan -c rule.yaml --exclude b . 2>/dev/null
```

gives no results

cli/src/semgrep/target_manager.py
maciejpirog: I'll add logs

cli/src/semgrep/target_manager.py
maciejpirog: We swallow exceptions to use the backup behviour. If we get no files from `git diff`, we try `git ls-files`, and if it fails, we use all files.

cli/src/semgrep/target_manager.py
maciejpirog: In Python, underscore is used for private methods, see, e.g., https://www.datacamp.com/tutorial/python-private-methods-explained

cli/src/semgrep/target_manager.py
maciejpirog: private method, as with `_survives`

cli/src/semgrep/target_manager.py
maciejpirog: added a log instead of `pass`

cli/src/semgrep/target_manager.py
maciejpirog: added a log instead of `pass`

PR #597, opened 2026-02-18
Fix lambda taint tracking in intrafile mode

corneliuhoffman:
## Summary

Fix taint tracking through nested lambdas in intrafile mode. Clean up Ruby call disambiguation. Add support for functional interface invoke methods.

### Nested lambda taint tracking (all languages)

- **Function_id.ml**: For lambda names (prefixed `_tmp_lambda`), extract position from fake tokens to distinguish different lambdas that would otherwise collide on the same name.
- **AST_to_IL.ml**: Use `_tmp_lambda` prefix for lambda fresh variables (instead of generic `_tmp`).
- **Dataflow_tainting.ml**: Collect all lambdas recursively in innermost-first order so nested lambda signatures are available when processing their parents. Add signature lookup fallback via `Fun` shape in `lval_env` for variable-assigned lambdas.
- **Graph_from_AST.ml**: Update `_tmp` to `_tmp_lambda` to match AST_to_IL.

### Ruby cleanup: remove `ident_function_call_hack`

- **AST_to_IL.ml**: Remove the `ident_function_call_hack` and all associated dead code (`IdentSet`, `ctx`, `empty_ctx`, `add_entity_name`, `?ctx` parameter). This is now handled by `Disambiguate_ruby_calls` at the AST level.
- **Disambiguate_ruby_calls.ml**: Visit Call callees for all non-`N(Id(...))` cases (not just DotAccess), so unresolved method calls like `helper` in `helper.process()` get properly wrapped.
- **Match_tainting_mode.ml**: Remove `ctx` parameter from `check_fundef` and the visitor that populated `entity_names`.

### Ruby/Scala do-block flattening

- **AST_to_IL.ml**: Flatten `Call(Call(f, args), [Lambda])` into `Call(f, args @ [Lambda])` for Ruby. Pattern-matches exactly a singleton Lambda argument.
- **Graph_from_AST.ml**: Same flattening for HOF callback extraction, gated by `lang = Ruby || Scala`.

### Functional interface invoke methods

- **Lang_config.ml**: Add `invoke_methods` field. Configured for Java (`run`, `call`, `apply`, `accept`, `invoke`), Kotlin (`invoke`), C# (`Invoke`), Ruby (`call`).
- **Graph_from_AST.ml**: When a call like `r.run()` has no callee match and the method is in `invoke_methods`, add a call graph edge to the matching lambda in scope.
- **Dataflow_tainting.ml**: In the signature lookup fallback, strip invoke method offsets to find the base variable's `Fun` shape. Consolidate the hardcoded callback invoke method checks to use `Lang_config.invoke_methods`.

### Tests

- Nested lambda tests for Go (5 patterns + 1 negative), Java, JS, PHP, Python, Rust
- Invoke method tests for Java (`apply`, `accept`, `run`, `call`), Kotlin (`invoke`), C# (`Invoke`), Ruby (`call`)
- Ruby: chained method taint test, updated HOF callback tests

### DIFF
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -873,19 +873,6 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
           (Tok.unbracket inner_args @ Tok.unbracket outer_args)
       in
       expr_aux env ~void (G.Call (callee, merged_args) |> G.e)
-  (* Ruby: when the callee is a plain identifier (G.N), evaluate it via
-     `lval` instead of `expr` to skip the `ident_function_call_hack` (see
-     the G.N arm below, ~line 892). That hack wraps bare identifiers in a
-     0-arg Call for Ruby (where `foo` can mean `foo()`), but here we already
-     have an explicit G.Call — going through `expr` would produce a spurious
-     nested Call(Call(f, []), args) instead of Call(f, args). *)
-  | G.Call (({ G.e = G.N _; _ } as e), args) when env.lang =*= Lang.Ruby ->
-      let tok = G.fake "call" in
-      let ss_callee, callee_lval = lval env e in
-      let callee_exp = mk_e (Fetch callee_lval) (related_exp e) in
-      let ss_args, il_args = arguments env (Tok.unbracket args) in
-      let ss_call, call_exp = call_instr tok eorig ~void (fun res -> Call (res, callee_exp, il_args)) in
-      (ss_callee @ ss_args @ ss_call, call_exp)
   | G.Call (e, args) ->
       let tok = G.fake "call" in
       call_generic env ~void tok eorig e args
@@ -902,23 +889,7 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
   | G.ArrayAccess (_, _)
   | G.DeRef (_, _) ->
       let ss_lv, lval = lval env g_expr in
-      let exp = mk_e (Fetch lval) eorig in
-      let ident_function_call_hack ss exp =
-        (* Taking into account Ruby's ability to allow function calls without
-         * parameters or parentheses, we are conducting a check to determine
-         * if a function with the same name as the identifier exists, specifically
-         * for Ruby. *)
-        match lval with
-        | { base = Var { ident; id_info; _ }; _ }
-          when env.lang =*= Lang.Ruby
-               && Option.is_none !(id_info.id_resolved)
-               && IdentSet.mem (H.str_of_ident ident) env.ctx.entity_names ->
-            let tok = G.fake "call" in
-            let call_ss, call_exp = call_instr tok eorig ~void (fun res -> Call (res, exp, [])) in
-            (ss @ call_ss, call_exp)
-        | _ -> (ss, exp)
-      in
-      ident_function_call_hack ss_lv exp
+      (ss_lv, mk_e (Fetch lval) eorig)
   (* x = ClassName(args ...) in Python *)
   (* ClassName has been resolved to __init__ by the pro engine. *)
   (* Identified and treated as x = New ClassName(args ...) to support
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -858,6 +858,34 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
                  |> G.e) ]
       in
       call_generic env ~void tok eorig e (Tok.unsafe_fake_bracket arg_container)
+  (* Ruby do-block flattening: `f(args) do |x| ... end` is parsed as
+     Call(Call(f, args), [Lambda]) but the block is semantically an argument
+     to f, not to its return value. Flatten into Call(f, args @ [Lambda]). *)
+  | G.Call ({ e = G.Call (callee, inner_args); _ }, outer_args)
+    when env.lang =*= Lang.Ruby
+         && List.exists
+              (function
+               | G.Arg { G.e = G.Lambda _; _ } -> true
+               | _ -> false)
+              (Tok.unbracket outer_args) ->
+      let merged_args =
+        Tok.unsafe_fake_bracket
+          (Tok.unbracket inner_args @ Tok.unbracket outer_args)
+      in
+      expr_aux env ~void (G.Call (callee, merged_args) |> G.e)
+  (* Ruby: when the callee is a plain identifier (G.N), evaluate it via
+     `lval` instead of `expr` to skip the `ident_function_call_hack` (see
+     the G.N arm below, ~line 892). That hack wraps bare identifiers in a
+     0-arg Call for Ruby (where `foo` can mean `foo()`), but here we already
+     have an explicit G.Call — going through `expr` would produce a spurious
+     nested Call(Call(f, []), args) instead of Call(f, args). *)
+  | G.Call (({ G.e = G.N _; _ } as e), args) when env.lang =*= Lang.Ruby ->
+      let tok = G.fake "call" in
+      let ss_callee, callee_lval = lval env e in
+      let callee_exp = mk_e (Fetch callee_lval) (related_exp e) in
+      let ss_args, il_args = arguments env (Tok.unbracket args) in
+      let ss_call, call_exp = call_instr tok eorig ~void (fun res -> Call (res, callee_exp, il_args)) in
+      (ss_callee @ ss_args @ ss_call, call_exp)
   | G.Call (e, args) ->
       let tok = G.fake "call" in
       call_generic env ~void tok eorig e args
@@ -988,7 +1016,7 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
   | G.Comprehension (_op, (_l, (er, clauses), _r)) ->
       comprehension env er clauses
   | G.Lambda fdef ->
-      let lval = fresh_lval (snd fdef.fkind) in
+      let lval = fresh_lval ~str:"_tmp_lambda" (snd fdef.fkind) in
       let final_fdef =
         (* NOTE: Reset control-flow labels so that break/continue/recur from
          * the enclosing scope don't bleed into the lambda body. *)
@@ -1408,7 +1436,7 @@ and record env ((_tok, origfields, _) as record_def) : stmts * exp =
               (* Some languages such as javascript allow function
                  definitions in object literal syntax. *)
               | G.FuncDef fdef ->
-                  let lval = fresh_lval (snd fdef.fkind) in
+                  let lval = fresh_lval ~str:"_tmp_lambda" (snd fdef.fkind) in
                   (* See NOTE about resetting control-flow labels for lambdas. *)
                   let fdef =
                     function_definition
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -852,16 +852,12 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
   (* Ruby do-block flattening: `f(args) do |x| ... end` is parsed as
      Call(Call(f, args), [Lambda]) but the block is semantically an argument
      to f, not to its return value. Flatten into Call(f, args @ [Lambda]). *)
-  | G.Call ({ e = G.Call (callee, inner_args); _ }, outer_args)
-    when env.lang =*= Lang.Ruby
-         && List.exists
-              (function
-               | G.Arg { G.e = G.Lambda _; _ } -> true
-               | _ -> false)
-              (Tok.unbracket outer_args) ->
+  | G.Call ({ e = G.Call (callee, inner_args); _ },
+            (_, ([ G.Arg { G.e = G.Lambda _; _ } ] as outer_arg), _ ))
+    when env.lang =*= Lang.Ruby ->
       let merged_args =
         Tok.unsafe_fake_bracket
-          (Tok.unbracket inner_args @ Tok.unbracket outer_args)
+          (Tok.unbracket inner_args @ outer_arg)
       in
       expr_aux env ~void (G.Call (callee, merged_args) |> G.e)
   | G.Call (e, args) ->
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -45,9 +45,6 @@ let log_error ?tok msg : unit = Log.err (fun m -> m "%s" (locate ?tok msg))
 (*****************************************************************************)
 (* Types *)
 (*****************************************************************************)
-module IdentSet = Set.Make (String)
-
-type ctx = { entity_names : IdentSet.t }
 type stmts = stmt list
 
 type rec_point_lvals =
@@ -63,18 +60,14 @@ type env = {
   break_labels : label list;
   cont_label : label option;
   rec_point_label : label option;
-  ctx : ctx;
   rec_point_lvals : rec_point_lvals option;
   inside_function : bool;
 }
-
-let empty_ctx : ctx = { entity_names = IdentSet.empty }
 
 let empty_env (lang : Lang.t) : env =
   { break_labels = [];
     cont_label = None;
     rec_point_label = None;
-    ctx = empty_ctx;
     rec_point_lvals = None;
     inside_function = false;
     lang }
@@ -257,8 +250,6 @@ let mk_class_constructor_name (ty : G.type_) cons_id_info : G.name option =
       Some (G.Id (id, cons_id_info))
   | __else__ -> None
 
-let add_entity_name ctx ident : ctx =
-  { entity_names = IdentSet.add (H.str_of_ident ident) ctx.entity_names }
 
 let def_expr_evaluates_to_value (lang : Lang.t) : bool =
   match lang with
@@ -2614,8 +2605,8 @@ and function_definition env fdef : function_definition =
 (* Entry points *)
 (****************************************************************************)
 
-let function_definition lang ?ctx fdef : function_definition =
-  let env = { (empty_env lang) with ctx = ctx ||| empty_ctx } in
+let function_definition lang fdef : function_definition =
+  let env = empty_env lang in
   function_definition env fdef
 
 let stmt lang st : stmts =
--- a/src/call_graph/Function_id.ml
+++ b/src/call_graph/Function_id.ml
@@ -20,8 +20,21 @@ let normalize_file (file : Fpath.t) : string =
   Fpath.to_string (Fpath.normalize file)
 
 let key ((id, tok) : t) =
+  (* For lambda names (starting with "_tmp_lambda"), extract position even from
+   * fake tokens that have position info. This is important for distinguishing
+   * different lambdas that would otherwise all collide on the same name.
+   * For regular functions with fake tokens, use empty key
+   * to preserve the original matching behavior. *)
+  let is_lambda_name = String.starts_with ~prefix:"_tmp_lambda" id in
   if Tok.is_fake tok then
-    (id, "", 0, 0)
+    if is_lambda_name then
+      match Tok.loc_of_tok tok with
+      | Ok loc ->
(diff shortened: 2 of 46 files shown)


src/call_graph/Function_id.ml
```diff
@@ -20,8 +20,25 @@ let normalize_file (file : Fpath.t) : string =
   Fpath.to_string (Fpath.normalize file)
 
 let key ((id, tok) : t) =
+  (* For Go lambda names (starting with "_tmp"), extract position even from fake
```
dimitris-m: But his is not Go only right?

src/call_graph/Function_id.ml
```diff
@@ -20,8 +20,25 @@ let normalize_file (file : Fpath.t) : string =
...
+   * tokens that have position info. This is important for distinguishing
+   * different lambdas that all have "_tmp" names.
+   * For other languages or regular functions with fake tokens, use empty key
+   * to preserve the original matching behavior. *)
+  let is_lambda_name = String.length id >= 4 && String.sub id 0 4 = "_tmp" in
+  let is_go_file loc =
```
dimitris-m: This is a more general problem right? Does not only apply to Go or am I missing something?

src/call_graph/Function_id.ml
```diff
@@ -20,8 +20,25 @@ let normalize_file (file : Fpath.t) : string =
   Fpath.to_string (Fpath.normalize file)
 
 let key ((id, tok) : t) =
+  (* For Go lambda names (starting with "_tmp"), extract position even from fake
+   * tokens that have position info. This is important for distinguishing
+   * different lambdas that all have "_tmp" names.
+   * For other languages or regular functions with fake tokens, use empty key
+   * to preserve the original matching behavior. *)
+  let is_lambda_name = String.length id >= 4 && String.sub id 0 4 = "_tmp" in
```
dimitris-m: `String.starts_with ~prefix:"_tmp" id`

src/tainting/Dataflow_tainting.ml
```diff
@@ -792,6 +792,17 @@ let lambdas_to_analyze_in_node env lambdas node =
...
 
+(* Collect ALL lambdas recursively from a fun_cfg, in innermost-first order.
+   This ensures nested lambda signatures are extracted before their parents. *)
+let rec collect_all_lambdas_innermost_first (fun_cfg : IL.fun_cfg)
+    : (IL.name * IL.fun_cfg) list =
+  IL.NameMap.fold (fun name lcfg results ->
+    (* First collect nested lambdas from this lambda *)
+    let nested = collect_all_lambdas_innermost_first lcfg in
+    (* Then add this lambda after its nested ones *)
+    results @ nested @ [(name, lcfg)]
```
dimitris-m, line 784: nit: `List.rev_append` ?

of course these are small lists so ok

tests/rules/cross_function_tainting/test_lambda_deeply_nested.yaml
dimitris-m, line 1: These are not really intrafile tests, this is why they pass when the rules do not have: `options: taint_intrafile: true.`

Should we move to another folder? Or add also true intrafile tests?

dimitris-m (commented):
Left some comments that I think need some clarification.

Also I recommend to have more informative commit messages.

src/call_graph/Function_id.ml
corneliuhoffman: I am not sure which other langs it applies to. It surely does not apply to ruby

tests/rules/cross_function_tainting/test_lambda_deeply_nested.yaml
corneliuhoffman: they do not pass without intrafile

corneliuhoffman: tests in this folder are always run with intrafile

src/tainting/Dataflow_tainting.ml
corneliuhoffman: I think these are fine past 5.1

src/call_graph/Function_id.ml
corneliuhoffman: fixd, added some more langs tests as well as a change to AST_to_IL

src/call_graph/Function_id.ml
corneliuhoffman: fixd, added some more langs as well as a change to AST_to_IL

src/call_graph/Function_id.ml
corneliuhoffman: changed

src/call_graph/dune
```diff
@@ -7,6 +7,7 @@
   semgrep_core
   semgrep.il
   semgrep.matching
+  semgrep.rule
```
dimitris-m: Is that used?

src/analyzing/AST_to_IL.ml
```diff
@@ -853,9 +845,36 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
...
+      let ss_callee, callee_exp =
+        match e.G.e with
+        | G.N _ when env.lang =*= Lang.Ruby ->
+            let ss, l = lval env e in
+            (ss, mk_e (Fetch l) (related_exp e))
+        | _ -> expr env e
+      in
+      let ss_args, il_args = arguments env (Tok.unbracket args) in
+      let ss_call, call_exp = call_instr tok eorig ~void (fun res -> Call (res, callee_exp, il_args)) in
+      (ss_callee @ ss_args @ ss_call, call_exp)
```
dimitris-m: But this is not gated for lang = Ruby, and what happened to the generic G.Call case that was here?

dimitris-m: This part seems mostly the same as call_generic, still not sure about the full changes -- would be clearer if we split the Ruby and non-Ruby cases

src/analyzing/AST_to_IL.ml
```diff
@@ -123,10 +123,9 @@ let fixme_stmt kind gany : stmts =
 
 let fresh_var ?(str = "_tmp") tok : name =
   let tok =
-    (* We don't want "fake" auxiliary variables to have non-fake tokens, otherwise
-       we confuse ourselves! E.g. during taint-tracking we don't want to add these
-       variables to the taint trace. *)
-    if Tok.is_fake tok then tok else Tok.fake_tok tok str
+    (* We always create a fake token to avoid confusing taint-tracking,
+       but preserve position info for distinguishing different lambdas. *)
+    Tok.fake_tok tok str
```
dimitris-m: What changed here? The token was already guaranteed to be fake no?

src/call_graph/Function_id.ml
```diff
@@ -20,8 +20,21 @@ let normalize_file (file : Fpath.t) : string =
   Fpath.to_string (Fpath.normalize file)
 
 let key ((id, tok) : t) =
+  (* For lambda names (starting with "_tmp"), extract position even from fake
+   * tokens that have position info. This is important for distinguishing
+   * different lambdas that all have "_tmp" names.
+   * For regular functions with fake tokens, use empty key
+   * to preserve the original matching behavior. *)
+  let is_lambda_name = String.starts_with ~prefix:"_tmp" id in
```
dimitris-m: But does _tmp imply lambda?
I was under the impression that it's only the other way around.

src/analyzing/AST_to_IL.ml
```diff
@@ -123,10 +123,9 @@ let fixme_stmt kind gany : stmts =
 
 let fresh_var ?(str = "_tmp") tok : name =
   let tok =
-    (* We don't want "fake" auxiliary variables to have non-fake tokens, otherwise
-       we confuse ourselves! E.g. during taint-tracking we don't want to add these
-       variables to the taint trace. *)
-    if Tok.is_fake tok then tok else Tok.fake_tok tok str
+    (* We always create a fake token to avoid confusing taint-tracking,
+       but preserve position info for distinguishing different lambdas. *)
```
dimitris-m: the original comment was more informative... eg "during taint-tracking we don't want to add these ..."

src/tainting/Graph_from_AST.ml
```diff
@@ -524,6 +464,8 @@ let extract_toplevel_calls ~(lang : Lang.t) ?(object_mappings = []) ?(all_funcs
   v#visit_program () ast;
   !calls |> dedup_fn_ids
 
+(* Detect if a function is a user-defined HOF by checking if it calls any of its parameters.
+   Returns a list of (parameter_name, parameter_index) for called parameters. *)
 (* Helper to extract callback name from an argument expression.
```
dimitris-m: 2 comments in a row here, also not sure how the new comment relates to the function (detect vs extract).

src/tainting/Dataflow_tainting.ml
```diff
@@ -2303,20 +2335,10 @@ let check_tainted_instr env instr : Taints.t * S.shape * Lval_env.t =
                       (* Check if this is a call to a function parameter (either direct or via method) *)
                       (match e_obj with
                       | `Obj (_obj_taints, S.Arg _fun_arg) ->
-                          (* This is a method call on a function parameter (e.g., callback.apply in Java).
-                           * Treat it as invoking the callback.
-                           * EXCEPTION: Ruby's .call method should NOT be treated this way during signature
-                           * extraction, as it creates infinite recursion. Ruby blocks are handled via
-                           * implicit lambda detection instead. *)
-                          let is_ruby_call_method =
```
dimitris-m, line 2311: How is infinite recursion avoided now? Since this comment was here, seems important to state somewhere why this is no longer the case.

src/call_graph/Function_id.ml
```diff
@@ -65,5 +76,8 @@ let of_il_name (n : IL.name) : t =
   n.IL.ident
 
 let to_file_line_col ((_, tok) : t) : string * int * int =
-  if Tok.is_fake tok then ("unknown", 0, 0)
+  if Tok.is_fake tok then
+    match Tok.loc_of_tok tok with
+    | Ok loc -> (normalize_file loc.Tok.pos.file, loc.Tok.pos.line, loc.Tok.pos.column)
+    | _ -> ("unknown", 0, 0)
```
dimitris-m, line 87: no need for `is_lambda_name`?

src/call_graph/dune
corneliuhoffman: Good catch — leftover from the earlier Go-specific version. Removed.

src/analyzing/AST_to_IL.ml
corneliuhoffman: You're right — reverted.

src/analyzing/AST_to_IL.ml
corneliuhoffman: split and explained

corneliuhoffman: split and explained

src/analyzing/AST_to_IL.ml
corneliuhoffman: reverted

src/call_graph/Function_id.ml
corneliuhoffman: Good point, made it more precise in AST_to_IL and changed it in other places in consequence.

src/tainting/Dataflow_tainting.ml
corneliuhoffman: This is dead code, Cannot remember why it was added but nevertheless it is in the "non intrafile" branch> I tried all kind of experiments (including gitlab) and found no issue.

src/call_graph/Function_id.ml
corneliuhoffman: commented

src/tainting/Graph_from_AST.ml
corneliuhoffman: yes first comment was an orphan. Removed

src/analyzing/AST_to_IL.ml
```diff
@@ -853,6 +853,34 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
...
+               | _ -> false)
+              (Tok.unbracket outer_args) ->
+      let merged_args =
+        Tok.unsafe_fake_bracket
+          (Tok.unbracket inner_args @ Tok.unbracket outer_args)
+      in
+      expr_aux env ~void (G.Call (callee, merged_args) |> G.e)
+  (* Ruby: when the callee is a plain identifier (G.N), evaluate it via
+     `lval` instead of `expr` to skip the `ident_function_call_hack` (see
+     the G.N arm below, ~line 892). That hack wraps bare identifiers in a
```
dimitris-m: But how relevant is that check referred to after PR #626 and #651 ?
Maybe we should cleanup instead of patching.

src/analyzing/AST_to_IL.ml
```diff
@@ -853,6 +853,34 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
...
+               | G.Arg { G.e = G.Lambda _; _ } -> true
+               | _ -> false)
+              (Tok.unbracket outer_args) ->
+      let merged_args =
+        Tok.unsafe_fake_bracket
+          (Tok.unbracket inner_args @ Tok.unbracket outer_args)
+      in
+      expr_aux env ~void (G.Call (callee, merged_args) |> G.e)
+  (* Ruby: when the callee is a plain identifier (G.N), evaluate it via
+     `lval` instead of `expr` to skip the `ident_function_call_hack` (see
```
dimitris-m: Is this relevant, we already approximate Ruby's method of choosing if something is a method call or not so...?

src/analyzing/AST_to_IL.ml
corneliuhoffman: clened up

src/analyzing/AST_to_IL.ml
corneliuhoffman: cleaned up

src/tainting/Graph_from_AST.ml
```diff
@@ -697,6 +697,16 @@ let extract_hof_callbacks ?(_object_mappings = []) ?(all_funcs = [])
...
+        (* Ruby/Scala block pattern: f(args) { block } is Call(Call(callee, inner_args), [block]).
+           Merge inner_args and block args so the HOF detection sees all arguments together. *)
+        | G.Call ({ e = G.Call (callee, inner_args); _ }, outer_args) ->
+            let merged_args = Tok.unsafe_fake_bracket
+              (Tok.unbracket inner_args @ Tok.unbracket outer_args) in
+            let found = extract_hof_callbacks_from_call
+              ~method_hofs ~function_hofs ~all_funcs ~caller_parent_path
+              callee merged_args
+            in
+            callbacks := found @ !callbacks
```
dimitris-m, line 733: Should this not be gated by language?

src/parsing/Disambiguate_ruby_calls.ml
```diff
@@ -23,8 +23,16 @@ class ['self] visitor =
...
+      (* Do not recurse into the direct callee of a Call, but DO visit
+         the receiver of a DotAccess callee — in `helper.process()`,
+         `helper` may be an unresolved method call that needs wrapping. *)
       | Call (callee, args) ->
+          let callee = match callee.e with
+            | DotAccess (receiver, tok, field) ->
+                let receiver = self#visit_expr env receiver in
+                { callee with e = DotAccess (receiver, tok, field) }
+            | _ -> callee
+          in
```
dimitris-m: A more robust fix would be to visit the callee for all non-N(Id(...)) cases, not just DotAccess, agree?

src/analyzing/AST_to_IL.ml
```diff
@@ -869,23 +884,7 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
...
-        | { base = Var { ident; id_info; _ }; _ }
-          when env.lang =*= Lang.Ruby
-               && Option.is_none !(id_info.id_resolved)
-               && IdentSet.mem (H.str_of_ident ident) env.ctx.entity_names ->
-            let tok = G.fake "call" in
-            let call_ss, call_exp = call_instr tok eorig ~void (fun res -> Call (res, exp, [])) in
-            (ss @ call_ss, call_exp)
-        | _ -> (ss, exp)
-      in
-      ident_function_call_hack ss_lv exp
```
dimitris-m, line 893: Good catch.

From the commit where this was introduced, in src/engine/Match_tainting_mode.ml we also have: 

```ocaml
      (* FIXME: This is no longer needed, now we can just check the type 'n'. *)
      let ctx = ref AST_to_IL.empty_ctx in
      Visit_function_defs.visit
        (fun opt_ent _fdef ->
          match opt_ent with
          | Some { name = EN (Id (n, _)); _ } ->
              ctx := AST_to_IL.add_entity_name !ctx n
          | __else__ -> ())
        ast;
```

This is probably also not needed, is that the case?

src/tainting/Dataflow_tainting.ml
```diff
@@ -1527,7 +1538,27 @@ let check_function_call env fun_exp args
...
+    if env.taint_inst.options.taint_intrafile then
+      let from_db = lookup_signature env fun_exp arity in
+      match from_db with
+      | Some _ -> from_db
+      | None ->
+          (* lookup_signature failed - check if callee has a Fun shape in lval_env.
+           * This handles the case where a lambda is assigned to a variable like:
+           *   callback := func(x) { sink(x) }
+           *   callback(source())
+           * The signature is stored under the lambda's internal name (_tmp:N),
```
dimitris-m: This comment is not longer accurate right? It would be `_tmp_lambda`.

dimitris-m (changes requested):
- Added some comments
- The PR description seems stale

src/analyzing/AST_to_IL.ml
corneliuhoffman: indeed removing lots now

src/tainting/Dataflow_tainting.ml
corneliuhoffman: fixed

src/parsing/Disambiguate_ruby_calls.ml
corneliuhoffman: added

src/tainting/Graph_from_AST.ml
corneliuhoffman: Done — gated with `when Lang.(lang =*= Ruby || lang =*= Scala)`.

src/parsing/Disambiguate_ruby_calls.ml
corneliuhoffman: Done — now visits the callee for all non-`N(Id(...))` cases.

src/analyzing/AST_to_IL.ml
corneliuhoffman: Yes — removed `ctx`, `empty_ctx`, `add_entity_name`, `IdentSet` from AST_to_IL, the `?ctx` param from `function_definition` and `cfg_of_gfdef`, and the visitor that populated it in Match_tainting_mode.ml.

src/tainting/Graph_from_AST.ml
dimitris-m: Do I remember correctly that kotlin has something like that?

src/analyzing/AST_to_IL.ml
```diff
@@ -853,6 +844,21 @@ and expr_aux env ?(void = false) g_expr : stmts * exp =
...
+         && List.exists
+              (function
+               | G.Arg { G.e = G.Lambda _; _ } -> true
+               | _ -> false)
+              (Tok.unbracket outer_args) ->
+      let merged_args =
+        Tok.unsafe_fake_bracket
+          (Tok.unbracket inner_args @ Tok.unbracket outer_args)
+      in
+      expr_aux env ~void (G.Call (callee, merged_args) |> G.e)
```
dimitris-m: But in that case, is outer_args exactly one argument matching the above pattern?

dimitris-m: I fixed that

dimitris-m (approved):
LGTM

PR #614, opened 2026-03-16
Elixir: Multi-clause functions and interpolated strings

maciejpirog:
This PR fixes two things:

## Multi-clause functions

Functions can be defined via multiple clauses, like:

```elixir
def factorial(0) do
    1
end
def factorial(n) do
    n * factorial(n - 1)
end
```

Before: these were understood as two separate functions with the same name (+ `when` clauses were not supported).
After: they are the same function (+`when` clauses work now).

Note that the mechanism works slightly differently than in Clojure, because in Elixir two functions with the same name but different arities are actually two functions. For example, the following code defines **two** functions:

```elixir
def foo(0), do: something()
def foo(x), do: something()
def foo(0, y), do: something()
def foo(1, y), do: something()
def foo(x, y), do: something()
```

## Interpolated strings

They didn't work before, they work now


languages/elixir/ast/Elixir_to_elixir.ml
```diff
@@ -101,11 +102,49 @@ class ['self] visitor =
...
+              f_body = (tdo, body, tend);
+              f_is_private = String.equal def_str "defp";
+            }
+          in
+          S (D (FuncDef [def]))
+      (* def foo(x) when guard, do: body *)
+      | ( I (Id (( "def" | "defp" ) as def_str, tdef)),
+          (_, ([ When (Call (I ident, args, None), _twhen, E guard) ],
+               [ Kw_expr ((X1 (do_kw, _), _tok_colon), body) ]), _),
+          None) when String.starts_with ~prefix:"do:" do_kw ->
```
dimitris-m, line 125: Why not `String.equal`?

languages/elixir/ast/Elixir_to_elixir.ml
```diff
@@ -101,11 +102,49 @@ class ['self] visitor =
...
+            {
+              f_def = tdef;
+              f_name = ident;
+              f_params = params;
+              f_guard = Some guard;
+              f_body = Tok.unsafe_fake_bracket [ body ];
+              f_is_private = String.equal def_str "defp";
+            }
+          in
+          S (D (FuncDef [def]))
```
dimitris-m: Maybe use an aux function to avoid duplication with the case above?

languages/elixir/ast/Elixir_to_elixir.ml
```diff
@@ -124,14 +163,69 @@ class ['self] visitor =
...
+(* Group consecutive FuncDef stmts with same key into a single FuncDef *)
+let rec group_func_clauses (stmts : stmts) : stmts =
+  match stmts with
+  | [] -> []
+  | S (D (FuncDef [ clause ])) :: rest ->
+      (match func_clause_key clause with
+       | None -> S (D (FuncDef [ clause ])) :: group_func_clauses rest
+       | Some key ->
+           let rec take_matching acc = function
+             | S (D (FuncDef [ c ])) :: tl when func_clause_key c = Some key ->
```
dimitris-m: polymorphic equality which I guess is unavoidable as long as the key is defined as is, but maybe if it was a record with derived equality ....

languages/elixir/ast/Elixir_to_elixir.ml
maciejpirog: For some reason parser (always? sometimes? not sure) captures the trailing whitespaces. To be fixed, but not in this PR

languages/elixir/ast/Elixir_to_elixir.ml
maciejpirog: Done

languages/elixir/ast/Elixir_to_elixir.ml
maciejpirog: I introduced a separate type with (hand-defined, but not polymorphic) equality

PR #626, opened 2026-03-27
Fix(Ruby): zero argument method call syntax

corneliuhoffman:
### Problem

In Ruby, method calls with no arguments don't require parentheses. `a.method` and `a.method()` are equivalent, but the parser produced different ASTs: `DotAccess(a, method)` vs `Call(DotAccess(a, method), [])`.

Similarly, a bare identifier like `source` (no parens, no receiver) is a method call when no prior assignment exists in scope, but tree-sitter parses it as a plain identifier.

### Solution

**Dot-access calls (`a.method`):** In the `lhs` handler of `Parse_ruby_tree_sitter.ml`, wrap the `DotAccess` produced by `call_` in `Call(DotAccess(...), [], None)`. This is the single call site where `call_` appears standalone (without a parent `call` that adds arguments).

**Bare identifiers (`source`):** A post-naming map visitor (`Disambiguate_ruby_calls.ml`) rewrites unresolved lowercase `N(Id(name))` to `Call(N(Id(name)), [])`, matching Ruby's own disambiguation rule: if no assignment to that name has been seen in scope, it's a method call.

### Note

The pattern `$OBJ.$METHOD` now only matches `a.method` or `a.method()` (zero-arg calls). To match `a.method(arg)`, use `$OBJ.$METHOD(...)`.

### TODO

In Ruby there are no standalone functions -- every call is a method call on `self` ([reference](https://en.wikipedia.org/wiki/Ruby_(programming_language))). Currently `source()` parses as `Call(N(Id("source")), [])` rather than the semantically correct `Call(DotAccess(IdSpecial(Self), source), [])`. A follow-up issue should fix the translation for all receiverless Ruby calls to include the implicit `self` receiver.

Specifically: "_Every function is a method and methods are always called on an object. Methods defined at the top level scope become methods of the Object class._"

### Related

For the todoruleid: PR #628

### DIFF
--- a/tests/rules/cross_function_tainting/test_ruby_zero_arg_dispatch.rb
+++ b/tests/rules/cross_function_tainting/test_ruby_zero_arg_dispatch.rb
@@ -50,6 +50,48 @@ def test_new_no_parens_get_no_parens
   sink(result)
 end
 
+# Bare identifier disambiguation: `source` (no parens) is a method call
+# when no prior assignment exists, and should be matched by `source(...)`.
+
+class ConverterBare
+  def initialize(data)
+    @data = data
+  end
+
+  def taint_method
+    return source
+  end
+
+  def get_data
+    @data
+  end
+
+  def use_bare_source_method
+    # ruleid: test-ruby-zero-arg-dispatch
+    sink(self.taint_method())
+  end
+end
+
+def test_bare_source_as_arg
+  obj = Converter.new(source)
+  result = obj.get_data()
+  # ruleid: test-ruby-zero-arg-dispatch
+  sink(result)
+end
+
+def test_bare_source_not_variable
+  x = source
+  # ruleid: test-ruby-zero-arg-dispatch
+  sink(x)
+end
+
+def test_bare_source_is_variable
+  source = "safe"
+  x = source
+  # ok: test-ruby-zero-arg-dispatch
+  sink(x)
+end
+
 class TaintedService
   def initialize
     @data = source
--- a/src/parsing/Disambiguate_ruby_calls.ml
+++ b/src/parsing/Disambiguate_ruby_calls.ml
@@ -1,0 +1,39 @@
+open AST_generic
+
+(* A bare lowercase identifier that naming could not resolve is a method call
+   in Ruby.  Rewrite [N(Id(name, info))] to [Call(N(Id(name, info)), [])]
+   so that downstream analyses (taint, matching) treat it the same as an
+   explicit [source()] call. *)
+
+(* A bare identifier is a method call candidate only when:
+   - naming left it unresolved (not a local variable or parameter)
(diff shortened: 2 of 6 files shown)


tests/rules/cross_function_tainting/test_ruby_zero_arg_dispatch.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - id: test-ruby-zero-arg-dispatch
+    message: Taint through zero-arg method call without parens
+    languages: [ruby]
+    severity: ERROR
+    mode: taint
+    pattern-sources:
+      - pattern: source
```
dimitris-m: but source will make any instance of source match, why not make it a function call?

src/engine/tests/Unit_engine.ml
```diff
@@ -752,6 +752,9 @@ let semgrep_rules_repo_tests () : Testo.t list =
              | s when s =~ ".*/semgrep-rules/clojure/lang/security/command-injection-shell-call.yaml" -> None
              (* FIXME: This fails, unblocking to create a binary for testing. *)
              | s when s =~ ".*/semgrep-rules/clojure/lang/security/documentbuilderfactory-xxe.yaml" -> None
+             (* FIXME: The rule is too restrictive, the patterns should contain a (...)> see the comment in PR
```
dimitris-m: seems there's an erroneous `>` in there?

tests/rules/cross_function_tainting/test_ruby_zero_arg_dispatch.rb
```diff
@@ -0,0 +1,75 @@
...
+  obj = TaintedService.new()
+  result = obj.get_data()
+  # todoruleid: test-ruby-zero-arg-dispatch
+  sink(result)
+end
+
+def test_zero_arg_new_no_parens
+  obj = TaintedService.new
+  result = obj.get_data()
+  # todoruleid: test-ruby-zero-arg-dispatch
```
dimitris-m, line 115: It is not clear why we have `todoruleid` here, could we add a comment somewhere to explain this ?

tests/rules/cross_function_tainting/test_ruby_zero_arg_dispatch.yaml
dimitris-m: Fixed

src/engine/tests/Unit_engine.ml
dimitris-m: Fixed

tests/rules/cross_function_tainting/test_ruby_zero_arg_dispatch.rb
dimitris-m: PR #628

PR #628, opened 2026-03-30
fix: proper constructor taint analysis

corneliuhoffman:
## Issue

Constructor calls (new Foo(), Foo(), Foo.new()) were not properly
  analyzed by the taint engine. Taint only flowed through constructors
  accidentally via all_args_taints leaking, causing both false negatives
  (zero-arg constructors with internal sources missed) and false positives
  (constructors ignoring tainted args still propagating taint).
## Solution 

 The fix connects constructor calls to constructor methods in the call
  graph so their taint signatures are extracted and available. At call
  sites, the constructor callee expression is remapped from ClassName()
  to obj.ClassName() so that Sig_inst can instantiate BThis field effects
  (like this.data = tainted_arg) onto the target variable. The constructor
  shape is then read back from lval_env so it propagates through
  intermediate assignments to the final object variable.
  
  When a method is called on a parameter whose class is unknown from
  object_mappings (e.g. a function argument, not a local variable),
  the call graph now falls back to the parameter's type annotation
  (e.g. Python's `def f(x: ClassName)`) to resolve the method to the
  correct class. This enables cross-function taint tracking through
  typed function boundaries.

  ### Changes:
  - Graph_from_AST: add G.New to extract_calls so constructor calls
    create call graph edges; resolve ClassName() and ClassName.new()
    to the constructor method using Lang_config; use SID-aware
    object_mappings lookup to disambiguate same-named variables;
    fall back to type annotations on parameters (e.g. Python's
    `def f(x: ClassName)`) to resolve method calls when
    object_mappings has no entry
  - AST_to_IL: remove id_resolved guard from mk_class_constructor_name
    so the IL New instruction always carries a constructor reference
  - Dataflow_tainting: use eorig tokens for call graph lookup in
    get_signature_for_object (fixes token mismatch); remap constructor
    callees to obj.Constructor so BThis maps correctly; prepend self
    arg for Python constructors; read back shape from lval_env after
    constructor ToLval effects; resolve obj.method() in ToSinkInCall

### DIFF
--- a/tests/rules/cross_function_tainting/test_constructor_taint_bugs.rb
+++ b/tests/rules/cross_function_tainting/test_constructor_taint_bugs.rb
@@ -1,0 +1,38 @@
+# FALSE NEGATIVE: zero-arg constructor with internal source.
+# The taint engine never analyzes initialize, so @data is not tainted.
+class InternalSource
+  def initialize
+    @data = source()
+  end
+
+  def get_data
+    @data
+  end
+end
+
+def test_false_negative
+  obj = InternalSource.new()
+  result = obj.get_data()
+  # ruleid: constructor-taint-bugs
+  sink(result)
+end
+
+# FALSE POSITIVE: constructor ignores its argument.
+# The taint engine leaks source() through all_args_taints
+# even though initialize never stores it.
+class IgnoresArg
+  def initialize(data)
+    @data = "safe"
+  end
+
+  def get_data
+    @data
+  end
+end
+
+def test_false_positive
+  obj = IgnoresArg.new(source())
+  result = obj.get_data()
+  # ok: constructor-taint-bugs
+  sink(result)
+end
--- a/tests/rules/cross_function_tainting/test_constructor_taint_bugs.yaml
+++ b/tests/rules/cross_function_tainting/test_constructor_taint_bugs.yaml
@@ -1,0 +1,10 @@
+rules:
+  - id: constructor-taint-bugs
+    message: Taint through constructor
+    languages: [ruby]
+    severity: ERROR
+    mode: taint
+    pattern-sources:
+      - pattern: source()
+    pattern-sinks:
+      - pattern: sink(...)
--- a/tests/rules/cross_function_tainting/test_constructor_taint_bugs_java.java
+++ b/tests/rules/cross_function_tainting/test_constructor_taint_bugs_java.java
@@ -1,0 +1,41 @@
+// FALSE NEGATIVE: zero-arg constructor with internal source.
+class InternalSource {
+    String data;
+
+    InternalSource() {
+        this.data = source();
+    }
+
+    String getData() {
+        return this.data;
+    }
+}
+
+// FALSE POSITIVE: constructor ignores its argument.
+class IgnoresArg {
+    String data;
+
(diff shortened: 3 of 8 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -633,11 +633,18 @@ let effects_of_call_func_arg fun_exp fun_shape args_taints =
...
-  (* First try to look up via call graph to get the correct node with definition token *)
+(* Extract the first real token from an expression's eorig, falling back
+   to the given default. Useful for call graph edge matching. *)
+let tok_of_eorig ~(default : Tok.t) (exp : IL.exp) : Tok.t =
+  match exp.eorig with
+  | SameAs orig_exp ->
+      (match AST_generic_helpers.ii_of_any (G.E orig_exp) with
+      | tok :: _ when not (Tok.is_fake tok) -> tok
+      | _ -> default)
+  | _ -> default
```
dimitris-m: It is not clear why this is needed.

I recommend to have smaller commits and at the minimum some explanation for every change.

Will try to adapt but it's hard to understand why some changes are needed.

dimitris-m: In particular "Useful for call graph edge matching" does not help much... want to know why this is needed now and how it relates to the aims of this PR.

src/analyzing/AST_to_IL.ml
```diff
@@ -237,12 +237,23 @@ let is_hcl lang : bool =
...
+   which meant it only worked when the pro engine's naming pass had run.
+   In OSS mode id_resolved is never set, so the IL New always got
+   constructor=None, falling through to all_args_taints (accidental leak).
+   The guard was removed because G.New is always a constructor call and
+   the fallback behavior is unchanged when no signature is found. *)
 let mk_class_constructor_name (ty : G.type_) cons_id_info : G.name option =
   match ty with
   | { t = TyN (G.Id (id, _)); _ }
-  | { t = TyExpr { e = G.N (G.Id (id, _)); _ }; _ }
-  (* FIXME: JS parser produces this ^ although it should be parsed as a 'TyN'. *)
```
dimitris-m, line 244: Is this comment still valid? We can delete it but if this is still true we should open an issue so it's not lost.

corneliuhoffman: I think the comment does not matter. The construct is used in their "tainting-pro" (equivalent to intrafile) and so removing it makes no difference to other things

src/tainting/Dataflow_tainting.ml
```diff
@@ -1971,11 +2023,29 @@ let call_with_intrafile lval_opt e env args instr =
...
+             * `_tmp = Foo(x); obj = _tmp`. Without this, the shape is lost
+             * at the assignment boundary. *)
+            let shape =
+              if resolves_to_constructor then
+                match lval_opt with
+                | Some lval -> (
+                    match Lval_env.find_lval lval_env lval with
+                    | Some (S.Cell (_, s)) when
+                      (match s with
+                      |S.Bot -> false
```
dimitris-m: Minor: missing space before `S.Bot`

corneliuhoffman: fixed

PR #633, opened 2026-03-31
Dockerfile: Preprocess for better line continuations and comments

### LINKED ISSUE #629: Dockerfile parsing errors
**Describe the bug**
We found a couple of bugs when parsing Dockerfile. Some of them seem to be existing from Semgrep.

# Opengrep Dockerfile Parser Bug Report

## Summary

The Opengrep Dockerfile parser crashes on three valid Dockerfile constructs, emitting
`[WARN] Syntax error` and skipping lines instead of continuing the scan. All files below
are valid Dockerfiles that build successfully with Docker. Each issue has a minimal
reproducer Dockerfile attached.

---

## Issue 1 — Trailing space after backslash line continuation

**File:** `issue2_trailing_space_after_backslash.Dockerfile`

**Trigger:** A `\` continuation character followed by a trailing space before the newline.

```dockerfile
FROM ubuntu:22.04

RUN apt-get update && \
    apt-get install -y curl \ 
    && apt-get clean
```
*(There is a trailing space after `\` on line 5 — the `curl` line)*

**Expected behaviour:** Parsed successfully. Docker is lenient about trailing spaces after `\`.

**Actual error:**
```
[WARN] Other syntax error at line issue2_trailing_space_after_backslash.Dockerfile:1:
 Tok.NoTokenLocation("Could not convert (line 6, column 31) into a byte offset in file
 issue2_trailing_space_after_backslash.Dockerfile. Invalid location for token "apt-get".")
```

**Root cause:** The trailing space after `\` causes the parser's position tracker to lose
sync with the actual byte offsets in the file. When it reaches the next line (line 6,
`&& apt-get clean`), it cannot map column 31 — where `apt-get` begins — to a valid byte
offset, resulting in a `Tok.NoTokenLocation` crash.

**Reference:** semgrep/semgrep Issue #10823 — confirmed bug, labelled `lang:dockerfile` + `parsing`.

---

## Issue 2 — Double backslash with trailing space (`\\ `)

**File:** `issue3_double_backslash_trailing_space.Dockerfile`

**Trigger:** A `\\` followed by a trailing space before the newline, used as a line
continuation in a multi-step `RUN` instruction.

```dockerfile
FROM ubuntu:22.04

RUN git clone https://github.com/example/repo.git \\ 
    && pip install -r repo/requirements.txt
```
*(There is a trailing space after `\\` on line 6 — the git clone line)*

**Expected behaviour:** Parsed successfully.

**Actual error:**
```
[WARN] Syntax error at line issue3_double_backslash_trailing_space.Dockerfile:7:
 `&& pip install -r repo/requirements.txt` was unexpected
```

**Root cause:** The `\\` before the trailing space is interpreted as a literal escaped
backslash (not a line continuation), so the parser treats the newline as ending the `RUN`
instruction. The next line (`&& pip install ...`) is then encountered outside of any
instruction context and flagged as unexpected.

**Observed in the wild:** Four identical real-world Dockerfiles in our corpus
(`dockerfile_4601`, `dockerfile_4791`, `dockerfile_4792`, `dockerfile_4793`) all hit
this crash at line 127, with the offending token being `pip`.

**Reference:** semgrep/semgrep Issue #10823.

---

## Issue 3 — Backtick `escape` directive (Windows Dockerfiles)

**File:** `issue6_backtick_escape_directive.Dockerfile`

**Trigger:** `# escape=\`` directive at the top of the file, setting the backtick as
the line-continuation escape character. This is standard practice for Windows/PowerShell
Dockerfiles to avoid conflicts with Windows path separators.

```dockerfile
# escape=`
FROM mcr.microsoft.com/windows/servercore:ltsc2022

RUN powershell -Command `
    Write-Host "Hello from Windows container"; `
    Get-Date
```

**Expected behaviour:** Parser recognises the `escape` directive and treats backtick
as the continuation character throughout the file.

**Actual error:**
```
[WARN] Syntax error at line issue6_backtick_escape_directive.Dockerfile:7:
 `Write-Host "Hello from Windows container"; `
    Get-Date` was unexpected
```

**Root cause:** The `tree-sitter-dockerfile` grammar hardcodes `\` as the only valid
escape/continuation character and does not implement the `# escape=` parser directive.
The backtick-terminated lines are not recognised as continuations, so the content on
the following lines is encountered outside of any valid instruction context and flagged
as unexpected. The error spans multiple lines because the parser picks up everything
up to the next recognisable token boundary.

**Reference:** tree-sitter/tree-sitter Discussion #1100 — backtick escape directive
is a documented open problem in the Dockerfile tree-sitter grammar.

---

## Common Impact Across All Three Issues

All three bugs result in the parser **silently skipping lines** past the error point.
Rules are not evaluated on those lines, producing **false-negative findings** with no
actionable warning beyond the `Partially scanned` summary count. In a large corpus scan
this can suppress security findings across hundreds of files.

---

## Environment

| Component | Details |
|---|---|
| Tool | Opengrep (fork of Semgrep ≥ v1.100.0) |
| Parser | `tree-sitter-dockerfile` (camdencheek/tree-sitter-dockerfile) |
| Confirmed reproducible | All 3 issues above |

## References

- semgrep/semgrep #10823 — Trailing space after backslash (confirmed bug)
- semgrep/semgrep #4956 — `RUN` with `\ \n #` parse failure (same parser, related)
- semgrep/semgrep #10306 — Square brackets crash parser (same parser, related)
- tree-sitter/tree-sitter Discussion #1100 — Backtick escape directive open problem

- [X] P1: important to fix or quite annoying

**Environment**
Local build, MacOSX

**Use case**
Create signatures for Docerfile

maciejpirog:
Closes #629 

It also fixes comments between line continuations which behaves differently to bash.

From https://docs.docker.com/reference/dockerfile/#format:

> 
> BuildKit treats lines that begin with # as a comment, unless the line is a valid [parser directive](https://docs.docker.com/reference/dockerfile/#parser-directives). A # marker anywhere else in a line is treated as an argument. This allows statements like:
> 
> ```dockerfile
> # Comment
> RUN echo 'we are running some # of cool things'
> ```
> 
> Comment lines are removed before the Dockerfile instructions are executed. The comment in the following example is removed before the shell executes the echo command.
> 
> ```dockerfile
> RUN echo hello \
> # comment
> world
> ```
> 
> The following examples is equivalent.
> 
> ```dockerfile
> RUN echo hello \
> world
> ```


languages/dockerfile/tree-sitter/Parse_dockerfile_tree_sitter.ml
```diff
@@ -1214,12 +1214,210 @@ let ensure_trailing_newline str =
...
+      is equivalent to "RUN echo a b".  tree-sitter's extras mechanism
+      already handles comments between tokens, but the OCaml post-processing
+      code (concat_shell_fragments) relies on the gap between shell-fragment
+      tokens to reconstruct positions.  A comment line in that gap does not
+      carry a synthetic continuation, so the chain can break.  We replace
+      comment lines with spaces of the same byte count and, when inside a
+      continuation, append a '\' at the last position so tree-sitter sees an
+      unbroken chain of line_continuation extras.
+
+   Position safety: line_col_to_pos reads the *original* file and maps
```
corneliuhoffman, line 1251: My question is what is the influence of whitespace removal on the bytepos_linecol_convertors ....

maciejpirog: We use the original file when creating the converter to bytepos used in `parse`, and the original string in `parse_pattern`. We only remove characters after which there is never a meaningful token (whitespace after line continuation and spanning-till-the-end-of-line comment), so no danger.

corneliuhoffman: I think that is ok, the only worry I had is whether when we do reporting we accidentally use  linecol_to_bytepos_fun to recover the actual finds

maciejpirog: when reporting, bytepos is taken from the tokens, which is set up correctly using the original file. At least this is my understanding. Tests show that the positions of findings are correct in files heavily preprocessed by this function

PR #638, opened 2026-04-03
Support chained method calls on constructor results with --taint-intrafile

corneliuhoffman:
Problem

  When a constructor result is used directly in a method call chain, taint doesn't flow through. The user must assign to an intermediate variable:

  ## This works:
```python
  obj = User(source())
  result = obj.get_profile()
  sink(result)  # finding
```
  ## This doesn't:
```python
  result = User(source()).get_profile()
  sink(result)  # no finding
```
  Same issue across all constructor syntaxes:
  - Java/JS/TS/C#: new User(source()).getProfile()
  - Ruby: User.new(source()).get_profile()
  - Kotlin/Scala: User(source()).getProfile()

  Root cause

  When the IL desugars Passthrough(source()).get_data() it produces:
```python
  _tmp1 = Passthrough(source())   # constructor call
  _tmp2 = _tmp1.get_data()        # method call
```
  Two problems:

  1. No call graph edge for the method: identify_callee had no pattern for DotAccess(Call(...), _, method) — it could resolve obj.method() (via object_mappings) and ClassName() (via constructor resolution), but not
   the combination where the receiver is itself a constructor call.
  2. Token collision: Both IL instructions' eorig point back to the same first AST token (Passthrough). The constructor edge and method edge would collide in the call graph. And resolves_to_constructor would
  incorrectly match _tmp1.get_data() to the constructor edge, remapping the callee to nonsense.

  Fix

  The fix has two parts: resolving the callee in the call graph, and a consistent token strategy that avoids collisions.

  Part 1: Chained call resolution (Graph_from_AST.identify_callee)

  New pattern matching for DotAccess(receiver, _, method) where the receiver is a constructor expression. Extracts the class name from three constructor syntaxes with language guards, then looks up
  ClassName#method_name in all_funcs:

  - Call(N(Id(ClassName)), args) — Python, Kotlin, Scala
  - New(_, type, _, args) — Java, JS, TS, C#
  - Call(DotAccess(N(Id(ClassName)), _, "new"), args) — Ruby

  Single match returns immediately. Multiple matches (overloaded methods in Java/Kotlin/C#/Scala) are disambiguated by call-site arity. Languages without overloading (Python, Ruby, JS, TS) always have at most one
  match by name, so arity never fires.

  The all_funcs lookup acts as an implicit constructor check: only classes with methods have two-element fn_id entries ([Some class; Some method]). If the receiver isn't actually a class constructor, the filter
  returns empty and we return None.

  Part 2: Token strategy simplification

  Previously, call graph edge tokens and lookups used different strategies:
  - Edges: first token of the full call expression (e.g., obj in obj.method())
  - Lookups: tok_of_eorig which chased the IL expression's eorig back-pointer to recover the original AST token

  This was fragile — for chained calls, tok_of_eorig on _tmp.get_data() returned Passthrough (the first AST token), colliding with the constructor edge.

  The new strategy is simpler and consistent: DotAccess edges are stored at the method name's token position, and lookups use the method name's IL token. These match because both point to the same source position.

  Concretely:

  - extract_calls: For DotAccess calls, use the method name token for the edge. Exception: Ruby's ClassName.new() keeps using the class name token (first token of expression), since the constructor machinery looks
  up by class name position.
  - get_signature_for_object: Simplified to a single lookup via Function_id.tok method_name, with a name-based fallback. No more tok_of_eorig.
  - this.method() / self.method() path: Uses snd method_name.IL.ident (the method token) instead of self_tok.
  - Simple function calls (func()): Uses snd name.ident directly — the IL variable's token is already at the right position.
  - tok_of_eorig removed: No longer needed. All lookups now use IL name tokens directly.
  - resolves_to_constructor: Two changes:
    - Dot guard: Skip the constructor check for Dot accesses (_tmp.get_data()) unless it's Ruby's ClassName.new(). Without this, the lookup would find the constructor edge via the shared class-name token and
  incorrectly remap the method call.
    - Token extraction: Uses snd name.ident from the callee's base variable instead of tok_of_eorig.

  Supporting changes

  - Function_id.tok: New accessor to extract the raw Tok.t from a function identifier, used by get_signature_for_object.
  - Call_graph.Display.edge_attributes: Show call-site line:col in DOT export for debugging.

  Tests

  Added chained method call examples to 8 existing constructor test files: Python, Ruby, Java, JavaScript, TypeScript, Kotlin, C#, Scala. All pass.

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -633,41 +633,16 @@ let effects_of_call_func_arg fun_exp fun_shape args_taints =
             (S.show_shape fun_shape));
       []
 
-(* Call graph edges are keyed by source-level token positions, but IL
-   expressions often carry tokens from IL construction (e.g., fresh
-   variables from class_construction). The eorig back-pointer preserves
-   the original source position, which we need for call graph lookups. *)
-let tok_of_eorig ~(default : Tok.t) (exp : IL.exp) : Tok.t =
-  match exp.eorig with
-  | SameAs orig_exp ->
-      (match AST_generic_helpers.ii_of_any (G.E orig_exp) with
-      | tok :: _ when not (Tok.is_fake tok) -> tok
-      | _ -> default)
-  | Related orig_any ->
-      (match AST_generic_helpers.ii_of_any orig_any with
-      | tok :: _ when not (Tok.is_fake tok) -> tok
-      | _ -> default)
-  | NoOrig -> default
 
-let get_signature_for_object graph caller_node db method_name obj (fun_exp : IL.exp) arity =
+let get_signature_for_object graph caller_node db method_name arity =
   let caller = Option.map Function_id.of_il_name caller_node in
   let method_tok = Function_id.tok method_name in
-  let call_tok = tok_of_eorig ~default:(snd obj.ident) fun_exp in
-  (* Try method name token first: for chained calls like
-     ClassName(...).method(), the eorig token (call_tok) points to
-     ClassName which collides with the constructor edge. The method
-     token points to the actual method and avoids this collision. *)
-  let try_graph tok =
-    match Call_graph.lookup_callee_from_graph graph caller tok with
-    | Some callee_node -> Shape_and_sig.(lookup_signature db callee_node arity)
-    | None -> None
-  in
-  match try_graph method_tok with
-  | Some _ as r -> r
-  | None ->
-      (match try_graph call_tok with
-      | Some _ as r -> r
-      | None -> Shape_and_sig.lookup_signature db method_name arity)
+  (* Look up via method name token — call graph edges for DotAccess calls
+     are stored at the method token position (see extract_calls). *)
+  match Call_graph.lookup_callee_from_graph graph caller method_tok with
+  | Some callee_node ->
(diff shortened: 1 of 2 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -1972,6 +1984,16 @@ let call_with_intrafile lval_opt e env args instr =
...
         let resolves_to_constructor =
+          (* Method calls on objects (e.g., _tmp.get_data()) should not be
+             remapped as constructors. Their eorig may share a token with a
+             constructor edge (e.g., in Passthrough(source()).get_data(), both
+             the constructor and the method eorig start at "Passthrough").
+             Skip the constructor check for Dot accesses unless it's Ruby's
+             ClassName.new() pattern. *)
+          (match e.e with
+          | Fetch { rev_offset = [{ o = Dot name; _ }]; _ }
+            when fst name.IL.ident <> "new" -> false
```
dimitris-m: if this is only for Ruby do we have access to lang so we can qualify this?

dimitris-m: something like:
```
when not (Lang.(equal env.taint_inst.lang Ruby) && fst name.IL.ident = "new") -> false
```

PR #640, opened 2026-04-03
elixir: fix taint propagation through map value expressions

dimitris-m:
## Summary

- Unwrap `OtherExpr("MapPairArrow"/"MapPairKeyword")` tags in the `dict` function of `AST_to_IL.ml` so map entries are processed as normal key-value pairs.
- Without this, taint through expressions nested inside map literals (e.g. `%{sink(k) => v}`) was lost because the wrapper made entries opaque to the IL translation.

Follow-up to #637.

## Test plan

- `taint-map-value` test covers taint flow through arrow keys (including nested expressions), keyword values, and direct calls.


corneliuhoffman (approved):
lgtm

PR #642, opened 2026-04-05
Elixir: Updates to syntax (part 1)

maciejpirog:
- Keep expression when translating `OtherPat (("ExprToPattern", ...), ...)` to IL (not specific to Elixir)
- Support `rescue` clause in function definition
- Represent structs using the `Constructor` expression
- Represent "join" in lists (e.g. `[1, 2 | t]`) as `Constructor`
- Support default values in function args (`def foo(x \\ "default value")`)

### DIFF
--- a/languages/elixir/ast/Elixir_to_elixir.ml
+++ b/languages/elixir/ast/Elixir_to_elixir.ml
@@ -54,7 +54,9 @@ class ['self] visitor =
         exprs
         |> List_.map (function
              | I id -> P { pname = id; pdefault = None }
-             (* TODO: recognize default value with \\ *)
+             (* In Elixir you can have a default value only for ident params (no pats) *)
+             | BinaryOp (I id, (ODefault, tok), d) -> 
+                 P { pname = id; pdefault = Some (tok, d) }
              | x -> OtherParamExpr x)
       in
       let ys =
--- a/tests/tainting_rules/elixir/taint-param-with-default.ex
+++ b/tests/tainting_rules/elixir/taint-param-with-default.ex
@@ -1,0 +1,4 @@
+def foo(x \\ "default value") do
+  # ruleid: taint
+  sink(x)
+end
--- a/tests/tainting_rules/elixir/taint-param-with-default.yaml
+++ b/tests/tainting_rules/elixir/taint-param-with-default.yaml
@@ -1,0 +1,17 @@
+rules:  
+- id: taint
+  languages: [elixir]
+  message: "tainted data reached sink"
+  severity: ERROR
+  mode: taint
+  pattern-sources:
+    - patterns:
+      - pattern-either:
+          - pattern-inside: |
+              def $_(..., $X, ...) do
+              ...
+              end
+      - focus-metavariable: $X
+  pattern-sinks:
+    - pattern: |
+        sink(...)


languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -118,6 +118,51 @@ let expr_of_expr_or_kwds (x : (G.expr, keywords_generic) Either_.t) : G.expr =
...
+               | G.OtherExpr
+                   ( (("MapPairKeyword" | "MapPairArrow"), _),
+                     [ G.E
+                         { e = G.Container (G.Tuple, (_, [ key_e; val_e ], _)); _ }
+                     ] ) ->
+                   let di_opt =
+                     match key_e.G.e with
+                     | G.N (G.Id (id, _)) -> Some [ id ]
+                     | G.L (G.Atom (_, id)) -> Some [ id ]
+                     | G.L (G.String (_, (s, tok), _)) -> Some [ (s, tok) ]
```
dimitris-m: These patterns could probably be combined, am I wrong?

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -118,6 +118,51 @@ let expr_of_expr_or_kwds (x : (G.expr, keywords_generic) Either_.t) : G.expr =
...
+                     | _ -> None
```
dimitris-m: Can this `None` case ever happen?

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -118,6 +118,51 @@ let expr_of_expr_or_kwds (x : (G.expr, keywords_generic) Either_.t) : G.expr =
...
+                   in
+                   Option.map (fun di -> (di, elixir_expr_to_pattern val_e)) di_opt
+               | G.Ellipsis _ -> None
```
dimitris-m: Presumably this is ok because for PatRecord we happen to not need Ellipsis, it is somehow implied in matching right? Smaller record in pattern matches larger record?

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -118,6 +118,51 @@ let expr_of_expr_or_kwds (x : (G.expr, keywords_generic) Either_.t) : G.expr =
...
+ *)
+let rec elixir_expr_to_pattern (e : G.expr) : G.pattern =
+  match e.G.e with
+  | G.Container (G.Dict, (l, items, r)) ->
+      let pat_fields =
+        items
+        |> List_.filter_map (fun item ->
+               match item.G.e with
+               | G.OtherExpr
+                   ( (("MapPairKeyword" | "MapPairArrow"), _),
```
dimitris-m: In H.expr_to_pattetn we have: `  | OtherExpr (tag, [ E e ]) -> OtherPat (tag, [ P (expr_to_pattern e) ])` which is on purpose so these forms can be distinguished.

Here they seem to be collapsed again.

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -118,6 +118,51 @@ let expr_of_expr_or_kwds (x : (G.expr, keywords_generic) Either_.t) : G.expr =
...
 
+(* Elixir-specific expression-to-pattern conversion.
+ * Unlike H.expr_to_pattern, this handles:
+ *   - Container(Dict, ...) → PatRecord  (plain maps: %{a: b, ...})
+ *   - Constructor(name, [dict]) → PatConstructor  (structs: %Foo{a: b, ...})
+ * and falls back to H.expr_to_pattern for everything else.
+ *)
+let rec elixir_expr_to_pattern (e : G.expr) : G.pattern =
+  match e.G.e with
+  | G.Container (G.Dict, (l, items, r)) ->
```
dimitris-m: Why is this case needed? H.expr_to_pattern can handle it no?

languages/elixir/generic/Elixir_to_generic.ml
maciejpirog: What would be the correct translation here? i.e. a pattern that distinguishes between the two and still propagates taint?

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -118,6 +118,51 @@ let expr_of_expr_or_kwds (x : (G.expr, keywords_generic) Either_.t) : G.expr =
...
+      (* Expand Join(head, tail) items inline so [a, b | rest] produces
+       * PatList([a_pat, b_pat, rest_pat]) with all vars directly visible. *)
+      let expand item =
+        match item.G.e with
+        | G.OtherExpr (("Join", _), [ G.E head_e; G.E tail_e ]) ->
+            [ elixir_expr_to_pattern head_e; elixir_expr_to_pattern tail_e ]
+        | _ -> [ elixir_expr_to_pattern item ]
+      in
+      G.PatList (l, List.concat_map expand items, r)
+  | _ -> H.expr_to_pattern e
```
dimitris-m: how about nested patterns? 
we never return to this function once we descend into `H.expr_to_pattern`.

tests/tainting_rules/elixir/taint-rescue.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+  pattern-sources:
+    - patterns:
+      - pattern-either:
+          - pattern-inside: |
+              def $_(..., $X, ...) do
+              ...
+              end
+          - pattern-inside: |
+              defp $_(..., $X, ...) do
```
dimitris-m: defp is not needed because "private" becomes an attribute, so if I remember correctly, `def` pattern is enough.

dimitris-m:
`elixir_expr_to_pattern` does not recurse back into itself when falling through to `H.expr_to_pattern`. Since `H.expr_to_pattern` recurses into itself (not back to `elixir_expr_to_pattern`), Elixir-specific pattern handling (Join expansion, Dict->PatRecord, Constructor->PatConstructor) is lost for anything nested inside a construct that only `H.expr_to_pattern` handles (e.g. tuples).

**Reproduction:**

```elixir
# pipe-in-tuple.ex
defmodule PipeInTuple do
  def top_level([a, b | rest]) do
    sink(a)
    sink(b)
    sink(rest)
  end

  def pipe_in_tuple({[a, b | rest]}) do
    sink(a)
    sink(b)
    sink(rest)
  end
end
```

```yaml
# rule.yaml
rules:
- id: pipe-in-tuple-taint
  mode: taint
  languages: [elixir]
  message: "tainted via param"
  severity: INFO
  pattern-sources:
    - patterns:
      - pattern-inside: |
          def $_(..., $X, ...) do
          ...
          end
      - focus-metavariable: $X
  pattern-sinks:
    - pattern: sink(...)
```

```
$ opengrep scan --config rule.yaml pipe-in-tuple.ex
Ran 1 rule on 1 file: 4 findings.
```

`top_level` finds all 3 sinks. `pipe_in_tuple` only finds `sink(a)` -- taint is lost for `b` and `rest`.

The named AST confirms: inside the tuple, the list falls through to `H.expr_to_pattern` which can't expand `Join`. `b` and `rest` end up in `OtherPat("ExprToPattern", ...)` with `id_resolved=Ref(None)`.

```
# top_level: all resolved
PatList([PatId("a", LocalVar 4), PatId("b", LocalVar 5), PatId("rest", LocalVar 6)])

# pipe_in_tuple: b and rest unresolved
PatTuple([PatList([PatId("a", LocalVar 8), OtherPat("ExprToPattern", [E(OtherExpr("Join", [E(Id("b", None)), E(Id("rest", None))]))])])])
```

languages/elixir/generic/Elixir_to_generic.ml
dimitris-m: Well in AST_to_IL we have:
```
  | G.OtherPat ((("MapPairArrow" | "MapPairKeyword"), _), [ G.P inner ])
    when env.lang =*= Lang.Elixir ->
    pattern env inner
```

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -471,6 +471,71 @@ and map_param_to_gparam env (p : parameter) : G.parameter =
...
+
+(* Wrap a body statement in a G.Try when there are rescue/catch/after clauses.
+ * `tok` is used as the try token. *)
+and wrap_with_rescue env tok body_stmt rescue =
+  match rescue with
+  | [] -> body_stmt
+  | extras ->
+      let catches = ref [] in
+      let try_else = ref None in
+      let finally = ref None in
```
dimitris-m: Are you really willing to lose 3 beers for this?!

dimitris-m: This PR _could_ have been perfect

tests/tainting_rules/elixir/taint-cond-case.ex
```diff
@@ -0,0 +1,7 @@
+def foo() do
+  x = source()
+  cond do
+    # ruleid: cond_taint_rule
+    y = sink(x) -> 3
+  end
+end
```
dimitris-m, line 7: I also tested the following which works ok, and maybe we could add something like that too:

```elixir
def foo() do
  cond do
    y = source() -> sink(y)
    true -> sink(y)
  end
end
```

Even though the second `y` has the same sid, the assignment seems to only take place in the first case, so we only get the expected result.

tests/tainting_rules/elixir/taint-rescue.yaml
dimitris-m: reminder: this case is not needed ^

tests/tainting_rules/elixir/taint-cond-case.ex
dimitris-m: Of course ideally the second branch's y should be unresolved.

languages/elixir/generic/Elixir_to_generic.ml
maciejpirog: fixed. (also, local ref is only half a beer!)

tests/tainting_rules/elixir/taint-cond-case.ex
maciejpirog: the documentation is a bit vague here, but my understanding of the semantics is that the assignment `y = source()` always takes place as a part of checking the conditions and `y` in the second branch is the same `y`, so it actually should get the same sid

tests/tainting_rules/elixir/taint-rescue.yaml
maciejpirog: fixed

tests/tainting_rules/elixir/taint-cond-case.ex
dimitris-m: Then my example should also have taint in the second branch, which is not the case because the assignment is confined in IL to the first branch?

maciejpirog: Oh, I see... We should address this at some point

PR #649, opened 2026-04-08
Fix obj[key] parsing in ruby

corneliuhoffman:
## Summary

   Fix Ruby `obj[key]` parsing: change from `Call(DotAccess(obj, Op_AREF), [key])` to `ArrayAccess(obj, [key])`.

   This is correct semantically and it also fixes taint propagation through hash access patterns like `params[:to]` and `continue_params[:name]` in intrafile mode, where the old Call-based representation caused the taint engine to lose taint through the spurious method call.

   ## Changes

   ### `languages/ruby/ast/ast_ruby.ml`
   - Add `ArrayAccess of expr * arguments bracket` variant to the `expr` type (replaces the `TODO` comment)

   ### `languages/ruby/tree-sitter/Parse_ruby_tree_sitter.ml`
   - Parse `obj[key]` as `ArrayAccess(obj, args)` instead of `Call(DotAccess(obj, Op_AREF), args)`

   ### `languages/ruby/generic/ruby_to_generic.ml`
   - Translate `ArrayAccess` to `G.ArrayAccess`, always wrapping arguments in `Container(Array, ...)` so that patterns like `DB[$QUERY,...]` with ellipsis match correctly against single-arg targets like `DB["query"]`

   ### `src/matching/Generic_vs_generic.ml`
   - Add Ruby-specific matching equivalence: pattern `N(Id(x))` matches target `Call(N(Id(x)), [])` for Ruby. After disambiguation, target bare identifiers inside `ArrayAccess` receivers become `Call(x, [])`, but patterns keep the bare `N(Id(x))` form. This equivalence bridges the gap so that existing rules  like `params[...]` continue to match. Excludes metavariables so they fall through to the general metavar matching case.

   ### Test
   - `tests/rules/cross_function_tainting/test_ruby_array_access.rb` + `.yaml`: New intrafile taint test where `continue_params[:to]` returns a tainted value. Previously failed because the Call-based parsing broke taint propagation through hash access.
   - tests/patterns/ruby/array_access.rb + .sgrep: Verifies that params[...] pattern matches bare identifier hash access (params[:id], params[:to] inside a call, params()[:name])
   - tests/patterns/ruby/array_access_call.rb + .sgrep: Verifies that params()[...] pattern matches the same cases, and does NOT match when params is a local variable (assigned before use)
   ## Context

   The old parsing of `obj[key]` as `Call(DotAccess(obj, Op_AREF), [key])` caused two problems:
   1. **Taint loss in intrafile mode**: The spurious `Call` to `[]` method had no signature, so taint from the receiver was lost
   2. **Semantic mismatch**: `obj[key]` is array/hash access, not a method call — `ArrayAccess` correctly represents this in the generic AST

   The disambiguation visitor (`Disambiguate_ruby_calls`) naturally visits `ArrayAccess` receivers, wrapping unresolved bare identifiers like `params` in `Call()`. This is correct for the IL (enables method call evaluation) but creates a mismatch with patterns. The matching equivalence resolves this without requiring changes to existing rules.


src/matching/Generic_vs_generic.ml
```diff
@@ -952,6 +952,15 @@ and m_expr ?(is_root = false) ?(arguments_have_changed = true) a b =
...
       >||> m_with_symbolic_propagation ~is_root (fun b1 -> m_expr a b1) b
+  (* Ruby equivalence: bare identifier `foo` matches zero-arg call `foo()`.
+     After disambiguation, target bare identifiers become Call(N(Id(x)), [])
+     but patterns keep the bare N(Id(x)) form. Exclude metavars so they
+     fall through to the general metavar case below. *)
+  | G.N (G.Id ((str, _), _) as na), B.Call ({ e = B.N (B.Id _ as nb); _ }, (_, [], _))
+    when not (Mvar.is_metavar_name str) ->
+      with_lang (fun lang ->
+          if lang =*= Lang.Ruby then m_name na nb
+          else fail ())
```
dimitris-m, line 963: Well, is there no way to write a pattern than will match only calls vs bare ident? 
Perhaps by adding ()? 

I don't see why we should identify such terms.

dimitris-m: Not convinced about this part.

corneliuhoffman: because in the pattern something like `params[...]` becomes `ArrayAccess(Id("params"), Array([Ellipsis]))`  and needs to match `params[:id]` which parses as `ArrayAccess(Call(Id("params"), []), Array([Ellipsis]))`
The point is taht you cannot modify the pattern parser to include teh Call bc you don't know if param is not going to appear as a local variable

corneliuhoffman: Thre might be a better solution, I just could not think of one

src/parsing/Disambiguate_ruby_calls.ml
```diff
@@ -23,6 +23,14 @@ class ['self] visitor =
 
     method! visit_expr_kind env ek =
       match ek with
+      (* Unwrap zero-arg Call around local variables: tree-sitter parses
+         bare `config` as Call(config, []) since it could be a method call,
+         but if naming resolved it as a local variable, drop the Call. *)
```
dimitris-m: I don't understand, tree-sitter does not parse bare config as Call, do you have an example?

dimitris-m: this might happen in ruby_to_generic, function `expr_as_stmt`; there is a comment in fact.

dimitris-m: https://github.com/opengrep/opengrep/pull/651

src/matching/Generic_vs_generic.ml
dimitris-m, line 963: This seems not needed, I think #651 is enough to solve the issues we had.

src/parsing/Disambiguate_ruby_calls.ml
```diff
@@ -23,6 +23,14 @@ class ['self] visitor =
...
+      | Call ({ e = N (Id (_name, info)); _ } as callee, (_lb, [], _rb))
+        when (match !(info.id_resolved) with
+              | Some (LocalVar, _) -> true
+              | _ -> false) ->
+          callee.e
```
dimitris-m: Also this seems not needed, #651 is enough I think.

src/matching/Generic_vs_generic.ml
dimitris-m: Needed as you pointed out to me offline

dimitris-m: Ok, we discussed, and indeed we will keep this.

dimitris-m (approved):

PR #652, opened 2026-04-09
Taint expression vars in patterns that come from `ExprToPattern`

maciejpirog:
When collecting variables in `Visit_pattern_ids`, we now also collect variables from expressions that cannot be meaningfully converted to patterns, but are instead wrapped in `ExprToPattern`, i.e., patterns of the shape

```OCaml
G.OtherPat (("ExprToPattern", _), [ G.E { e; _ } ])
```

**Example:**

In Elixir, we have a pattern that splits a string into a prefix and suffix:

```Elixir
def foo("abc" <> x) do
  sink(x)
end
```

It is translated to a call with `<>` as the function and `"abc"` and `x` as the arguments.

I think this solution is better than:
- Having a special case in Elixir, and other languages
- Adding special language-dependent cases in `AST_generic_helpers.expr_to_pattern` (esp that `AST_generic` or the helpers are not language-dependent anywhere at the moment)

### DIFF
--- a/src/analyzing/Visit_pattern_ids.ml
+++ b/src/analyzing/Visit_pattern_ids.ml
@@ -15,10 +15,31 @@
 
 module G = AST_generic
 
-class ['self] pat_id_visitor =
+(* This one is used in OtherPattern(("ExprToPattern"),...) [E e]).
+ * In such a case variables in e should be understood as pattern variables.
+ * NOTE: we need two visitors, because not all expr-variables are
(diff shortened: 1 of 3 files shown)


src/analyzing/Visit_pattern_ids.ml
```diff
@@ -15,10 +15,31 @@
 
 module G = AST_generic
 
-class ['self] pat_id_visitor =
+(* This one is used in OtherPattern(("ExprToPattern"),...) [E e]).
```
dimitris-m, line 18: OtherPat

PR #659, opened 2026-04-14
Elixir: distinguish field access from zero-arity remote call

dimitris-m:
## Summary

Elixir's `foo.bar` (no parens, no args, no do-block) is map/struct field access; `foo.bar(...)` is a remote function call. The parser was collapsing both into `Call(DotAccess, [])`, so `taint_assume_safe_functions: true` silently dropped taint at every field access on a tainted receiver (e.g. `conn.params`, `upload.path`).

- Add `FieldAccess of remote_dot` in `AST_elixir.ml` alongside `DotRemote`.
- In `Parse_elixir_tree_sitter.ml`, emit `FieldAccess rdot` when the `_remote_call_without_parentheses` CST case has no args and no do-block; keep `Call (...)` otherwise. The tree-sitter grammar already separates the two non-terminals, so no token introspection is needed.
- In `Elixir_to_generic.ml`, translate `FieldAccess` to `G.DotAccess` -- the same shape used by other languages for field access. Explicit `foo.bar()` still becomes `Call(DotAccess, [])` and remains subject to `taint_assume_safe_functions`.

New test: `tests/tainting_rules/elixir/taint-field-access.{yaml,ex}` covers field access (must propagate) and the four call forms -- zero-arity `.path()`, `.compute(1,2)`, no-parens-with-args `.compute 1, 2`, and do-block `.with_block do ... end` (all should drop taint under `taint_assume_safe_functions: true`).

### DIFF
--- a/languages/elixir/tree-sitter/Parse_elixir_tree_sitter.ml
+++ b/languages/elixir/tree-sitter/Parse_elixir_tree_sitter.ml
@@ -674,7 +674,7 @@ and map_capture_expression (env : env) (x : CST.capture_expression) =
               | other -> other
             in
             (match actual_fun_name with
-            | I _ | Alias _ | DotAlias _ | DotRemote _ ->
+            | I _ | Alias _ | DotAlias _ | DotRemote _ | FieldAccess _ ->
                 (* Convert &fun/arity to &(fun(&1, &2, ...)) *)
                 let arity_int = Int64.to_int arity in
                 (* Create PlaceHolder arguments: &1, &2, ... *)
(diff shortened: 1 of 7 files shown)


languages/elixir/tree-sitter/Parse_elixir_tree_sitter.ml
corneliuhoffman: I think there is a problem with the Short lambda parsing `&Class.method/1` parses as s fraction. you need to add the Filed_acces  to the list

dimitris-m: you are right, we should have a test and this should have been caught!
will fix and add.

dimitris-m: However I added a test and it failed, so there is a problem that has do do with HOFs in that case, which I will try to fix now.

dimitris-m: Fixed

corneliuhoffman (commented):
LGTM now

corneliuhoffman (approved):

Issue #615, opened 2026-03-16
--taint-intrafile: false positive when two functions share the same name


dimitris-m:
## Reproduction

`taint-func-param.yaml`:
```yaml
rules:
  - id: taint-func-param
    message: Tainted data reaches sink
    languages: [go]
    severity: ERROR
    mode: taint
    pattern-sources:
      - patterns:
          - focus-metavariable: $SOURCE
          - pattern: func (..., $SOURCE $TYPE, ...) { ... }
    pattern-sinks:
      - pattern: sink($X)
```

`taint-func-param.test.go`:
```go
package main

func test(input string) {
	var fn = func(s string) {
		// ruleid: taint-func-param
		sink(s)
	}
	fn("")
}

func test(input string) {
	var fn = func(s string) {
		// ok: taint-func-param
		sink(s)
	}
}
```

```
opengrep --test . --taint-intrafile
```

## Expected

Line 6: finding. Line 14: no finding.

## Actual

Both line 6 and line 14 are flagged. Calling `fn("")` in the first `test` function causes a finding in the second `test` function where `fn` is never called.

PR #663, opened 2026-04-16
unconditional taints in lambdas

corneliuhoffman:
## Summary

Under `--taint-intrafile`, a rule that matches a concrete source inside
a lambda produced no finding when the lambda was not called. For example

```javascript
function test1(a) {
    var x = function(data) {
        var t = source();
        sink(t); };
}
```

is not found, but

```javascript
function test1(a) {
    var x = function(data) {
        var t = source();
        sink(t); };
    x(a);
}
```

is found.

## Cause

`run_check_fundef_if_needed` in `Match_tainting_mode.ml` skipped
`check_fundef` for `is_lambda_assignment` entries; the lambda's effects
were stored in the signature but never materialised as findings when no
caller site could be resolved.

## Change

`run_check_fundef_if_needed` now always runs `check_fundef`. For lambda
assignments the returned effects are filtered to keep only `ToSink`
effects whose taint has a concrete `Src` origin, preserving the sink's
original precondition. Parameterised (`BArg`) taints remain in the
signature and fire at resolved call sites as before, via signature
instantiation.

## Tests

- New `tests/rules/cross_function_tainting/test_lambda_in_object_literal.{yaml,js}`
  covers the object-literal case above.
- Updated `tests/rules/cross_function_tainting/test_same_name_functions.go`;
  the #615 regression test previously conflated "same-name confusion"
  with the uncalled-lambda case.

### Closures over outer-scoped taint

If the lambda closes over a tainted variable, the taint is still not
propagated through unless the lambda is called. For example

```javascript
function test1() {
    var t = source();
    var x = function() { sink(t); };
}
```

is not found, but

```javascript
function test1() {
    var t = source();
    var x = function() { sink(t); };
    x();
}
```

is.

### DIFF
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -635,35 +635,31 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
                Effects whose taint is purely parameterized (BArg) still ride
                through the signature at resolved call sites; effects mixing
                both get an Src-only slice surfaced here. *)
+            let keep_src_toSink_only (eff : Effect.t) : Effect.t option =
+              match eff with
+              | Effect.ToSink si ->
+                  let items, precond = si.taints_with_precondition in
+                  let src_items =
+                    List.filter
+                      (fun (i : Effect.taint_to_sink_item) ->
+                        match i.taint.orig with
+                        | Taint.Src _ -> true
+                        | _ -> false)
+                      items
+                  in
+                  if List_.null src_items then None
+                  else
+                    Some
+                      (Effect.ToSink
+                         {
+                           si with
+                           taints_with_precondition = (src_items, precond);
+                         })
+              | _ -> None
+            in
             let effects_to_record =
               if info.is_lambda_assignment then
-                fdef_effects
-                |> Effects.elements
-                |> List.filter_map (fun eff ->
-                       match eff with
-                       | Effect.ToSink sink_info ->
-                           let items, _ =
-                             sink_info.taints_with_precondition
-                           in
-                           let src_items =
-                             List.filter
-                               (fun (item : Effect.taint_to_sink_item) ->
-                                 match item.taint.orig with
-                                 | Taint.Src _ -> true
-                                 | _ -> false)
-                               items
-                           in
-                           if List_.null src_items then None
-                           else
-                             Some
-                               (Effect.ToSink
-                                  {
-                                    sink_info with
-                                    taints_with_precondition =
-                                      (src_items, Rule.PBool true);
-                                  })
-                       | _ -> None)
-                |> Effects.of_list
+                Effects.filter_map keep_src_toSink_only fdef_effects
               else fdef_effects
             in
             record_matches effects_to_record;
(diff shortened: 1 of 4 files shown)


maciejpirog (approved):

dimitris-m:
I might ask some questions / request some change.

tests/rules/cross_function_tainting/test_same_name_functions.go
```diff
@@ -11,6 +11,7 @@ func test(input string) {
 func test(input string) {
 	var fn = func(s string) {
 		// ok: taint-func-param
-		sink(s)
+		safe(s)
 	}
+	fn("")
```
dimitris-m, line 16: not sure this test makes sense now

src/engine/Match_tainting_mode.ml
```diff
@@ -622,17 +622,52 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
+                               items
+                           in
+                           if List_.null src_items then None
+                           else
+                             Some
+                               (Effect.ToSink
+                                  {
+                                    sink_info with
+                                    taints_with_precondition =
+                                      (src_items, Rule.PBool true);
```
dimitris-m: it's not clear why `Rule.PBool true` is needed and what the consequences are.

src/engine/Match_tainting_mode.ml
```diff
@@ -622,17 +622,52 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
+               concrete pattern-source match (e.g. a parameter declared as a
+               source via `pattern-inside: function $X(..., $RES, ...) {...}`).
+               Effects whose taint is purely parameterized (BArg) still ride
+               through the signature at resolved call sites; effects mixing
+               both get an Src-only slice surfaced here. *)
+            let effects_to_record =
+              if info.is_lambda_assignment then
+                fdef_effects
+                |> Effects.elements
+                |> List.filter_map (fun eff ->
```
dimitris-m: There is also `Effects.filter_map` which can be used directly.

src/engine/Match_tainting_mode.ml
```diff
@@ -622,17 +622,52 @@ let check_rule per_file_formula_cache (rule : R.taint_rule) match_hook
...
+                           else
+                             Some
+                               (Effect.ToSink
+                                  {
+                                    sink_info with
+                                    taints_with_precondition =
+                                      (src_items, Rule.PBool true);
+                                  })
+                       | _ -> None)
+                |> Effects.of_list
```
dimitris-m: This can also be removed once `Effects.filter_map` is used.

src/engine/Match_tainting_mode.ml
dimitris-m: Can we not inherit the existing precondition?

tests/rules/cross_function_tainting/test_same_name_functions.go
corneliuhoffman: It does, the whole point iof it was that "same name" functions shoudl not confuse the thing (fails before this test was introduced in https://github.com/opengrep/opengrep/pull/617

src/engine/Match_tainting_mode.ml
corneliuhoffman: overeager, wanted to only find things that happen inside the lambda so forgetting other conds but I think there are cases where that will create weird things. Changing

src/engine/Match_tainting_mode.ml
corneliuhoffman: thanks

src/engine/Match_tainting_mode.ml
corneliuhoffman: yes

dimitris-m (approved):

PR #664, opened 2026-04-16
Make switch usable as expression in IL (with applications to Elixir)

maciejpirog:
For some reason `Switch` is a statement in AST, which makes it hard to meaningfully translate code such as this in Elixir:

```elixir
def foo() do
  a = source ()
  b = case a do
        0 -> a
        x -> x
      end
  sink(b)
end
```

In this PR:
- We add an additional case to `stmt_expr` for when `Switch` is used as expression `StmtExpr(Switch((), ...))`
- We add a proper translation of `case` expressions in Elixir to make use of this change

### DIFF
--- a/languages/elixir/generic/Elixir_to_generic.ml
+++ b/languages/elixir/generic/Elixir_to_generic.ml
@@ -151,7 +151,19 @@ let rec expr_to_pattern (e : G.expr) : G.pattern =
       end
   (* TODO: PatKeyVal and more *)
   | _ -> OtherPat (("ExprToPattern", Tok.unsafe_fake_tok ""), [ G.E e ])
-           
+
+let pats_of_args (args : G.argument list) : G.pattern list =
+  List_.map
+    (function
+      | G.OtherArg (("ArgKwdQuoted", _), [ G.E e ]) -> expr_to_pattern e
+      | arg -> H.argument_to_expr arg |> expr_to_pattern)
+    args
+
+let wrap_when (when_opt : (Tok.t * G.expr) option) (pat : G.pattern) : G.pattern =
+  match when_opt with
+  | None -> pat
+  | Some (_tok, e) -> G.PatWhen (pat, e)
+
 (* TODO: lots of work here to detect when args is really a single
  * pattern, or tuples *)
 let pat_of_args_and_when (args, when_opt) : G.pattern =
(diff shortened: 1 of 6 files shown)


languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -185,6 +185,28 @@ let case_and_body_of_stab_clause (x : stab_clause_generic) : G.case_and_body =
...
+  let pats =
+    List_.map
+      (function
+        | G.OtherArg (("ArgKwdQuoted", _), [ G.E e ]) -> expr_to_pattern e
+        | arg -> H.argument_to_expr arg |> expr_to_pattern)
+      args
+  in
+  let pat =
+    match pats with
+    | [] -> G.PatLiteral (G.Null (G.fake "no_arg"))
```
corneliuhoffman, line 194: [] is not possible right?

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -185,6 +185,28 @@ let case_and_body_of_stab_clause (x : stab_clause_generic) : G.case_and_body =
   let stmt = G.stmt1 stmts in
   G.case_of_pat_and_stmt (pat, stmt)
 
+let case_and_body_of_case_clause (x : stab_clause_generic) : G.case_and_body =
```
corneliuhoffman, line 190: Nonblocking: there is huge overlap with pat_of_args_and_when

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -958,6 +980,21 @@ and map_body env v : G.stmt list =
...
+      (_, ([ scrutinee ], []), _),
+      Some (_tdo, (Clauses clauses, []), _tend) ) ->
+      let e = map_expr env scrutinee in
+      let xs = map_clauses env clauses in
+      let cases = xs |> List_.map case_and_body_of_case_clause in
+      G.Switch (tcase, Some (G.Cond e), cases) |> G.s |> G.stmt_to_expr
+  | _ ->
   (* Special handling for DotAnon - extract the inner expression to use as callee *)
   let e = match v1 with
     | DotAnon (inner_expr, _tdot) -> map_expr env inner_expr
```
dimitris-m: indentation

languages/elixir/generic/Elixir_to_generic.ml
maciejpirog: I guess it is not, but sometimes the recovery mechanism in tree-sitter produces weird syntax for incorrect programs, so I think it's always better to have something than to fail the entire to-ast-generic.

languages/elixir/generic/Elixir_to_generic.ml
maciejpirog: fixed

languages/elixir/generic/Elixir_to_generic.ml
maciejpirog: extracted two common subfunctions

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -958,15 +971,27 @@ and map_body env v : G.stmt list =
...
+  match (v1, v2, v3) with
+  (* https://hexdocs.pm/elixir/Kernel.SpecialForms.html#case/2
+   * case expr do pattern -> body end *)
+  | ( I (Id ("case", tcase)),
+    (_, ([ scrutinee ], []), _),
+    Some (_tdo, (Clauses clauses, []), _tend) ) ->
+      let e = map_expr env scrutinee in
+      let xs = map_clauses env clauses in
+      let cases = xs |> List_.map case_and_body_of_case_clause in
+      G.Switch (tcase, Some (G.Cond e), cases) |> G.s |> G.stmt_to_expr
```
dimitris-m: To be consistent, should this not be in languages/elixir/ast/Elixir_to_elixir.ml ?

It could require a new Elixir AST type etc, but this is how things are right now...

maciejpirog: Done, as a separate commit

dimitris-m (approved):
lgtm

Issue #657, opened 2026-04-14
Python parser fails on valid `match` statements with `Syntax error ... \`message\` was unexpected`


raphaeldavidf:
**Describe the bug**

`opengrep 1.19.0` fails to parse valid Python 3.11 `match` statements in at least two small repro cases. The same files are parsed successfully by `semgrep 1.156.0`.

Primary repro error:

```text
Syntax error at line repro.py:12:
 `message` was unexpected
```

Secondary repro error:

```text
Syntax error at line repro_event.py:2:
 `event` was unexpected
```

This was originally observed on `opengrep 1.16.5` and reproduces identically on `1.19.0`.

**To Reproduce**

Minimal repro 1 — save as `repro.py`:

```python
from contextlib import nullcontext


def collect() -> None:
    with (
        nullcontext() as config,
        nullcontext() as service,
    ):
        _ = config, service

    message = {"type": "PRE"}
    match message["type"]:
        case "PRE":
            return None
```

Run:

```bash
opengrep scan -q -e '$X' -l python --json-output out.json repro.py
```

Output:

```json
{
  "version": "1.19.0",
  "results": [],
  "errors": [
    {
      "code": 3,
      "level": "warn",
      "type": "Syntax error",
      "message": "Syntax error at line repro.py:12:\n `message` was unexpected",
      "path": "repro.py"
    }
  ]
}
```

Minimal repro 2 — save as `repro_event.py`:

```python
def run(event: dict[str, str]) -> None:
    match event["error"]:
        case "States.Timeout":
            return None
        case _ as error:
            raise RuntimeError(error)
```

Run:

```bash
opengrep scan -q -e '$X' -l python --json-output out.json repro_event.py
```

Output:

```json
{
  "version": "1.19.0",
  "results": [],
  "errors": [
    {
      "code": 3,
      "level": "warn",
      "type": "Syntax error",
      "message": "Syntax error at line repro_event.py:2:\n `event` was unexpected",
      "path": "repro_event.py"
    }
  ]
}
```

For comparison, the same command with `semgrep 1.156.0` parses both files cleanly and returns matches with no parser error:

```bash
semgrep scan -q -e '$X' -l python --json-output out.json repro.py
```

**Expected behavior**

Both files should parse successfully and produce normal findings for the trivial `$X` pattern, matching the behavior of upstream Semgrep.

**Screenshots**

N/A — JSON output included above.

**What is the priority of the bug to you?**

[...]

PR #665, opened 2026-04-16
python: parse PEP 634 match/case statements in the menhir parser

raphaeldavidf:
## Summary

The menhir-based Python parser (`languages/python/menhir/`) had no grammar rules for PEP 634 `match`/`case` statements. When it failed to parse such a file, the tree-sitter fallback would parse most of it but lose findings around the `match` block and emit a "partially analyzed" warning. This is the symptom reported in #657.

This PR adds native `match`/`case` support to the menhir parser. The tree-sitter fallback is no longer required for the common forms.

## Changes — three files in `languages/python/menhir/`

- **`Parsing_hacks_python.ml`** — adds a soft-keyword pre-pass (`rewrite_match_case`, called from `fix_tokens`) that rewrites `NAME("match")` / `NAME("case")` into dedicated `MATCH` / `CASE` tokens **only when the surrounding shape is unambiguously a match statement**. Plain identifiers (`match = 1`, `def case(x):`, `obj.match()`, dict keys, parameter names, decorator targets, etc.) keep working. Uses an indent-depth stack so nested `match` statements and `case`-as-identifier inside a case body are handled correctly.
- **`Parser_python.mly`** — declares `MATCH` / `CASE` tokens and adds `match_stmt` / `case_block` grammar rules. Patterns are represented as plain expressions (`AST_python.pattern = expr`), mirroring the tree-sitter frontend. Reuses the existing `Switch` / `Case` / `CasesAndBody` AST nodes — no AST changes. **Zero new grammar conflicts** (still 2 s/r, 4 r/r, identical to baseline). Also accepts `case pattern as name:` (PEP 634 as-pattern).
- **`Token_helpers_python.ml`** — visitor arms for the two new tokens.

## Validation

Built locally with OCaml 5.3.0 (`make minimal-build`) and exercised via the freshly built `opengrep` binary:

- Both repros from #657 now scan cleanly via the menhir parser, no tree-sitter fallback, no "partially analyzed" warning, all `return`/`raise` findings inside `case` bodies surfaced.
- Custom test harness covering: nested `match`, multi-element subjects (`match a, b:`), multi-pattern cases (`case 1, 2:`), `case _:` wildcard, `case _ as error:` as-pattern, `case [x] as pair:` sequence-with-as, `case 1 | 2 as n:` or-with-as, and `case` used as a variable name inside an `if` nested in a case body — all behave correctly.
- No-regression check on the existing `as` keyword usages: `import x as y`, `with x as y:`, `except E as e:` all still parse.
- Across the 435 `.py` files under `tests/`: **432 parse via menhir** (vs 430 on `main`). The 3 remaining failures are the same 3 pre-existing intentional ones (`tests/TODO/bad.py`, `tests/parsing_errors/unbalanced_brace.py`, `tests/patterns/python/cp_label.py`).
- Ran the patched binary against a ~1 MM LoC internal Python codebase — works as expected, no parse regressions observed.

## Known limitations (still routed to tree-sitter fallback)

- **`case pattern if guard:`** — would require non-trivial grammar reshuffling because an unadorned `if` after a `test` collides with the conditional-expression ternary `a if b else c`. Documented inline.
- **`as`-binding name dropped from AST** — preserving it would need an `AST_python.case` extension and a re-think of the as-scoping (PEP 634 scopes `as` to the rightmost element of a comma-separated pattern list, not the whole tuple). TODO noted in the rule.

### DIFF
--- a/languages/python/menhir/Parser_python.mly
+++ b/languages/python/menhir/Parser_python.mly
@@ -130,6 +130,10 @@ let mk_str ii =
  NONE TRUE FALSE
  ASYNC AWAIT
  NONLOCAL
+ (* python3.10+: soft keywords, produced by Parsing_hacks_python
+  * out of NAME("match")/NAME("case") only in match-statement contexts
+  * so that plain identifiers "match"/"case" remain valid names. *)
+ MATCH CASE
  (* python2: *)
  PRINT EXEC
 
@@ -550,6 +554,7 @@ compound_stmt:
   | for_stmt    { $1 }
   | try_stmt    { $1 }
   | with_stmt   { $1 }
+  | match_stmt  { $1 }
 
   | funcdef     { $1 }
   | classdef    { $1 }
@@ -638,6 +643,32 @@ with_inner_in_parens:
   | test AS expr ","                      { fun (t, body) -> With (t, ($1, Some $3), body) }
   | test         "," with_inner_in_parens { fun (t, body) -> With (t, ($1, None), [$3 (t, body)]) }
   | test AS expr "," with_inner_in_parens { fun (t, body) -> With (t, ($1, Some $3), [$5 (t, body)]) }
+
+(* python3.10+ (PEP 634): structural pattern matching.
+ * `match` and `case` are *soft keywords* in Python — they are valid
+ * identifiers outside of match-statement contexts. The MATCH/CASE tokens
+ * used here are synthesized by Parsing_hacks_python only when the
+ * surrounding shape is unambiguously a match statement.
+ * Patterns are represented as plain expressions (AST_python.pattern = expr),
+ * which matches what the tree-sitter Python frontend already does.
+ * Note: `case pattern if guard:` is not accepted here — an unadorned `if`
+ * after a test would collide with the conditional-expression ternary
+ * (`a if b else c`) and introduce a shift/reduce conflict. Files that rely
+ * on case guards will fall back to the tree-sitter parser. *)
+match_stmt:
+  | MATCH tuple(namedexpr_test) ":" NEWLINE INDENT case_block+ DEDENT
+      { Switch ($1, tuple_expr $2, $6) }
+
+case_block:
+  | CASE tuple(test_or_star_expr) ":" suite
+      { CasesAndBody ([Case ($1, tuple_expr $2)], $4) }
+  (* PEP 634 as-pattern `case pattern as name:`. TODO: preserve the binding.
+   * Note: PEP 634 scopes `as` to the rightmost element of a comma-separated
+   * pattern list, but here `as` binds to the whole tuple — which is fine
+   * because we discard the name; revisit if the binding is ever preserved.
+   *)
+  | CASE tuple(test_or_star_expr) AS NAME ":" suite
+      { CasesAndBody ([Case ($1, tuple_expr $2)], $6) }
 
 (* python3-ext: *)
 async_stmt:
--- a/languages/python/menhir/Parser_python.mly
+++ b/languages/python/menhir/Parser_python.mly
@@ -88,6 +88,13 @@ let mk_str ii =
   let s = Tok.content_of_tok ii in
   Str (s, ii)
 
+(* Match the tree-sitter Python frontend's shape: a match-statement subject
+ * and each case pattern are always wrapped in a Tuple, even for a single
+ * element, so semgrep matching stays uniform across parser backends.
+ * The Param context mirrors `no_ctx` in Parse_python_tree_sitter.ml. *)
+let match_tuple xs =
+  Tuple (CompList (Tok.unsafe_fake_bracket xs), Param)
+
 %}
 
 (*************************************************************************)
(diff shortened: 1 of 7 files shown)


dimitris-m:
The match/case grammar works, but it produces an AST shape that diverges from the tree-sitter Python frontend, and that divergence breaks matching across the parser boundary.

### Repro

Rule:
```yaml
rules:
  - id: match-one
    pattern: |
      match $X:
          case 1:
              $Y
    languages: [python]
    severity: WARNING
    message: "found"
```

Target A — pure match/case, parses via menhir:
```python
def f1(x):
    match x:
        case 1:
            print("one")
```
→ **1 finding**.

Target B — same `f1`, plus a function with `case 1 if x > 0:` that forces the whole file onto the tree-sitter fallback:
```python
def f1(x):
    match x:
        case 1:
            print("one")
def f2(x):
    match x:
        case 1 if x > 0:
            print("guarded one")
```
→ **0 findings** (`f1` is identical to Target A and should still match).

### Cause

`Parse_python_tree_sitter.ml:1443,1557` always wraps the subject and case patterns in `Tuple (CompList (fb (e :: es)), no_ctx)`, even for a single element. The new menhir productions route through `tuple_expr` (`Parser_python.mly:42-45`) which collapses `Single e -> e`. So:

- Pattern (menhir): `Switch(_, Cond(N "\$X"), [Case(_, PatLiteral(Int 1)) ...])`
- Target A (menhir): `Switch(_, Cond(N "x"),  [Case(_, PatLiteral(Int 1)) ...])` — matches.
- Target B (tree-sitter): `Switch(_, Cond(Container(Tuple, [N "x"])), [Case(_, PatTuple([PatLiteral(Int 1)])) ...])` — extra `Tuple` / `PatTuple` wrappers, no match.

Net effect: identical user code can now produce different findings depending on whether anything *else* in the file forces the tree-sitter fallback. Pre-PR every match-containing file went through tree-sitter, so behaviour was at least consistent.

### Suggested fix

In `match_stmt` and `case_block`, wrap the subject and the case pattern unconditionally in `Tuple (CompList (fb [...], Load))` instead of `tuple_expr \$2`, mirroring the tree-sitter shape.

raphaeldavidf:
Thanks for the careful review and the very clear A/B repro — that made the divergence easy to confirm and reproduce locally.

Pushed `d8f707c41` adopting your suggestion: added a small `match_tuple` helper in the prolog and routed the subject and both `case_block` arms through it, mirroring `Tuple (CompList (fb (e :: es)), no_ctx)` from the tree-sitter frontend (with `Param` to match `no_ctx`). The grammar comment is updated to call out the wrapping invariant.

Verified locally:
- Your Target A still returns 1 finding; Target B now returns 2 (the `f1` match you expected, plus `f2` because the tree-sitter frontend drops the `if` guard, so `case 1:` matches `case 1 if x > 0:` — pre-existing semgrep behavior, not introduced here).
- My earlier regression suite (basic repros, nested match, multi-element subjects, `case` as identifier inside an `if` body, as-pattern, plus `import/with/except … as`) still passes.
- Full Python corpus: same pass count as `main`, no new failures.

Happy to iterate further if anything else is off.

dimitris-m:
On `case pattern if guard:`: the conflict with the ternary `a if b else c` comes from `test` (`Parser_python.mly:942-945`) including `or_test IF or_test ELSE test`. `test_nocond` (`Parser_python.mly:1060-1062`) already exists — same as `test` minus the ternary — and is used in `comp_if` (`Parser_python.mly:1056-1058`) for exactly this reason: the comprehension `[x for x in xs if cond]` would otherwise collide with the ternary.

Same fix here: switch the case-pattern nonterminal from `tuple(test_or_star_expr)` to `tuple(test_nocond)` and add a guarded alternative to `case_block`:

```
case_block:
  | CASE tuple(test_nocond) ":" suite
      { ... }
  | CASE tuple(test_nocond) IF test ":" suite
      { ... }   (* new: guarded case *)
```

The dropped `star_expr` arm only allowed `case *xs:` at the top level, which isn't valid PEP 634 — starred patterns appear inside sequence patterns (`case [a, *rest]:`) and are parsed via the list-literal atom, not at the case_block level.

This can be parse-and-discard: the **tree-sitter Python frontend already drops the guard** today (`Parse_python_tree_sitter.ml:1433-1443` computes `cond` and never uses it because `AST_python.case = Case of tok * pattern` has no guard slot). Discarding the guard in the menhir grammar matches existing behaviour and removes the fallback for guarded files immediately.

Once this PR is merged, the proper fix is to extend `AST_python.case` to carry an `expr option` for the guard, plumb it through both parsers, and emit `G.PatWhen (pat, guard)` in `Python_to_generic.case` — same pattern as `ocaml_to_generic.ml`, `scala_to_generic.ml`, and `Parse_rust_tree_sitter.ml`.

dimitris-m:
Suggested regression test — two pairs in `tests/patterns/python/` (each `.py` needs its own `.sgrep` with matching base filename, content duplicated):

`match_cross_parser.py`:

```python
def f(x):
    # ERROR:
    match x:
        case 1:
            print("a")
```

`match_cross_parser.sgrep`:

```python
match $X:
    case 1:
        $Y
```

`match_cross_parser_fallback.py` (the guard forces the file onto tree-sitter):

```python
def f(x):
    # ERROR:
    match x:
        case 1:
            print("a")
def g(x):
    # ERROR:
    match x:
        case 1 if x > 0:
            print("b")
```

`match_cross_parser_fallback.sgrep` (identical to the first):

```python
match $X:
    case 1:
        $Y
```

dimitris-m (commented):
Once the comments are addressed, we can get this merged.

You do **not** need to implement this part in this PR:
> Once this PR is merged, the proper fix is to extend AST_python.case to carry an expr option for the guard, plumb it through both parsers, and emit G.PatWhen (pat, guard) in Python_to_generic.case — same pattern as ocaml_to_generic.ml, scala_to_generic.ml, and Parse_rust_tree_sitter.ml.

raphaeldavidf:
Thanks again for the thoughtful suggestions — the `test_nocond` trick was the missing piece for me, and the cross-parser fixtures make the regression scenario crystal-clear.

Pushed `1f4ef8229`:
- Switched the case-pattern non-terminal to `tuple(test_nocond)` and added the `IF test` arm. As you noted, the guard is parsed-and-discarded for now, mirroring `Parse_python_tree_sitter.ml:1433-1443`. Kept the as-pattern arms my earlier commit added, so all four PEP 634 combinations parse: bare, `if guard`, `as name`, and `as name if guard`.
- Dropping `star_expr` at the top of the case pattern fell out of the `test_nocond` switch — agreed it tightens the grammar without losing anything valid.
- Added the two fixture pairs (`match_cross_parser` and `match_cross_parser_fallback`) under `tests/patterns/python/`, with `# ERROR:` markers placed per `Test_compare_matches.ml:48-55`. Both run green locally.
- Documented the deferred AST extension (extend `AST_python.case` with `expr option`, plumb through both parsers, emit `G.PatWhen`) as a TODO in the rule comment, citing `ocaml_to_generic`, `scala_to_generic`, and `Parse_rust_tree_sitter` as prior art per your suggestion.

Conflict count is still 2 s/r / 4 r/r baseline. Happy to address anything else.

PR #673, opened 2026-04-20
Exclude more synthetic identifiers from the PCRE prefilter

dimitris-m:
Follow-up to #671. The same prefilter leak affected three other encodings of synthetic identifiers.

- **Scala** `cases_to_lambda` injected `!hidden_scala_param!` for `{ case … }` lambdas. Switched to the existing `AST_generic.implicit_param`, already excluded via `Pattern.is_special_identifier`.
- **Kotlin** injected `it` as a synthetic pname for parameterless lambdas. Marked its `pinfo` hidden.
- **Elixir** `^` pin operator already built a hidden `PatId "__tmp"`, but the prefilter visitor only consulted the hidden flag on `N (Id _)` names. Extended `Analyze_pattern.ml` to also honour it on `Param` and `PatId`.

Each case has a fixture under `tests/rules/` demonstrating a file the prior binary skipped entirely.


corneliuhoffman (approved):
LGTM

PR #677, opened 2026-04-20
hide synthetic idents, add prefix-metavar matching, fix go chan direction

### LINKED ISSUE #678: Prefilter / matcher layer: residual gaps for setter-shaped patterns
Context

  Follow-up to #677. After the cleanup of parser-synthesised identifiers leaking into the PCRE prefilter, three connected gaps remain. They all surface around
  C#/Apex property accessors and around the overloading of the hidden id-flag.

  1. value / field pnames in: C# and Apex setters

  The C#/Apex parsers inject synthetic parameters named value (both langs) and field (C#14) into setter/property AST. Those words do appear in source when the
  accessor body uses them, but not always — set { _x = 5; } is legal and contains neither.

  We deliberately left the pnames non-hidden in #677  because Eval_generic.text_of_metavariable gates on IdFlags.is_hidden, and marking them hidden would break
  metavariable-regex: value / : field rules (e.g. tests/tainting_rules/csharp/implicit_vars_in_props.yaml). The prefilter therefore still rejects any file
  scanned by set { $S = $V; }-style patterns whose source setter body happens not to mention value.

  2. Prefix-metavariable decomposition in Ids

  Even with the prefilter cleared, patterns like:
```csharp
  class $C {
      public int $P {
          get { return $G; }
          set { $S = $V; }
      }
  }
```

  3. C++ anonymous-type handling

  _name_option in languages/cpp/generic/cpp_to_generic.ml:85 constructs a _ANON placeholder for anonymous structs/classes/unions. The function is currently
  dead code (no callers; the leading underscore suppresses the unused-binding warning). Before hooking it up to actually model anonymous C++ types, it should:

  - Emit the Id with ~hidden:true so the prefilter doesn't demand _ANON in source.
  - Use a name convention consistent with the new prefilter_hidden flag (item 1) or the !!_implicit_param! prefix.

  If no near-term work needs anonymous C++ type support, consider deleting the function to remove the foot-gun.

corneliuhoffman:
## Summary

  Follow-up to #673. The PCRE prefilter rejects files whose source text doesn't contain strings the pattern parser has synthesised into the AST but that never  appear in source. This PR covers the remaining parser-synthesised idents we hadn't addressed, adds the matcher support needed so mangled accessor names still work as pattern targets, and fixes two latent bugs found along the way.

## Covered — synthetic idents now marked hidden

  - Go — send / recv / bidirectional synthesised from chan<- T / <-chan T / chan T.
  - Apex — __trigger_events and __trigger_object synthesised for trigger X on Y (events).
  - Swift — Dictionary synthesised when [K: V] literals desugar to Dictionary<K, V>.
  - C# — System and Index synthesised for arr[^N] from-end indexing.
  - C++ — DecltypeId synthesised for decltype(E)::foo qualified names (renamed to !!_implicit_param!decltype so is_implicit_param already filters it).
  - Elixir — __pN__ synthesised for multi-clause def lowering. Replaces the earlier is_elixir_synthetic_param string-hack in Pattern.ml with the proper hidden plumbing via param_of_id ~hidden:true.

 ## Matcher addition: prefix-metavariable in idents
This fixes #678:
C# and Apex property accessors are synthesised with CLR-style names: public int X { set { … } } becomes an entity named set_X. That naming is load-bearing for intrafile taint signature uniqueness and call-graph resolution within a class, so we keep it. But the matcher previously treated a pattern ident set_$P as one opaque string rather than set_ + metavar $P, so patterns targeting property accessors never matched source.

  Generic_vs_generic now has a match_prefix_metavar helper used by m_ident and m_ident_and_id_info: when the pattern ident is shaped <prefix>$METAVAR and the source ident starts with <prefix>, bind the metavar to the remainder. Pattern set_$P binds $P = "X" against source set_X.

  The accessor entity is still built with ~hidden:true so the mangled string doesn't leak into the prefilter regex. Applies uniformly to C# property braces, C# arrow-property (public int P => 3;), and Apex property accessors.

 ## Related fixes

  - Go tree-sitter parser direction swap. chan<- T was labelled TRecv and <-chan T was labelled TSend. The menhir parser had it right per the Go spec (chan<- is send-only, <-chan is receive-only); tree-sitter was inverted. Fixed so source and pattern AST agree. Rules keying on chan direction will shift after this lands.
  - C++ _name_option / _ANON dead code. The function at cpp_to_generic.ml:85 built a _ANON placeholder for anonymous C++ types but had no callers (the leading underscore silenced the unused-value warning). Deleted — no observable behaviour change. Any future support for anonymous C++ types should build the ident  with a hidden id_info.

###  Why not hidden on value / field pnames

  Eval_generic.text_of_metavariable gates metavariable-regex evaluation on IdFlags.is_hidden, so marking the C#/Apex setter's synthetic value pname — or C#'s synthetic field pname — hidden would break existing rules like tests/tainting_rules/csharp/implicit_vars_in_props.yaml that use metavariable-regex: value or metavariable-regex: field. Those words do appear in source text when the accessor body uses them, so we instead add scoped entries to Pattern.is_special_identifier (lang-gated on C#/Apex) that skip them from the prefilter string set without touching is_hidden.

 ## Plumbing

  - AST_generic_helpers.name_of_id / name_of_ids and AST_generic.param_of_id gain a ?hidden parameter.
  - Analyze_pattern.ml visit_name now short-circuits on IdQualified whose name_info is hidden (used by the C# System.Index fix).
  - Generic_vs_generic: new match_prefix_metavar helper shared by m_ident and m_ident_and_id_info.

 ## Tests

  New tests/rules/*_prefilter.{yaml,<ext>} fixtures covering every case above: Go chan directions (send/recv/bidir), Apex triggers (events + object), Swift [K: V], C# arr[^N], C++ decltype(E)::foo, Elixir multi-clause def, C# arrow-property + brace property + auto-property, Apex property setter. 16 tests, all pass.

### DIFF
--- a/languages/cpp/generic/cpp_to_generic.ml
+++ b/languages/cpp/generic/cpp_to_generic.ml
@@ -254,10 +254,13 @@ and map_qualifier env = function
          It's like a dynamic IdQualified, really more suited for something
          like a DotAccess of a Call to `decltype`, for programs like
          decltype(e)::foo
-         Let's just make something up for now.
+         We use the `!!_implicit_param!` prefix so the prefilter's
+         is_implicit_param check skips this synthesised ident.
       *)
       let _v1 = map_tok env v1 and _v2 = map_bracket env (map_expr env) v2 in
-      let v1 = ("DecltypeId", G.fake "DecltypeId") in
+      let v1 =
+        (G.implicit_param ^ "decltype", G.fake (G.implicit_param ^ "decltype"))
+      in
       (v1, None)
 
 and map_a_class_name env v = map_name env v
--- a/languages/cpp/generic/cpp_to_generic.ml
+++ b/languages/cpp/generic/cpp_to_generic.ml
@@ -82,14 +82,6 @@ let expr_option t eopt =
   | Some e -> e
   | None -> G.L (G.Unit t) |> G.e
 
-let _name_option t nopt =
-  match nopt with
-  | Some n -> n
-  | None ->
-      (* TODO? gensym? *)
-      let fake_id = ("_ANON", t) in
-      H.name_of_id fake_id
-
 let distribute_access (xs : (G.field, G.attribute) Either.t list) : G.field list
     =
   let rec aux attr_opt xs =
--- a/languages/cpp/generic/cpp_to_generic.ml
+++ b/languages/cpp/generic/cpp_to_generic.ml
@@ -251,7 +251,7 @@ and map_qualifier env = function
       *)
       let _v1 = map_tok env v1 and _v2 = map_bracket env (map_expr env) v2 in
       let v1 =
-        (G.implicit_param ^ "decltype", G.fake (G.implicit_param ^ "decltype"))
+       let impl_param = G.implicit_param ^ "decltype" in (impl_param, G.fake impl_param)
       in
       (v1, None)
 
--- a/languages/csharp/generic/Parse_csharp_tree_sitter.ml
+++ b/languages/csharp/generic/Parse_csharp_tree_sitter.ml
@@ -3267,15 +3267,16 @@ and declaration ?(this_param=None) (env : env) (x : CST.declaration) : stmt =
               accs
               |> List_.map (fun (attrs, id, fbody) ->
                      let iname, itok = id in
-                     (* Use the source keyword (`get`/`set`/`init`) as
-                        the accessor entity name rather than the CLR-style
-                        `get_X`/`set_X`. The mangling broke pattern
-                        matching of `set { ... }` against `$P { set; }`
-                        because `set_$P` is treated as one opaque ident
-                        instead of `set_` + metavar `$P`, and it's not
-                        needed for taint analysis at this layer. *)
-                     let _fname = fname in
-                     let ent = basic_entity (iname, itok) ~attrs in
+                     (* CLR-style `get_<PropName>` / `set_<PropName>`;
+                        the prefix-metavar support in m_ident (see
+                        Generic_vs_generic.decompose_prefix_metavar)
+                        lets patterns like `set_$P` match these. Mark
+                        hidden so the prefilter regex doesn't demand
+                        the literal mangled string. *)
+                     let ent =
+                       basic_entity ~hidden:true
+                         (iname ^ "_" ^ fname, itok) ~attrs
+                     in
                      let valparam =
                        Param
                          {
@@ -3510,10 +3511,12 @@ and declaration ?(this_param=None) (env : env) (x : CST.declaration) : stmt =
                        else G.KeywordAttr (G.Setter, itok)
                   in
                   let attrs = attrs in
-                  (* Keep `get`/`set`/`init` as the entity name — see
-                     the other accessor site above for rationale. *)
-                  let _fname = fname in
-                  let ent = basic_entity (iname, itok) ~attrs in
+                  (* CLR-style accessor name; see other accessor site
+                     above for rationale. *)
+                  let ent =
+                    basic_entity ~hidden:true
+                      (iname ^ "_" ^ fname, itok) ~attrs
+                  in
                   let itok_loc = Tok.unsafe_loc_of_tok itok in
                   let new_loc loc n =
                       Tok.({
@@ -3575,12 +3578,15 @@ and declaration ?(this_param=None) (env : env) (x : CST.declaration) : stmt =
             ((open_br, funcs, close_br), v2)
         | `Arrow_exp_clause_SEMI (v1, v2) ->
             (* public int SomeProp => 3;
-             * Convert it to `get_SomeProp { return 3; }`
-             *)
+             * Convert it to `get_SomeProp { return 3; }` — CLR-style
+             * accessor name; patterns use `set_$P` / `get_$P` and the
+             * prefix-metavar support in m_ident binds the metavar.
+             * Hidden so the prefilter regex doesn't demand the mangled
+             * string literally. *)
             let v1 = arrow_expression_clause env v1 in
             let v2 = token env v2 (* ";" *) in
             let arrow, expr = v1 in
-            let ent = basic_entity ("get_" ^ fname, arrow) in
+            let ent = basic_entity ~hidden:true ("get_" ^ fname, arrow) in
             let funcdef =
               FuncDef
                 {
--- a/languages/csharp/generic/Parse_csharp_tree_sitter.ml
+++ b/languages/csharp/generic/Parse_csharp_tree_sitter.ml
@@ -272,8 +272,13 @@ let linq_to_expr (from : linq_query_part) (body : linq_query_part list) =
   | _ -> raise Impossible
 
 let new_index_from_end tok expr =
+  (* `arr[^N]` is desugared to `new System.Index(N, true)`. Neither
+     `System` nor `Index` appears in the source `^N` syntax, so mark
+     the synthesised qualified name hidden so the prefilter regex
+     doesn't demand them. *)
   let name =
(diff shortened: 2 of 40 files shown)


dimitris-m (commented):
### 1. Accessor naming inconsistency in C# arrow-expression properties

`Parse_csharp_tree_sitter.ml:3583` — for `public int SomeProp => 3;` the entity is still `get_SomeProp`:

```ocaml
| `Arrow_exp_clause_SEMI (v1, v2) ->
    (* public int SomeProp => 3;
     * Convert it to `get_SomeProp { return 3; }`
     *)
    ...
    let ent = basic_entity ("get_" ^ fname, arrow) in
```

The rationale for dropping `get_X`/`set_X` mangling (`set_$P` is treated as one opaque ident instead of `set_` + `$P`) applies here. `prop { get; set; }` and `prop => expr;` now produce different entity names; a pattern matching one misses the other. Indexer accessors at L3347/L3369/L3388 (`get_Item`/`set_Item`, kept for operator dispatch) should stay as-is.

### 2. `let _fname = fname` workaround

`Parse_apex_tree_sitter.ml:1689` and `Parse_csharp_tree_sitter.ml` at ~L3277, ~L3308, ~L3508. Cleaner options:

- Rebind the outer destructuring: `let _fname, _ftok = id in`.
- Or `ignore fname` if the binding is kept for readability.

### 3. Positive-only test fixtures

Each new `tests/rules/*_prefilter.{yaml,<ext>}` has a single `// ruleid:` and no `// ok:` counterpart. A regression that lets every file through would still pass. Each fixture should include a sibling fragment that must not match.

### 4. Title

`pcre fixes and tests` doesn't cover two of the changes:

- `Parse_go_tree_sitter.ml:1100-1115` — chan direction swap. The menhir parser already had `chan<- = TSend, <-chan = TRecv`; tree-sitter was inverted. Rules keying on chan direction will shift after this lands.
- C#/Apex accessor naming change is a matcher fix, not a prefilter fix.

Something like `prefilter: cover remaining synthetic idents (apex, c#, go, swift, c++, elixir); fix go chan direction; drop c#/apex accessor name mangling` would reflect the actual scope.

### 5. `name_of_ids ~hidden:true` only flags the `IdQualified` envelope

`AST_generic_helpers.ml:103-119` — `~hidden` is passed to `empty_id_info` for the wrapping `name_info`, but the inner qualifier idents in `name_middle` (e.g. `System` in `System.Index`) are constructed via `(id, None)` with no hidden flag. The new `Analyze_pattern.visit_name` stops recursion at the IdQualified, so this path is fine. Any other visitor that walks `name_middle` directly will see the qualifier idents as non-hidden. Two options:

- Mark the inner ids hidden too.
- Document on `name_of_ids` that the hidden flag travels only via the `IdQualified` envelope.

### 6. Issue #678 item 3

`cpp_to_generic.ml:85` — `_name_option` builds a `_ANON` placeholder for anonymous C++ types and currently has no callers. If reactivated without `~hidden:true`, the prefilter will reject valid C++. Delete it in this PR (4 lines) or mark hidden now.

### 7. Trivial

`AST_generic.ml:2380-2381` — stray double blank line after `is_elixir_synthetic_param` was removed.

languages/csharp/generic/Parse_csharp_tree_sitter.ml
```diff
@@ -3262,16 +3267,26 @@ and declaration ?(this_param=None) (env : env) (x : CST.declaration) : stmt =
...
-                     in
+                     (* Use the source keyword (`get`/`set`/`init`) as
+                        the accessor entity name rather than the CLR-style
+                        `get_X`/`set_X`. The mangling broke pattern
+                        matching of `set { ... }` against `$P { set; }`
+                        because `set_$P` is treated as one opaque ident
+                        instead of `set_` + metavar `$P`, and it's not
+                        needed for taint analysis at this layer. *)
+                     let _fname = fname in
+                     let ent = basic_entity (iname, itok) ~attrs in
```
maciejpirog: I don't think it's safe to name all getters `get` snd all setters `set`. We don't support properties in intrafile tainting, but one day we might, and with this solution we wouldn't have means to connect the name of the property at the call site with its implementation

corneliuhoffman:
Not sure I understand the positive only test fixture issue. EVen if I had som ok tests, a regression that lets all tetss pass will still not find those

corneliuhoffman:
fil fix tomorrow morning

dimitris-m:
> Not sure I understand the positive only test fixture issue. EVen if I had som ok tests, a regression that lets all tetss pass will still not find those



> Not sure I understand the positive only test fixture issue. EVen if I had som ok tests, a regression that lets all tetss pass will still not find those

ignore, you are right

languages/cpp/generic/cpp_to_generic.ml
```diff
@@ -254,10 +246,13 @@ and map_qualifier env = function
...
          like a DotAccess of a Call to `decltype`, for programs like
          decltype(e)::foo
-         Let's just make something up for now.
+         We use the `!!_implicit_param!` prefix so the prefilter's
+         is_implicit_param check skips this synthesised ident.
       *)
       let _v1 = map_tok env v1 and _v2 = map_bracket env (map_expr env) v2 in
-      let v1 = ("DecltypeId", G.fake "DecltypeId") in
+      let v1 =
+        (G.implicit_param ^ "decltype", G.fake (G.implicit_param ^ "decltype"))
```
dimitris-m:
```
       let impl_param = G.implicit_param ^ "decltype" in (impl_param, G.fake impl_param)
```

tests/rules/csharp_arrow_property_prefilter.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - id: csharp-arrow-property
+    pattern: |
+      class $C {
+          public int $P => $E;
+      }
+    languages: [csharp]
+    message: >
+      Arrow-property `public int P => E;`. Previously this was parsed
+      with a CLR-mangled `get_P` accessor entity name; now uses `get`
+      so $P can bind as a metavariable.
```
dimitris-m: This should be updated now after the last commit?

corneliuhoffman: removed the extra comment

dimitris-m (commented):
### Gate `match_prefix_metavar` on the pattern-side hidden flag

The accessor entity on both sides is built with `~hidden:true`. In `m_ident_and_id_info`, replace the current guard with:

```ocaml
| (stra, _), b1
  when String.contains stra '$'
       && IdFlags.is_hidden !(a2.G.id_flags) -> (
    match match_prefix_metavar stra b1 with
    | Some tin -> tin
    | None -> ...)
```

and drop the case from `m_ident` (no id_info there; entity comparisons flow through `m_ident_and_id_info`).

### VB.NET has the same issue and isn't handled

In `languages/vbnet/Vbnet_parser.ml`, the Get and Set accessor branches build entities `GET_<PropName>` / `SET_<PropName>` via `make_name` — no `hidden` flag, no parameter for one. VB source has `Get`/`Set` keywords separately from the property name, so `GET_Bar` / `SET_Bar` never appear literally in source. Same prefilter rejection as C#/Apex.

The minimal fix is at the two accessor sites:

```ocaml
G.EN (G.Id (("GET_" ^ property_name_str, get.tok),
            G.empty_id_info ~case_insensitive:true ~hidden:true ()))
```

(and the `SET_` branch). Or extend `make_name` with `?hidden`.

I suggest a new test `tests/rules/vbnet_prop_setter_prefilter.{vb,yaml}` to verify the fix.

dimitris-m (approved):

maciejpirog (approved):

PR #684, opened 2026-04-22
Taint: guarded signatures, field sensitivity, improved lowering to IL, better HOF callback detection

dimitris-m:
## What this does

Improvements to cross-function (intrafile) taint analysis.

**Effect guards.** Each taint effect carries a boolean condition that must hold at the caller. At a call site the actual arguments are substituted into the condition and it is partially evaluated: an effect whose guard folds to `false` is dropped; an undecidable guard is kept. Recognised conditions are anchored on a parameter — truthiness (`if flag`), equality to a literal (`if code == 0`), `length(path) == n`, nested-field truthiness, and the same conditions reached through a literal bound to a local. Guards apply to sinks, returns, and l-value writes, and rebind through forwarding chains and higher-order dispatch, including a callback whose return is guarded. A guard is dropped once its variable is reassigned on the path (the IL is not in SSA form, so `if v: …; v = …; if not v: sink(x)` is not treated as a contradiction). Clojure multi-arity dispatch reuses this: one signature per function with a `length(impl) == n` guard per leg, instead of N split signatures.

```python
def consumer(opts, x):
    if len(opts["data"]) == 2:
        sink(x)
consumer({"data": [1]},    source())   # no finding
consumer({"data": [1, 2]}, source())   # finding
```

**Field-sensitive taint** through destructured map/record parameters (Clojure, Elixir, JS/TS, Rust), parameter-pattern HOF callbacks (`#(… %)`, `fn %{k: v} ->`, `|(a, _)|`, `|(a, b)|`), Clojure keyword-as-function (`(:body m)`), and library map accessors (`getattr`, `Map.get`/`get_in`, `fetch`/`dig`/`send`, dictionary indexers, …) across Python, Elixir, Ruby, Java, C#, Kotlin, Scala, and Go. Taint routes per key rather than to the whole parameter.

**IL lowering.** Switch / `if let` cases lower to a structural condition (so constant-false branches prune); a new `Slice` offset for trailing-rest destructure (`[a, b, & rest]` / `...rest` / `[h | t]`); Rust struct literals lower as `RecordOrDict` so caller and callee field offsets line up; destructured fields carry a `Dot` offset.

**Multi-arity self-recursion** resolves in both directions (lower-arity calling higher and vice versa) via an outer fixpoint over signature extraction — Clojure-gated, bounded by `taint_MAX_SELF_SIG_PASSES`, and it warns if the cap is hit.

**Unreachable-branch pruning.** A branch whose condition constant-folds against its direction marks the environment dead; the dead environment is discarded at the Join, so observations inside it do not survive.

**Symbolic propagation** extended to sequence/dict/record containers and lambda expressions; `length(x)` folds on literal containers and strings — this is what makes the arity and `len(…)` guards fire.

Polymorphic `Arg`-shape offsets are length-capped (`taint_MAX_POLY_OFFSET`) so deep forwarding chains don't degrade shape-comparison performance.


corneliuhoffman:
I think these fail:
clojure
```clojure
(defn f [a & rest]
  (sink rest))

(defn caller-rest-pos1 []
  (f "safe" "ok" (source)))
```

javascript
```javascript
// /tmp/var-js.js
  function f(a, ...rest) {
      sink(rest);
  }

  // All three should fire — none do.
  function callerPos1() { f("safe", source()); }
  function callerPos2() { f("safe", "ok", source()); }
  function callerInline() { f("safe", source(), "x"); }
  ```
  
 elixir:
 ```elixir
   defmodule M do
    def f([head | _]), do: sink(head)        # head should be source — 0 findings
    def g([_ | tail]), do: sink(tail)        # tail should contain source — 0 findings
    def caller_head(), do: f([source(), "ok"])
    def caller_tail(), do: g(["safe", source()])
  end
  ```

**Comment by corneliuhoffman** [comment]
the more sophisticated guards fail in ToReturn and Lval

ToReturn
```python
  def inner(opts):
      if len(opts["data"]) == 2:
          return source()
      return ""

  def outer(p):
      return inner(p)              # forwards p; should propagate inner's guard

  def caller():
      sink(outer({"data": [1]}))   # length 1 ≠ 2 — should NOT fire
```

Lval

```python
class C: pass

  def tagger(obj, flag, val):
      if flag:
          obj.x = val             # guarded write

  def caller():
      c = C()
      tagger(c, False, source())  # flag false — should NOT taint c.x
      sink(c.x)
```

src/tainting/Dataflow_tainting.ml
```diff
@@ -2815,6 +3267,7 @@ and fixpoint_aux taint_inst func ?(needed_vars = IL.NameSet.empty)
       lval_env = enter_lval_env;
       needed_vars;
       effects_acc = ref Effects.empty;
+      did_self_recurse = ref false;
```
corneliuhoffman, line 3455: shoudl this be a ref? In fact
  1. effects_acc is only ever extended (set-union) during a pass — never reset, replaced, or shrunk.
  2. did_self_recurse is reset at pass start and only ever set, never cleared, by the transfer.
  3. Neither ref is touched by code outside this loop's control during a pass.

Suggestion:
 Make all cross-pass state explicit:

    - effects_acc becomes pass-local — fresh allocation per pass, never read across passes.
    - Add a read-only self_sig_seed : Effects.t field on env. self_sig_if_recursive reads from it instead of
  effects_acc.
    - did_self_recurse is dropped as a field; the inner pass returns a did_recurse : bool value alongside the end
  mapping.
    - run_to_sig_fixpoint becomes value-passing: prev_effects → (new_effects, did_recurse), with the outer loop
  deciding to recurse based on the returned flag and Effects.equal prev_effects new_effects.

tests/sarif/targets/taint/taint.py
```diff
@@ -1,7 +1,7 @@
 def foo():
     x = 'test string format {}'.format('foo')
     a = source1()
-    if True:
```
corneliuhoffman, line 4: why? explain?

tests/tainting_rules/python/tainting_top.py
```diff
@@ -4,5 +4,5 @@
     b = sanitize1()
 else:
     b = a
-#ruleid: tainting
```
corneliuhoffman, line 7: why did this happen?

src/tainting/Dataflow_tainting.ml
```diff
@@ -2576,21 +2869,146 @@ let mk_lambda_in_env env lcfg =
...
+  lval_env
+
+(* At [TrueNode] / [FalseNode], if [cond] evaluates to a constant boolean that
+ * contradicts the branch direction, the branch is unreachable and every
+ * tracked l-value on the incoming env is marked [`Clean]. The pruned branch
+ * still executes, but its reads see no taint, so sinks inside it cannot fire;
+ * at the Join, [Xtaint.union (Tainted, Clean) = Tainted] preserves the live
+ * branch's taints. Applies to any constant-folded condition, e.g. [if (true)]
+ * or [if (length(x) == 1)] when the shape of [x] is statically known. *)
+let prune_branch_if_unreachable (lang : Lang.t) (cond : IL.exp)
```
corneliuhoffman, line 3130: The pruner's reach is determined by what Eval_il_partial.eval can fold, which in turn depends on what Dataflow_svalue.is_symbolic_expr propagates. The two mechanisms are coupled through id_svalue but neither file references the other, and there's no test that pins their composed surface. Extending is_symbolic_expr (e.g. adding G.Cast or another container) silently widens what conditions the pruner can fold — possibly desirable, but unlikely to be deliberate without a test. Maybe a comment at is_symbolic_expr noting that changes affect branch pruning, and a fixture that asserts the prunable surface for each kind currently in the match. And maybe some tests?

src/tainting/Dataflow_tainting.ml
```diff
@@ -2576,21 +2869,146 @@ let mk_lambda_in_env env lcfg =
...
+               (i + 1, lval_env)
+           | IL.Param _
+           | IL.ParamRest _
+           | IL.ParamFixme ->
+               (i + 1, lval_env))
+         (0, lval_env)
+  in
+  lval_env
+
+(* At [TrueNode] / [FalseNode], if [cond] evaluates to a constant boolean that
```
corneliuhoffman: I think this is a bit overstated. For example

```python
  # (b) Fresh taint produced INSIDE the dead branch.
  def case_fresh_in_branch():
      if False:
          sink(source())

  # (c) Same shape, with an intermediate local.
  def case_fresh_via_local():
      if False:
          x = source()
          sink(x)
```

are both found while I guess they should not?

src/tainting/Taint_shape.ml
```diff
@@ -170,24 +197,33 @@ and unify_shape shape1 shape2 =
...
+       * common prefix of the two paths. If the paths agree we keep the full
+       * offset; if they diverge we retain whatever they share and forget
+       * the rest. *)
+      let rec longest_common_prefix l1 l2 =
+        match (l1, l2) with
+        | x :: xs, y :: ys when T.equal_offset x y ->
+            x :: longest_common_prefix xs ys
+        | _ -> []
+      in
+      Arg (arg1, longest_common_prefix off1 off2)
```
corneliuhoffman: something funny happens here, for example
```python
def my_hof(opts, flag):
    if flag:
        cb = opts["a"]
    else:
        cb = opts["b"]
    return cb(opts["data"])

def handler_a(x):
    # todoruleid: lcp-unify-fn
    sink(x)

def handler_b(x):
    # todoruleid: lcp-unify-fn
    sink(x)

def caller_true():
    my_hof({"a": handler_a, "b": handler_b, "data": source()}, True)

def caller_false():
    my_hof({"a": handler_a, "b": handler_b, "data": source()}, False)
```

finds nothing but if I put

```python 
def my_hof(opts, flag):
    if flag:
        cb = opts["a"]
    else:
        cb = opts["a"]
    return cb(opts["data"])
```
(same value for cb) one finds 2 taints.

src/configuring/Limits_semgrep.ml
```diff
@@ -69,11 +69,28 @@ let taint_MAX_OBJ_FIELDS = 10
  *)
 let taint_MAX_TAINT_SET_SIZE = 25
 
-(** Bounds the length of the offsets we can track per arg/poly-taint. *)
-let taint_MAX_POLY_OFFSET = 1
+(** Bounds the length of the offsets we can track per arg/poly-taint.
```
corneliuhoffman, line 72: ;; Clojure 3-level destructure exceeds MAX_POLY_OFFSET = 2 because the
;; implicit_param + Switch encoding contributes an extra [Oint 0] arity
;; layer on top of the three keyword-field offsets, total path length 4.
;; Truncation collapses sibling fields at depth 3.
;;
;; Caller binds source() to :other (not :body); sink reads :body.
;; Field-sensitive should produce 0 findings. Currently produces 1.

(defn handler [{outer :outer}]
  (let [{middle :middle} outer
        {body :body} middle]
    ;; todoruleid: clj-destruct-depth3-fp
    (sink body)))

(defn caller-neg-depth3 []
  (handler {:outer {:middle {:body "safe" :other (source)}}}))

languages/elixir/generic/Elixir_to_generic.ml
```diff
@@ -85,12 +85,34 @@ let expr_of_quoted (quoted : quoted_generic) : G.expr =
...
+ * [body] in both [keyval_of_pair] (expression-side map literals) and
+ * [expr_to_pattern] below (pattern-side map destructures). This keeps
+ * caller-side [%{body: v}] and callee-side [%{body: v}] projecting the
+ * same offset. *)
+let strip_keyword_suffix (id : string AST_generic.wrap) :
+    string AST_generic.wrap =
+  let s, tok = id in
+  (String_.strip_wrapping_char ':' s, tok)
+
 let keyval_of_pair p : G.expr =
```
corneliuhoffman, line 100: some recursion here that needs to be noted:

```elixir
defmodule M do
  # Pre-PR: the keyword `body:` parsed with key as N(Id "body").
  # Post-PR (commit 3093b09a3 + Elixir parser changes): key is now
  # L(String "body"). Rules constraining $K via metavariable-pattern with
  # an identifier-pattern silently break — they used to match, now produce
  # 0 findings.
  def m1(), do: %{body: "x"}
  def m2(), do: %{user: "y"}
end
```

for the rule
```yaml
rules:
  - id: elixir-key-id-pattern
    message: |
      Rule constrains $K via metavariable-pattern with a bare-identifier
      pattern. Pre-PR this matched; post-PR $K binds to a string literal
      and the identifier-pattern fails. Expected behaviour pre-PR: 1
      finding on m1. Actual post-PR: 0 findings.
    languages: [elixir]
    severity: WARNING
    patterns:
      - pattern: |
          %{$K: $V}
      - metavariable-pattern:
          metavariable: $K
          pattern: body
```

src/naming/Naming_AST.ml
corneliuhoffman, line 1007: The new ParamPattern of pattern * parameter_classic shape was specifically designed to retire several pre-existing workarounds:
  
  - Parse_typescript_tree_sitter.ml:1404, 2018 — type still wrapped in Cast (pat, colon_tok, type_) instead of lifted  to parameter_classic.ptype.
  - Python_to_generic.ml:486 — ParamPattern (PatternTuple pat, _) still discards topt.
   - Parse_rust_tree_sitter.ml:2438 — type still stuffed in PatTyped(pat, ty) instead of ptype.
   - ruby_to_generic.ml:317 — OtherPat(("ParamPattern", …)) workaround still present.
   - This site (the JS-guard) — TODO comment notes "we should have a pattern PatKeyVal where key is PatId(foo)", which is what the new shape's destructuring lowering produces. Whether the guard is now dead code or still load-bearing  isn't clear.
 
 None of these were retired in the PR.

tests/sarif/targets/taint/taint.py
dimitris-m: the SARIF test expects a finding to match the stored output, now with True `b` was sanitised and the else was not considered at all, hence no finding.

tests/tainting_rules/python/tainting_top.py
dimitris-m: the guards ensure that the else's effects are discarded and the if-branch has no taint flow from a to b to sink

languages/elixir/generic/Elixir_to_generic.ml
dimitris-m: I think this now works as expected

corneliuhoffman (approved):
I think this went throught so much testing and checking that needs to me merged

maciejpirog (approved):

PR #693, opened 2026-04-27
signatures fixed point for mutually recursive functions

corneliuhoffman:
### Problem

  `Match_tainting_mode` walks the relevant call subgraph in a single
  topological pass. OCamlGraph's topo fold doesn't iterate cycles, so
  mutually recursive functions get analysed once in arbitrary order —
  at least one cycle member sees its mutual callee with no signature
  and produces an incomplete one. 

  ### Fix

  Two phases:
  1. **Signature fixpoint per SCC.** Decompose into SCCs (callees
     first, via `List.rev (SCC.scc_list)`), iterate signature
     extraction over each SCC until `SignatureSet.equal` stabilises.
     No match emission.
  2. **Single match-emission pass.** Walk the converged DB and run
     the existing `run_check_fundef_if_needed`.

  New `src/call_graph/Graph_fixpoint.{ml,mli}` — a generic functor
  over `GRAPH × LATTICE × STORE`, with `max_iter` and `widen` hooks.
  Singleton-no-self-loop SCCs run `analyze` once (current behaviour).

  ### Tests

  `tests/rules/cross_function_tainting/test_recursion_*.{js,ts,py,java,go}`
  — 3-cycle `p→q→r→p` (p forwards, q returns `r(source())`, r
  identity-passes). Verified to fail on baseline and pass with the
  fixpoint. JS also has a `counter` self-recursion case with a
  comment noting it'll discriminate once guarded effects land.

### DIFF
--- a/src/call_graph/Graph_fixpoint.ml
+++ b/src/call_graph/Graph_fixpoint.ml
@@ -1,0 +1,82 @@
+module type LATTICE = sig
+  type t
+
+  val equal : t -> t -> bool
+end
+
+module type GRAPH = sig
+  type t
+
+  module V : sig
+    type t
+
+    val compare : t -> t -> int
+    val equal : t -> t -> bool
+    val hash : t -> int
+  end
+
+  val mem_edge : t -> V.t -> V.t -> bool
+end
+
+module type STORE = sig
+  type t
+  type node
+  type lattice
+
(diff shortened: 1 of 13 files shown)


corneliuhoffman:
will come with interfile

PR #711, opened 2026-05-19
Fix taint not flowing through return value of `returns_this` methods

maciejpirog:
When a method is modelled with `returns_this = true` (e.g. `StringBuilder.append` in Java), the taint analysis was failing to propagate taint to the return value in patterns like:

```java
String s = source();
StringBuilder b = new StringBuilder();
String v = b.append(s).toString();  // v should be tainted, but wasn't
sink(v);
```
**Root cause:** The signature for such a method contains two effects: a `ToLval` that taints this with the argument, and a `ToReturn(BThis)` that returns this's taint. Both effects are instantiated simultaneously from the pre-call environment, before `ToLval` has updated this. So when `ToReturn(BThis)` is instantiated, b is still untainted, the resolved taint set is empty, and the effect is dropped.

**Fix:** Instead of modelling the return value purely as `BThis` (resolved at instantiation time), the `ToReturn` effect now includes both `BThis` (for any pre-existing taint on the receiver) and the argument's taint directly. Since both are resolved from the pre-call environment, the argument taint is correctly captured without depending on the ordering of effect application.

### DIFF
--- a/src/tainting/Builtin_models.ml
+++ b/src/tainting/Builtin_models.ml
@@ -271,7 +271,14 @@ let add_arg_taints_this_signatures db method_names arity ~taint_arg_index
   let to_lval = to_lval_this (arg_taint_set taint_arg_index) in
   let effects =
     if returns_this then
-      Effects.of_list [ to_lval; return_effect (this_taint_set ()) ]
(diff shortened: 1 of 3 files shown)


dimitris-m (approved):
lgtm

PR #716, opened 2026-06-01
Go: Allow tilde and union in function args

maciejpirog:
The fix adds a new `type_parameter_declaration` grammar rule (separate from `parameter_declaration`) that allows type parameter constraints to be either a regular type or a `constraint_elem` (tilde/union forms like `~uint8 or ~int | ~string`). Examples:

```go
func Foo[S ~uint8]() {}                        // single tilde
func Foo[S ~uint8 | ~string]() {}              // tilde union
func Foo[S, T ~int]() {}                       // shared tilde constraint
type Stack[T ~[]E, E any] struct{}             // tilde in type def**
```

In the OCaml parser, constraint elements are converted to a `TInterface` wrapping `Constraints`, matching the representation already used for interface body type constraints. Note that the tilde syntax `~uint8` in `[S ~uint8]` is a type constraint, not a type — and in Go's type system, inline constraints like ~uint8 are semantically equivalent to interface{~uint8}. Wrapping in `TInterface` with `Constraints` reuses the existing Go AST representation for "a set of type constraints", which is exactly what `~uint8` means regardless of whether it appears in an interface body or a function's type parameter list.

**TODO:**

- [x] Merge the Go PR https://github.com/opengrep/semgrep-go/pull/2 and update the submodule

### DIFF
--- a/languages/go/tree-sitter/Parse_go_tree_sitter.ml
+++ b/languages/go/tree-sitter/Parse_go_tree_sitter.ml
@@ -1076,16 +1076,44 @@ and type_spec (env : env) ((v1, v2, v3) : CST.type_spec) =
   let v3 = type_ env v3 in
   DTypeDef (v1, tparams_opt, v3)
 
+and type_parameter_declaration (env : env)
+    ((v1, v2, v3) : CST.type_parameter_declaration) : parameter_binding list =
+  let id1 = identifier env v1 in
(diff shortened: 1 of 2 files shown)


dimitris-m (approved):
lgtm

csbbandelgar:
TYSM!

PR #719, opened 2026-06-01
feat(trivy): suppress CVE-2021-28676 in test fixture

Leathal1:
Suppresses CVE-2021-28676 related to python-pillow in the  test fixture by adding it to . This CVE is identified in  and confirmed to be a test fixture due to explicit version pinning.


dimitris-m:
Sorry, once more you are fixing our ... tests? There is no vulnerability here!

And who is TITO? ...

Unfortunately you will have to be blocked now...

PR #722, opened 2026-06-04
Allow "opengrep-" prefix in lang submodules

maciejpirog:
The Crystal tree-sitter parser lives in `languages/crystal/tree-sitter/opengrep-crystal` (using the `opengrep-` prefix rather than `semgrep-`), but `list-tree-sitter-langs.sh` only scanned `semgrep-*` directories. As a
  result, Crystal was never included in the language list fed to `flags.sh`, so `-ltree_sitter_crystal_stubs` was never added to the macOS linker flags, causing a build error. This PR fixes this by scanning `opengrep-` dirs as well.


dimitris-m (approved):
lgtm, nice

PR #725, opened 2026-06-10
Ast-to-IL: Translate more types of function params

maciejpirog:
Used for kwargs in Python and receiver params in Go (see tests).

Notes (to be translated to issues unless decided otherwise during review)
- Kwargs don't have field sensitivity:

```python
def my_function(**kwargs):
  # ruleid: safe
  sink(kwargs["safe"])
  # ruleid: taint
  sink(kwargs["src"])

my_function(safe = 123, src = source())
```

- General args in patterns don't capture kwargs, e.g. the pattern

```python
def $FUNC($PARAM):
   ...
```

will not capture the program

```python
def foo(**param):
  ...
```

So if someone just wants a rule like "taint comes from a function param (whatever form)", we need to duplicate patterns

### DIFF
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -3279,28 +3279,25 @@ and cond env cond_e : stmts * exp =
       expr env e
 
 and for_var_or_expr_list env xs : stmts =
-  let list_of_lists =
-    List.map
-      (fun x ->
-        match x with
-        | G.ForInitExpr e ->
-            let ss, _eIGNORE = expr env e in
-            ss
-        | G.ForInitVar (ent, vardef) -> (
-            (* copy paste of VarDef case in stmt *)
-            match vardef with
-            | { G.vinit = Some e; vtype = opt_ty; vtok = _ } ->
-                let ss1, e' = expr env e in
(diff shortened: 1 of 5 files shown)


dimitris-m (approved):
lgtm

src/analyzing/AST_to_IL.ml
```diff
@@ -3332,13 +3329,19 @@ and parameters params : param list =
...
            let pname = var_of_id_info i pinfo in
            ParamPattern ({ pname; pdefault }, pat)
-       | G.ParamReceiver _param ->
-           (* TODO: Treat receiver as this parameter *)
-           ParamFixme (* TODO *)
+       | G.ParamReceiver { pname = Some i; pinfo; pdefault; _ } ->
+           (* Go receiver: treat as a regular named parameter for now.
+            * TODO: Ideally it would map to VarSpecial(This), but that requires
+            * rewriting all uses of the receiver variable in the body. *)
+           Param { pname = var_of_id_info i pinfo; pdefault }
```
corneliuhoffman, line 3336: I will probably have an issue with this in interfile. in interfile I introduced a special Il.ParamReceiver for exactly that reason

Issue #727, opened 2026-06-10
install.sh binary smoke test fails inside docker buildx multi-platform builds


bchandler-12:
**Describe the bug**
When building a Docker image that uses install.sh to install opengrep as part of a `docker buildx build --platform linux/amd64,linux/arm64` run, the build fails at the binary smoke test with:

```/tmp/install.sh: line 226: /root/.opengrep/cli/v1.22.0/opengrep: cannot execute: required file not found```

This happens during the `linux/amd64` platform build when QEMU is also registered for the concurrent linux/arm64 build. The exact mechanism is unclear, but the ENOENT ("required file not found") is consistent with the kernel routing a binary execution through a binfmt_misc handler whose interpreter path doesn't exist inside the container. The binary itself downloads and passes cosign verification, only the smoke test fails.

**To Reproduce**
1. On an x86_64 Linux host with Docker and QEMU installed (e.g. via docker/setup-qemu-action), create a Dockerfile:
```
FROM python:3.13-alpine
RUN apk update && apk add bash curl cosign
RUN curl -Lo /tmp/install.sh https://raw.githubusercontent.com/opengrep/opengrep/main/install.sh \
    && chmod +x /tmp/install.sh \
    && /tmp/install.sh
```

2. Build with:

`docker buildx build --platform linux/amd64,linux/arm64 .`

3. Observe the `linux/amd64` build fail at `Testing binary...` with `cannot execute: required file not found.`

**Expected behavior**
The install succeeds on both platforms, or the script provides a flag/env var (e.g. OPENGREP_SKIP_BINARY_TEST=1) to bypass the smoke test for environments where binary execution is known to be unreliable (cross-compilation, QEMU emulation).

**Screenshots**
N/A

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Opengrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
 GitHub Actions runner (ubuntu-latest, x86_64) using docker/setup-qemu-action + docker/setup-buildx-action for multi-platform builds. Reproduced on an Amazon Linux EC2 instance with the same docker buildx setup.

**Use case**
Building multi-architecture Docker images that ship opengrep for use in CI security scanning pipelines.

dimitris-m:
PR is merged, thanks

PR #728, opened 2026-06-10
install.sh: add OPENGREP_SKIP_BINARY_TEST env var to bypass smoke test

bchandler-12:
**Fixes:** #727

**What Changed:**
- Added `OPENGREP_SKIP_BINARY_TEST=1` environment variable to `install.sh` that bypasses the binary smoke test.

**Why:**
`docker buildx build --platform linux/amd64,linux/arm64` fails at the smoke test (`Testing binary...`) on the `linux/amd64` build when QEMU is registered for the concurrent `arm64` build. The error is `cannot execute: required file not found` — consistent with the kernel routing binary execution through a `binfmt_misc` handler whose interpreter path doesn't exist inside the container. The binary itself downloads correctly; only the smoke test fails.

Setting `OPENGREP_SKIP_BINARY_TEST=1` bypasses the test and allows multi-platform builds to proceed. Verified on an Amazon Linux EC2 instance with `docker buildx build --platform linux/amd64,linux/arm64`.

**Usage:**
```dockerfile
RUN OPENGREP_SKIP_BINARY_TEST=1 /tmp/install.sh
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -221,21 +221,27 @@ main() {
         fi
 
         echo "Testing binary..."
-        # Test by calling --version on the downloaded binary
-        STDERR_LOG=$(mktemp)
-        TEST=$("${INST}/opengrep" --version 2>"$STDERR_LOG" || true)
-        if [ -z "$TEST" ]; then
-            echo "Failed to execute installed binary: ${INST}/opengrep." 1>&2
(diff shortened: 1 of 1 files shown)


dimitris-m (approved):
lgtm, thanks!

PR #729, opened 2026-06-16
Fix Windows '..' walk-up in reported file paths

corneliuhoffman:
> So we're scanning with arguments:
> opengrep-cli.exe scan --sarif --experimental C:\Users\e\Source\Repos\A --config path-to-rules
>
> And in the results, the path for each file seems to change from C:\Users\e\Source\Repos\A to
> C:\Users\e\Source\Repos\A\..\..\source\repos\A.

On case-insensitive filesystems (Windows), when the scanning root the user typed differs from git's canonical path — in **case** (`Source\Repos` vs `source\repos`) or as an **8.3 short name** (`RUNNER~1` vs `runneradmin`) — native osemgrep reported file paths with a `..` walk-up (e.g. `..\..\source\repos\A\file.cs` instead of `A\file.cs`). The path still resolves on disk, but it breaks any SARIF/JSON consumer that string-compares or de-duplicates locations.

Native target discovery relativizes the typed scanning root against git's canonical paths with the purely syntactic `Fpath.relativize`; when the two spell a shared directory differently it can't see the target as a descendant and emits a "walk up, then re-descend" path full of `..`. (The Python CLI is unaffected — it runs `git ls-files` from the resolved root and prefixes the typed path, so it never relativizes divergent spellings.)

### Fix
- Add `Rpath.canonical_if_win`: on Windows, resolve a path to its physical real path (normalises case **and** 8.3 short names); a no-op on case-sensitive filesystems.
- In `Find_targets` and `Git_wrapper`, canonicalise `cwd` and the scanning root for the relativize comparison only, and keep the user's typed root as the output prefix — so reported paths stay exactly as typed (matching the Python CLI) and contain no `..`.

### Tests
- `.github/workflows/windows-path-regression.yml` (manual, `workflow_dispatch`): builds on a Windows runner, then scans a `MixedRepo` via its lower-cased root with `opengrep-cli scan --experimental --json` and asserts the result paths have no `..`. The bug only reproduces on Windows, so there's no cross-platform unit test.

### DIFF
--- a/.github/workflows/windows-path-regression.yml
+++ b/.github/workflows/windows-path-regression.yml
@@ -1,0 +1,70 @@
+# Windows-only end-to-end regression for the path '..' walk-up in osemgrep
+# target discovery (Find_targets / Git_wrapper). On case-insensitive and 8.3
+# short-name filesystems the typed scanning root and cwd can diverge from
+# git's canonical paths; we canonicalise them (Rpath.canonical_if_win) before
+# relativizing, so reported paths keep the typed prefix and have no '..'.
+#
+# The bug only reproduces on Windows, so this workflow is the regression guard
+# (there is no cheap cross-platform unit test). It runs only on demand.
+name: windows-path-regression
+
+on:
+  workflow_dispatch:
+
+jobs:
+  # Reuse the existing Windows build (it already declares workflow_call) so we
+  # don't duplicate the opam/cygwin/dune setup. It uploads
+  # 'opengrep-core-and-dependent-libs-w64-artifact' (opengrep-core.exe + DLLs).
+  build:
+    uses: ./.github/workflows/build-test-windows-x86.yml
+
+  regression:
+    needs: build
+    runs-on: windows-2022
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/download-artifact@b14cf4c92620c250e1c074ab0a5800e37df86765 # ratchet:actions/download-artifact@v4
+        with:
+          name: opengrep-core-and-dependent-libs-w64-artifact
+      - name: unpack binary + dlls
+        run: tar xvfz artifacts.tgz
+      - name: case-insensitive path regression (native osemgrep)
+        run: |
+          set -eu
+          # 'scan' is dispatched by argv[0]: copy opengrep-core.exe to
+          # opengrep-cli.exe (same bytes, DLLs alongside) so it runs the osemgrep
+          # CLI, and pass --experimental to force native osemgrep (Find_targets)
+          # instead of delegating to pyopengrep -- otherwise the buggy discovery
+          # code never runs.
+          cp artifacts/opengrep-core.exe artifacts/opengrep-cli.exe
+          CLI="$PWD/artifacts/opengrep-cli.exe"
+          base="$(mktemp -d)"
+          mkdir -p "$base/MixedRepo/sub"
+          ( cd "$base/MixedRepo" \
+              && git init -q \
+              && git config user.email a@b.c \
+              && git config user.name ci \
+              && printf 'print("foo")\n' > sub/test.py \
+              && git add -A )
+          printf 'rules:\n  - id: t\n    pattern: print(...)\n    message: t\n    severity: ERROR\n    languages: [python]\n' > "$base/rules.yml"
+          winbase="$(cygpath -w "$base")"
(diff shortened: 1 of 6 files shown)


.github/workflows/build-test-windows-x86.yml
```diff
@@ -361,6 +364,48 @@ jobs:
...
+          paths="$(grep -o '"path":[[:space:]]*"[^"]*"' out.json || true)"
+          printf '%s\n' "$paths"
+          # Any '..' segment in a result path is the walk-up (the '...' ellipsis
+          # lives in the rule, never in a path value).
+          if printf '%s\n' "$paths" | grep -qE '\.\.[\\/]'; then
+            echo "REGRESSION: '..' walk-up in result path:"
+            printf '%s\n' "$paths" | grep -E '\.\.[\\/]'
+            exit 1
+          fi
+          echo "OK: no walk-up"
```
dimitris-m: But is the intention to maintain this code in this workflow?

.github/workflows/build-test-windows-x86.yml
```diff
@@ -447,3 +492,30 @@ jobs:
...
+          printf 'rules:\n  - id: t\n    pattern: print(...)\n    message: t\n    severity: ERROR\n    languages: [python]\n' > "$base/rules.yml"
+          winbase="$(cygpath -w "$base")"
+          winlower="${winbase}\\mixedrepo"
+          winrules="$(cygpath -w "$base/rules.yml")"
+          echo "=== raw: Path.resolve() + git ls-files with verbatim lower-cased cwd ==="
+          python -c "import subprocess,pathlib; p=pathlib.Path(r'${winlower}'); rp=p.resolve(); print('resolve():', rp); r=subprocess.run(['git','ls-files'],cwd=str(rp),capture_output=True,text=True); print('ls-files stdout:', repr(r.stdout)); print('ls-files stderr:', repr(r.stderr))"
+          echo "=== wheel scan from lower-cased root --json ==="
+          PYTHONIOENCODING=utf-8 opengrep scan --json --output "$base/out.json" -f "$winrules" "$winlower" 2>&1 | tail -3 || true
+          echo "--- result paths ---"
+          grep -o '"path":[[:space:]]*"[^"]*"' "$base/out.json" 2>/dev/null || echo "(no out.json / no paths)"
```
dimitris-m: As above, I understand in this case we had to check in CI, but a separate workflow would be better, and in any case our main windows workflow should not carry these things right?

libs/commons/Fpath_.ml
```diff
@@ -49,6 +49,97 @@ end
...
+   see that [p] is a descendant of [root]: it finds a shorter common prefix and
+   emits a "walk up then descend again" path with '..' segments that re-descends
+   using the other casing (e.g. root '/c/Source/Repos/' and target
+   '/c/source/repos/file' relativize to '../../source/repos/file' rather than
+   'file'). Folding case before comparing keeps a case-insensitive descendant
+   relativizing to a clean relative path.
+
+   Returns None when [root] and [p] differ in absoluteness or volume, matching
+   Fpath.relativize. *)
+let relativize ?(caseless = case_insensitive_fs) ~root p =
```
dimitris-m: Why not have the new codepath _only_ when case_insensitive_fs ?

This reduces a bit the blast radius just in case.

dimitris-m:
This situation re-enforces the need to ensure that we have tests passing on windows. We should probably prioritise this.

.github/workflows/build-test-windows-x86.yml
corneliuhoffman: well it is rather fast so why not?

.github/workflows/build-test-windows-x86.yml
corneliuhoffman: separated

libs/commons/Fpath_.ml
corneliuhoffman: good comment, changed

.github/workflows/build-test-windows-x86.yml
dimitris-m: my objection is that if we do this, we will eventually accumulate regression tests in the wrong place; our only choice is to make sure tests pass on Windows CI and then we can add a normal test...

left the new workflow but but it will only be triggered manually (because building win....)

corneliuhoffman (commented):
yeah this is a lot better

PR #734, opened 2026-06-23
C#: Translation fixes

maciejpirog:
# Make C# translation to `AST_generic` more complete (records, pattern matching) + teach the IL about pattern guards

## Why

The C# frontend (`languages/csharp/generic/Parse_csharp_tree_sitter.ml`) lowers the
tree-sitter CST into `AST_generic`. A number of constructs parsed fine but were
then thrown away during that lowering: they were translated to the
`OtherXxx`/`todo` escape hatches (`OtherStmt (OS_Todo, …)`, `OtherPat ("Todo", …)`,
`OtherType ("Todo", …)`), which keep a token for range purposes but **discard the
node's structure**. Anything downstream — matching, naming, dataflow, taint — then
sees an opaque blob, so no rule could match those constructs and no taint could
flow through them.

This PR replaces those drops with real `AST_generic` nodes, and then fixes a
related gap in the IR lowering (`AST_to_IL`) so the newly-produced pattern guards
actually participate in condition evaluation.

You don't need to know C# to review this; everything below is described in terms of
`AST_generic` / `IL` constructs.

### Records → `ClassDef`
Record declarations (a class-like type with a generated constructor/equality, very
common in modern C#) were dropped wholesale, so their entire body — fields, methods —
was lost. They are now lowered like any other class:

- `DefStmt (entity, ClassDef { … })` with `ckind = Class`.
- A `KeywordAttr RecordClass` is added to the entity's attributes so a record is
  distinguishable from a plain class.
- The positional/"primary constructor" parameters become `field`s (same treatment
  classes/structs already get for their primary constructors).
- The base list becomes `class_parent`s; the first parent may carry constructor
  arguments, modelled as `(type_, Some arguments)`.
- Both the `{ … }` body and the bodyless `;` form are handled.

### Pattern matching
The C# pattern grammar (used in `is`-expressions and `switch`) was almost entirely
lowered to `OtherPat ("Todo", …)`. Each pattern now maps to the closest
`AST_generic` pattern:

| source pattern | `AST_generic` |
|---|---|
| relational (`< 5`, `>= 0`, …) | `OtherPat (op_string, [E rhs])` — see "relational patterns" |
| `or` | `PatDisj` |
| `and` | `OtherPat ("AndPattern", [P a; P b])` |
| `not` | `OtherPat ("NotPattern", [P a])` |
| parenthesized | the inner pattern (parens dropped) |
| type pattern | `PatType` |
| property pattern `{ K: p, … }` | `PatRecord` with `dotted_ident` keys (incl. extended `A.B:` keys) |
| list slice `..` | `OtherPat ("SlicePattern", [])` |
| recursive pattern w/ binder | `PatAs`; positional+property combined preserved via `OtherPat` |

#### Relational patterns: where the comparison lives
A relational pattern only gives us the right-hand side (`< 5`); the value being
matched — the *subject* — lives in the enclosing `is`/`switch`, not at the pattern
node. It is tempting to thread that subject into the pattern and build
`PatWhen (PatWildcard, subject < 5)`, but that bakes the subject *expression* into
the AST: it then appears both as the scrutinee and inside the guard (the same
mutable `G.expr`), and in the IL the scrutinee is lowered twice — a side-effecting
`foo() is < 5` would evaluate `foo()` twice.

Instead we keep the subject **implicit**, exactly as Kotlin's `when` conditions do
(`in 1..10` is recorded with no left operand). The frontend emits only the operator
(as the `todo_kind` string) and the rhs: `OtherPat ("<", [E 5])`. There is no
subject in the AST at all — no sharing. The real comparison `scrutinee < 5` is then
reconstructed in the IL by `pattern_match_cond` (Part 2), against the single,
already-materialised scrutinee. Matching still works because it is tag-based
(`OtherPat ("<", …)` only matches `OtherPat ("<", …)`).

### Other dropped constructs
- `fixed (…) stmt` → `WithUsingResource` (a scoped resource acquisition, like `using`).
- Function-pointer types (`delegate*<T1, …, TRet>`) → `TyFun`, **preserving** the
  per-parameter `ref`/`in`/`out` modifiers as attributes (rules may match on them).
- `goto case e;` / `goto default;` → `OtherStmt` that *keeps* the target expression
  and tokens, instead of dropping them.
- Custom property-accessor names are preserved instead of becoming `"Todo"`.
- A bodyless event field no longer emits a stray todo statement.

All `todo_pat`/`todo_attr`/`todo_type` helpers are now unused and removed.

## Part 2 — `AST_generic` → `IL` (`AST_to_IL.ml`)

The inert pattern nodes from Part 1 only become useful to dataflow once the IR
turns them into a condition. `pattern_match_cond env scrutinee pat` builds the
boolean predicate "does `scrutinee` match `pat`?" — and it is *handed* the
scrutinee, so it is the natural place to synthesise the comparison once. It already
does this for other patterns (`PatLiteral 1` → `scrutinee == 1`, `PatList` →
`length(scrutinee) == n && …`, `PatType Foo` → `scrutinee is Foo`). We add to that
table:

- **Relational patterns.** `OtherPat (("<" | "<=" | ">" | ">="), [E rhs])` →
  `Operator (op, [scrutinee; rhs])`, i.e. `scrutinee < rhs`. The scrutinee is the
  one the switch already lowered (a single `Fetch`), so it is evaluated once and —
  when the scrutinee is a parameter — stays anchored on that parameter, which is
  what lets guarded taint signatures fold it via `Eval_il_partial`.
- **`or`/`and`/`not` combinators.** `PatDisj` → `wrap_or`,
  `OtherPat ("AndPattern", …)` → `wrap_and`, `OtherPat ("NotPattern", …)` →
  `wrap_not`, so combined relational range checks (e.g. `>= 0 and < 10`) stay
  decidable.

One subtlety: the switch-to-cond lowering (`switch_expr_and_cases_to_exp`) has a
shortcut that treats `Case (_, OtherPat (_, [E c]))` as an *equality*
(`c == switch_expr`). A relational `OtherPat ("<", [E rhs])` has that exact shape,
so a relational case is matched *before* that shortcut and routed through
`pattern_match_cond`; otherwise `case < 0:` would silently become `0 == scrutinee`.

`PatWhen` itself is left unchanged (its guard is still dropped in cond extraction);
nothing in this PR relies on it anymore.

## Tests
- `tests/patterns/csharp/`: matching tests for records and the relational/property/
  not/or patterns (verifying the translations round-trip: pattern and target lower
  identically and match).
- `tests/rules/cross_function_tainting/`:
  - `test_guards_relational_pattern_csharp.*`: `guarded_taint_signatures` over a
    `switch` whose sink sits in a relational / `and`-range / `or` case; effects are
    dropped when the caller's constant makes the case non-matching, kept otherwise.
    Exercises the Part 2 changes end-to-end.
  - `test_csharp_record_taint.*`: taint flowing through a method defined in a record
    body (impossible before, since records were dropped).

All new tests pass, and the change does not regress pattern/switch/taint tests in
other languages.


corneliuhoffman (approved):
I think this is good,

dimitris-m (approved):
lgtm

PR #735, opened 2026-06-24
Split Graph_from_AST into three concept-aligned modules

corneliuhoffman:
Behavior-preserving refactor. Graph_from_AST.ml (1258 lines) is split into:

- Callee_resolution: fn_id/func_info/node types, resolution primitives (find_func_in_scope, pick_by_arity, ...), resolve_constructor_from_type, and identify_callee.
- Callback_extraction: higher-order-function callback machinery (extract_hof_callbacks and friends).
- Graph_from_AST: orchestration only (build_call_graph, find_functions_containing_ranges, extract_calls, ...). Public .mli unchanged.

Dependency DAG is acyclic: Callee_resolution <- Callback_extraction <- Graph_from_AST. The moved code is identical; consumers bring the new modules into scope with `open`, so Graph_from_AST.ml changes by only the two added opens. No external files reference the moved (module-internal) functions.


src/tainting/Callee_resolution.ml
```diff
@@ -0,0 +1,342 @@
...
+                      | _ -> false
+                  ) all_funcs in
+                  (* Debug: show all function names *)
+                  let all_names =
+                      all_funcs
+                      |> List.map (fun f -> show_fn_id f.fn_id)
+                      |> String.concat ", "
+                  in
+                  Log.debug (fun m -> m "CALL_EXTRACT: In class %s, call to %s, checking %d funcs, method_exists=%b, ALL: [%s]"
+                      class_name_str callee_name_str (List.length all_funcs) (Option.is_some method_match) all_names);
```
dimitris-m: `all_names` maps `show_fn_id` over every entry in `all_funcs` and concatenates on every class-context call, whether or not debug logging is on, because the `let` sits outside the `Log.debug` thunk. Build it inside the thunk so it only runs when the log level is active:

```ocaml
Log.debug (fun m ->
    let all_names =
      all_funcs |> List.map (fun f -> show_fn_id f.fn_id) |> String.concat ", "
    in
    m "CALL_EXTRACT: In class %s, call to %s, checking %d funcs, method_exists=%b, ALL: [%s]"
      class_name_str callee_name_str (List.length all_funcs)
      (Option.is_some method_match) all_names);
```

dimitris-m: fixed

maciejpirog (approved):

PR #739, opened 2026-06-29
test(pyopengrep): default --test to no timeout

dimitris-m:
Also honour --timeout/--timeout-threshold there. Avoids a spurious
'rule id mismatch' when a slow rule hits the timeout under load and
drops all its findings on a test file.


corneliuhoffman (commented):
lgtm

corneliuhoffman (approved):

PR #741, opened 2026-06-29
Java: Fixes in Java -> AST

maciejpirog:
A few updates to how Java is translated to the AST. Most of the changes requires updating Java's AST and the Menhir parser. 

- Support for Java 16's syntax `o instanceof Type s`, which bounds `s` of type `Type` after the cast. The produced AST is based on the exact same construction in C#,
- Java 9 module descriptors (`requires`, `exports`, `opens`, etc.) are now fully translated instead of ignored,
- Annotation type elements (the entries declared inside a Java `@interface`, like `String value();` or `int count() default 0;`) are now read as proper methods (with their name and return type) instead of being thrown away as an unknown "todo" node. Their optional default value is also preserved, so rules can now find and match these annotation members,
- Method-reference type arguments: `Foo::<String> bar` is no longer dropping `<String>`,
- Qualified instance creation (`x.new Y(...)`): `NewQualifiedClass` flattened everything into `any`s, with no real new node. Factored a shared `new_class_or_anon` helper (also now used by `NewClass`, removing duplication) so the inner `new Y(args){body}` is a real `G.New/AnonClas`s subexpression nested inside the `OtherExpr`.


dimitris-m (approved):
LGTM

PR #742, opened 2026-06-30
JS/TS: Fixes in translation to AST

maciejpirog:
Fixes several JS/TS translation gaps where TypeScript constructs were dropped, given placeholder names, or parsed inconsistently between the two parsers (tree-sitter for targets, menhir for patterns) — making them unmatchable. Where needed, fixes were applied to both parsers so patterns and targets produce the same generic AST.

**Changes**
- **Exports:** `export type { … }`, `export = X`, and `export as namespace X` now produce AST instead of being dropped.
- **Namespaces/modules:** now translate to `ModuleDef` instead of a fake IIFE; `namespace $X { … }` binds correctly.
- **Call/index signatures:** replaced placeholder names `CTOR??TODO` / `IndexMethod??TODO?` with `()` / `[]` and real tokens (similarly to other languages with `operator()`-style constructs),
- **Interface/class members:** interface bodies were emptied (matched vacuously) and class method/index signatures were dropped — both now preserved.
- **Index-signature matching** (`[k: string]: V`) now works in interfaces and classes with literal and metavariable keys/values, enforcing key and value types.
- **Function types:** the pattern parser dropped the `TyFun` wrapper for `(): T` / `() => T`, so `$F(): string` and `type $T = () => string` couldn't match — now built correctly.

### DIFF
--- a/languages/javascript/ast/ast_js.ml
+++ b/languages/javascript/ast/ast_js.ml
@@ -422,6 +422,9 @@ and definition_kind =
   | FuncDef of function_definition
   | VarDef of variable_definition
   | ClassDef of class_definition
+  (* TS 'namespace Foo { ... }' / 'module Foo { ... }'; the stmt is the body
+   * block (None for an ambient declaration without a body) *)
+  | ModuleDef of stmt option
   | DefTodo of a_todo_category * any list
 
 and variable_definition = {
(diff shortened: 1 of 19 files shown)


dimitris-m (approved):
LGTM

PR #744, opened 2026-06-30
Dart: Translation fixes

maciejpirog:
- Cascade expressions (`..`), which previously crashed the parser entirely,
- Remaining parser-crashing constructs — the `??=` null-aware assignment operator, `native` function bodies, configurable imports, and `var`-plus-explicit-type declarations
- Generic constructor calls like `Bar<String, int>(y)` now keep their type arguments


dimitris-m (approved):
Do we not need to also handle lowering to IL for this new OtherExpr?

maciejpirog:
> Do we not need to also handle lowering to IL for this new OtherExpr?

Yes, but I wanted to do it in a separate PR

PR #754, opened 2026-07-07
Elixir: Make condition's side effects visible in other branches of `cond` expression

maciejpirog:
Closes issue #644

### DIFF
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -310,6 +310,16 @@ let is_map_pair_pattern (p : G.pattern) : bool =
       true
   | _ -> false
 
+(* An Elixir `cond` branch condition is an expression, not a pattern:
+ * [expr_to_pattern] tags it [OtherPat("ExprToPattern", …)] (possibly wrapped
+ * in a singleton [PatTuple] by [pat_of_args_and_when]). Real `case` patterns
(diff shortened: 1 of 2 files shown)


src/analyzing/AST_to_IL.ml
```diff
@@ -4039,9 +4049,26 @@ and cases_and_bodies_to_stmts env switch_expr_opt tok break_label translate_case
         | _ -> []
       in
 
+      (* Elixir `cond`: a branch condition is an expression (tagged
+       * [ExprToPattern]), evaluated in sequence — its side effects (e.g.
+       * `y = source()`) precede and are visible to later branches. Emit them
+       * before the branch dispatch rather than scoping them to the matched
+       * arm (issue #644). Only `cond` produces [ExprToPattern], so real
+       * `case`/`switch` bindings keep their per-arm scoping. *)
```
dimitris-m: Could any other language require such a fix? Clojure? OCaml?

Btw the comment is not completely accurate since we also restrict to `env.lang =*= Lang.Elixir`.

dimitris-m (approved):

src/analyzing/AST_to_IL.ml
maciejpirog: nope, I didn't find any language that translates the patterns in a way that would also fit this PR. I shortened the comment (which explicitly says it is for Elixir) too

PR #760, opened 2026-07-15
fix the matching on ellipsis metavariable in js/ts

### LINKED ISSUE #392: JS / TS no matches when ellipsis metavariable is used next to regular meta-variable
**Describe the bug**
When an ellipsis metavariable is used next to regular meta-variable the pattern does not match.

Found this to be the case for JS / TS, but it might be also be the case for other languages.

**To Reproduce**
Using the folllowing JS / TS code:
```js
var config = {
    key1: value1,
    key2: value2,
    key3: value3,
};
```

With the following rules:
```yaml
rules:
  # this rule does NOT match
  - id: ellipsis-metavariables
    languages: [js, ts]
    severity: ERROR
    message: match for `ellipsis-metavariables`
    pattern-either:
      # these patterens do NOT work  
      - pattern: |
          { $...HEAD, key2: value2, $...TAIL }
      - pattern: |
          { $...START, key3: value3 }
      - pattern: |
          { key1: value1, $...END}

  # these rules DO natch
  - id: ellipsis-no-metavariables-head-and-tail
    languages: [js, ts]
    severity: INFO
    message: match for `ellipsis-no-metavariables-head-and-tails`
    pattern: |
      { ..., key2: value2, ... }
  - id: ellipsis-no-metavariables-start
    languages: [js, ts]
    severity: INFO
    message: match for `ellipsis-no-metavariables-start`
    pattern: |
      { ..., key3: value3 }
  - id: ellipsis-no-metavariables-end
    languages: [js, ts]
    severity: INFO
    message: match for `ellipsis-no-metavariables-end`
    pattern: |
      {key1: value1, ...}
```

**Expected behavior**
Currently the `ellipsis-no-metavariables-`.. rules will match, but the `ellipsis-metavariables` rule will not match.

**Screenshots**
Image

- [ X ] P1: important to fix or quite annoying

Currently investigating if maybe I could switch from using ast-grep, for seaching code, linting and (simple) codemod, to opengrep.
Due to the bug in the ellipsis metavariables I will probably keep using ast-grep for codemods and linting that required autofixes, and probably only use opengrep (and its powerful symantic search) for search en maybe for linting that require no autofix.

**Environment**
Using Linux x64 (Ubuntu 22.04)

**Use case**
Being able to use ellipsis metavariables with autofixes.

corneliuhoffman:
Fixes #392.
Patterns like `{ $...HEAD, key2: value2, $...TAIL }` never matched: field matching (m_fields / m_list__m_field) only special-cased plain `...` fields, so a `$...X` field — parsed in JS/TS objects as ashorthand property whose name and value are both the metavariable —fell into the generic case, was matched against a single concrete field, and failed.

This in now fixed but we note a subtlety, the normal record patterns match the order of the fields do not matter, in the case of ellipsio metavars they do (see the diff between `tests/patterns/js/metavar_ellipsis_fields` and `tests/patterns/js/unordered_fields2`

Added both pattern and autofix tests

### DIFF
--- a/perf/opengrep-scripts/README.md
+++ b/perf/opengrep-scripts/README.md
@@ -45,6 +45,14 @@ Binaries are cached in `binaries/<label>/venv/bin/opengrep`.
 
 Unstaged changes are allowed when the reference is cached (no checkout required).
 
+Both sides usually report the same opengrep version during development (main is
+typically still at the last release), so each side is identified by its commit as
+well: `<tag>-<short-sha>` when the commit is tagged, otherwise the short SHA
+(plus `-dirty` for a HEAD with uncommitted changes). These identifiers are passed
+to `bench.sh` as `--reference-label`/`--binary-label` and show up in the hyperfine
+result tables, `metadata.json`, and the report header, e.g.
+`opengrep-1.15.0@v1.15.0-a1b2c3d` vs `opengrep-1.15.0@d77978d-dirty`.
+
 ### `bench.sh`
 
 Core benchmarking script using [hyperfine](https://github.com/sharkdp/hyperfine). Compares two binaries (reference vs comparison).
@@ -57,6 +65,8 @@ Options:
 - `--reference-binary PATH` - Reference opengrep binary (required; `~/.local/bin/opengrep` is a common location)
 - `--binary PATH` - Binary to compare (required)
 - `--binary-variant TYPE` - Type of binary: `opengrep` or `semgrep` (auto-detected if binary is `semgrep` or ends with `/semgrep`)
+- `--reference-label LABEL` - Extra identifier (tag/sha) for the reference binary, shown as `opengrep-<version>@<label>` in results and metadata
+- `--binary-label LABEL` - Extra identifier (tag/sha) for the comparison binary
 - `--repo NAME[:RUNS]` - Benchmark specific repo (can be repeated)
 - `--taint-rules-only` - Use only taint rules
 - `--no-taint-rules` - Use only non-taint (search) rules
--- a/perf/opengrep-scripts/bench-against-sha.sh
+++ b/perf/opengrep-scripts/bench-against-sha.sh
@@ -28,9 +28,11 @@ GREEN='\033[0;32m'
 YELLOW='\033[1;33m'
 NC='\033[0m' # No Color
 
-log_info() { echo -e "${GREEN}==>${NC} $1"; }
-log_warn() { echo -e "${YELLOW}Warning:${NC} $1"; }
-log_error() { echo -e "${RED}Error:${NC} $1" >&2; }
+# printf, not `echo -e`: when the script is run as `sh bench-against-sha.sh`,
+# echo is POSIX and prints the -e literally.
+log_info() { printf '%b==>%b %s\n' "$GREEN" "$NC" "$1"; }
+log_warn() { printf '%bWarning:%b %s\n' "$YELLOW" "$NC" "$1"; }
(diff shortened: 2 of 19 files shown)


dimitris-m (commented):
### Suggestion: one `$...NAME` per object/record pattern, matched exactly like `...`

The core point: with a single ellipsis metavariable we regain the order-independence that plain `...` already has. `$...NAME` should match like `...` — fields in any order, not necessarily contiguous — and additionally bind the leftover fields to `NAME`. Capture is then the *only* difference between `...` and `$...NAME`, not ordering.

Today they diverge. `...` is order-free because `m_fields` partitions it out and it only sets `less_is_ok` over the order-independent `m_list__m_field`. `$...X` instead routes to `m_list_with_dots_and_metavar_ellipsis`, the positional walker for arguments and parameters, so concrete keys become order-sensitive and contiguous:

- `{ key1: v1, ... }` matches `{ key0, key1, key2 }`; `{ key1: v1, $...END }` does not — `key1` is pinned to the front.
- `{ ..., keyA, keyB, ... }` matches both `{ keyB, keyA }` and `{ keyA, mid, keyB }`; the `$...HEAD/$...TAIL` equivalent matches neither — the concrete keys must be adjacent and in order.

Swapping `...` for a capturing `$...X` therefore silently tightens matching, which is the subtlety the PR description already calls out.

**Proposal.** In unordered contexts (objects/records, `TyRecordAnon`, class bodies) allow at most one `$...NAME`, matched order-independently: match the concrete keys with `m_list__m_field` as today, then bind the single metavar to the leftover fields in source order. Keep the positional walker for arguments and parameters, which are genuinely ordered. Clean line: ordered contexts may take several positional ellipses; unordered contexts take one, and it behaves like `...`.

**Why one is enough.** With a single `$...X` the binding is unambiguous — exactly the fields no concrete key matched — so nothing forces positional matching. The two-ellipsis form used here, `{ $...HEAD, key2: value2, $...TAIL }`, is a workaround for the absence of order-free matching: its goal — match `key2` anywhere and keep the other fields — is met by `{ key2: value2, $...REST }` once matching is order-independent. Splitting the remainder into separate before/after captures has no stable meaning for an unordered record, and two or more ellipses are what reintroduce the positional ambiguity.

**Enforcement.** `Check_pattern.ml` (`metavar_checker` / `check`) already walks the pattern AST, knows `Mvar.is_metavar_ellipsis`, and raises `CheckFailure`. Count metavar-ellipsis fields per `Record`; more than one → a parse-time error ("at most one `$...FIELDS` metavariable is allowed in an object pattern") rather than an order-sensitive match with no diagnostic.

**Fixtures.**
- `tests/patterns/js/metavar_ellipsis_fields.{sgrep,js}`: `{ $...HEAD, key2: value2, $...TAIL }` → `{ key2: value2, $...REST }`.
- `tests/patterns/js/metavar_ellipsis_fields_order.*`: under order-independence both objects match, so flip the expectation — or repurpose it as a metachecking fixture for the more-than-one-ellipsis error.
- `tests/autofix/js/metavar_ellipsis_fields.*`: rewrite the pattern the same way.

**One tradeoff to weigh.** For autofix, `{ $...HEAD, key2, newkey, $...TAIL }` reproduces the surrounding fields in place, giving a minimal diff. A single trailing `$...REST` re-emits the captured fields at the metavar's position, so an interior anchor moves to the front. Matching semantics get cleaner; autofix that must preserve field position is the one case where two positional ellipses still read better.

dimitris-m (approved):

PR #764, opened 2026-07-22
[WIP] Fix paren ranges

maciejpirog:
In most of the langs, the ranges were off even in languages marked as "fixed" in #272. Reason: for pattern `$_ + $_` and expression `x = (2 + 3)` the result was reported together with parens as `(2 + 3)`. We fix it here by adding a special case in Parse_XXX_tree_sitter.ml for expressions that, if parenthesised, may influence the range of the outer expression.

Languages fixed in this PR:

- lua
- solidity
- circom
- hack
- swift
- c++
- aptos
- sui
- java
- jsonnet
- js


dimitris-m:
Can't we rebase instead of merging main into the Pr branch?

PR #770, opened 2026-07-25
TS: Fix parsing of 'new' as a property name in type declarations

### LINKED ISSUE #768: TypeScript: parse error on `new` used as a property name in type position
## Summary

A property literally named `new` inside a **type** (type literal or interface body) makes opengrep fail to parse the file. The file is reported as *"partially analyzed"* and its rules silently stop applying — the scan still exits `0`, so coverage drops with no signal.

The same property name in **value** position parses fine, so it is specific to type members.

## Reproduction

`opengrep 1.25.0`, with any rule (used a trivial `pattern: $X` probe):

| Snippet | opengrep | `tsc` 6.0.3 (`--strict`) |
|---|---|---|
| `type C = { new: string };` | ❌ partially analyzed | ✅ accepts |
| `type C = { new?: string };` | ❌ partially analyzed | ✅ accepts |
| `interface I { new: string }` | ❌ partially analyzed | ✅ accepts |
| `const o = { new: "y" };` | ✅ ok | ✅ accepts |

```console
$ cat a.ts
type C = { new: string };

$ opengrep scan --config rule.yaml a.ts
Some files were skipped or only partially analyzed.
  Partially scanned: 1 files only partially analyzed due to parsing or internal Opengrep errors
```

Renaming the property (`new` → `updated`) with everything else identical scans cleanly, which isolates the trigger to the name itself.

## Likely cause

TypeScript type members allow **construct signatures** (`new (): T`), so `new` is contextually special in that position. The parser looks like it commits to the construct-signature production and does not back off when the token is followed by `:` (an ordinary property named `new`). `new` is only reserved in expression position; as a property name it is valid TS.

## Impact

Quietly reduces coverage on real code. We hit this on a NestJS + React codebase whose audit/diff payloads use the shape `{ old: T; new: T }` — a natural way to model a change record. Nine source files (services, interceptors, React components) were partially analyzed while the scan reported success, and the property cannot simply be renamed: it is a persisted payload shape consumed by the frontend.

The silence is the sharp edge — `--error` does not catch it, so a green run can hide unscanned files. Even without a parser fix, it would help a lot if partial parses could optionally be made to fail the run.

## Environment

- opengrep 1.25.0 (`dev-util/opengrep-bin`, Gentoo)
- TypeScript 6.0.3 for the cross-check

Possibly the same family as #461 (`TypeScript: Unexpected satisfies`) — another valid TS construct the parser rejects.

JoeJoeflyn:
Fixes #768.

The TypeScript parser misparsed `type C = { new: string }`: `new` was dropped and `string` was treated as the field name. The `property_signature` grammar rule only allowed `_property_name` for the name field, excluding the `new` keyword. Add `alias('new', $.property_identifier)` so `new` is recognized as a valid property identifier, while construct signatures (`{ new(): string }`) still work.

Also enables strict parsing tests for TS and adds a regression test.


CHANGELOG.md
dimitris-m, line 1: We don't do this as you can see from other PRs.

dimitris-m:
Submodules are not pointing to something we can check.

Have you forked them? If so we could consider the contribution assuming other aspects are ok. For example that there are no unrelated changes, correctness, etc.

JoeJoeflyn:
Updated `.gitmodules` to point back to the org repos. The submodule commits are on my forks for now — could you pull them into `opengrep/semgrep-typescript` and `opengrep/semgrep-tsx` before merging? Branch is `fix/new-as-property-name` on both forks.

PR #772, opened 2026-07-28
JS/TS: Taint `..rest` of object in destructuring

maciejpirog:
E.g. when `x` is tainted, we want `rest` to be tainted after

```js
const { a, b, ...rest } = x;
```

Field sensitivity is achieved by:
- `rest` gets all the taints from `x`
- the other fields (`a` and `b` in the example`) are cleared from `rest` by setting them to untainted constant nulls.


corneliuhoffman (approved):
This LGTM (it is another one off the list of ts/js issues)

PR #784, opened 2026-08-09
Use lazy-with-restart in `lazy_content` and `lazy_ast_and_errors`

maciejpirog:
A per-rule timeout could take down the scan of the *whole file* — surfacing as an unhandled, rule-less fatal instead of a normal `Timeout` on the single rule that ran out of time. This happened because a timeout is delivered as an asynchronous exception, and when it interrupted the forcing of a **shared per-file `lazy`** (the parsed AST or file contents), OCaml's `Lazy` **memoized that exception** and re-raised it for every subsequent rule that touched the same value.

The fix introduces a small `Lazy_with_restart` module (a lazy that memoizes only on *success* and re-runs its computation after an interrupted force) and uses it for the two `Xtarget` fields that are shared across all rules on a file.

## The bug

- **Symptom:** intermittently, a timeout surfaces as an unhandled, rule-less `Memprof_limits.Limit_reached: token set` fatal (recorded as a generic file error such as "Other syntax error") instead of the normal per-rule `Timeout`.
- **Root cause:** per-rule timeouts are implemented with `memprof-limits`, which raises an *asynchronous* exception at an allocation point when the time budget is exceeded. The parsed AST lives in `Xtarget.lazy_ast_and_errors`, a `lazy` built once per file and forced on demand by the first rule that needs it — inside that rule's timeout scope. If the timeout fires while that force is in progress, the exception propagates, and `Stdlib.Lazy` stores a *re-raising closure* in the thunk (`Lazy.mli`: "If it raised an exception, the same exception is raised again"). Every later rule that forces the same thunk then re-raises that stored interrupt. Because `Memprof_limits.Limit_reached` is a private exception, no downstream handler recognizes it, so it escapes as a rule-less fatal. In other words: **one rule's timeout kills the other rules on that file.**
- **When it triggers:** whenever the time to force a shared lazy exceeds the per-rule timeout — in practice a very small `-timeout`, or a large enough file that parsing alone runs past the timeout.

## The fix

- **New module `libs/commons/Lazy_with_restart`** — a lazy that only memoizes a *successful* result. If forcing raises, the thunk is left unforced, so the next `force` re-runs ("restarts") the computation. (`Stdlib.Lazy` offers no way to avoid memoizing the exception, and `Lazy.force_val` leaves the thunk in an `Undefined` state instead — neither retries.) API: `make`, `from_val`, `force`, `is_val`.
- **Use it for `Xtarget.lazy_content` and `Xtarget.lazy_ast_and_errors`.** Now if a rule's timeout interrupts the read/parse, the shared value is not poisoned: a later (often simpler/faster) rule simply re-runs the computation, and the rule that was interrupted is reported as a normal `Timeout`. As a bonus, unlike `Stdlib.Lazy`, `Lazy_with_restart` does not raise `Lazy.Undefined` on concurrent forcing — at worst it recomputes — which matches the existing behavior of `Common.memoized`.

The success path is unchanged from `Lazy` (compute once, then return the cached value), so this only changes behavior in the exceptional case.

## Why only these two fields, and not every `lazy` in the engine

The bug needs a lazy that is **both**:

1. **shared across rules** — so that a *later* rule can force it again and inherit the poisoned thunk; and
2. **forced inside a per-rule timeout** — so the async interrupt can strike mid-computation.

Only `Xtarget`'s per-file fields (`lazy_ast_and_errors`, `lazy_content`) satisfy both: they are built once per target and read by every rule during matching.

The other lazy values do not qualify, so converting them would be needless churn (and, for some, a behavior change):

- **Per-match `tokens` lazies** (`Match_patterns`, `Match_search_mode`, …) are created per match, not shared across rules; a poisoned one dies with the rule that created it and can't reach another rule.
- **The report / code-snippet path** reads file content through `Common.memoized` (a Kcas hashtable), which — unlike `lazy` — does **not** memoize exceptions, so it is already safe.
- **Module-level / global lazies** (e.g. compiled regexes) are computed once, typically before matching, and are not tied to a rule's timeout.
- **`Lockfile_xtarget`'s** `lazy_content` / `lazy_dependencies` are a separate SCA-only type; left as-is here and can be migrated if the same pattern is confirmed for lockfile scanning.

Keeping the surface minimal also keeps the change low-risk.

### DIFF
--- a/libs/commons/Lazy_with_restart.ml
+++ b/libs/commons/Lazy_with_restart.ml
@@ -1,0 +1,27 @@
+(* See Lazy_with_restart.mli for the rationale (a lazy that restarts its
+   computation instead of memoizing an exception). *)
+
+type 'a state =
+  | Not_forced of (unit -> 'a)
+  | Forced of 'a
+
+type 'a t = { mutable state : 'a state }
+
+let from_fun f = { state = Not_forced f }
+
+let from_val v = { state = Forced v }
+
+let force t =
+  match t.state with
+  | Forced v -> v
+  | Not_forced f ->
+      (* If [f ()] raises, we never reach the assignment below, so [t.state]
+         stays [Not_forced] and the next [force] re-runs [f]. *)
+      let v = f () in
+      t.state <- Forced v;
+      v
+
+let is_val t =
+  match t.state with
+  | Forced _ -> true
+  | Not_forced _ -> false
--- a/libs/commons/Lazy_with_restart.mli
+++ b/libs/commons/Lazy_with_restart.mli
@@ -1,0 +1,46 @@
+(* A lazy value that does NOT memoize exceptions.
+
+   [Stdlib.Lazy] stores the exception raised while forcing into the thunk, so
+   every subsequent [Lazy.force] re-raises that same exception (see the note in
+   the manual: "If it raised an exception, the same exception is raised again").
+   That is a problem when the forced computation can be interrupted by an
+   *asynchronous* exception (e.g. a per-rule timeout interrupt from
+   memprof-limits): a transient failure gets baked into the thunk permanently
+   and is re-raised by every later reader of the shared value.
+
+   [Lazy_with_restart] only memoizes on *success*. If forcing raises, the thunk
+   is left unforced, so the next [force] re-runs ("restarts") the computation.
+
+   Because forcing can restart, the suspended computation may run more than once
+   (whenever a previous force raised), so it should be free of non-idempotent
+   side effects.
+
+   NOT concurrency-safe. This module is meant to be forced by a single forcer at
+   a time (opengrep forces each [Xtarget] from a single task/domain). A
+   concurrent [force] will not crash and will not observe a torn value (under the
+   OCaml 5 memory model a reader sees either the unforced thunk or a fully
+   initialised result). But it also does nothing to prevent two forcers from
+   both running the computation and each installing its own result: for a large
+   value such as an AST that silently yields two live copies for the same target
+   (extra memory charged against [--max-memory], and physically distinct tokens
+   handed to different callers), with no error to signal it — precisely the
+   situation [Stdlib.Lazy] turns into a loud [Lazy.Undefined]. Do not force the
+   same value from multiple threads or domains. *)
+
+type 'a t
+
+val from_fun : (unit -> 'a) -> 'a t
+(** [from_fun f] is a suspension of [f]. [f] is not run until the value is forced.
+    Replaces the [lazy (expr)] keyword: write [from_fun (fun () -> expr)]. *)
+
+val from_val : 'a -> 'a t
+(** [from_val v] is an already-forced suspension of the value [v]. *)
+
+val force : 'a t -> 'a
+(** [force t] returns the value of the suspension, computing it on the first
+    call. If the computation raises, the exception is propagated and the
+    suspension is left unforced, so a later [force] retries it. *)
+
+val is_val : 'a t -> bool
+(** [is_val t] is [true] iff [t] has already been forced successfully (i.e. its
+    computation returned without raising). *)
--- a/libs/commons/tests/Commons_tests.ml
+++ b/libs/commons/tests/Commons_tests.ml
@@ -12,4 +12,5 @@ let tests =
       Unit_String_.tests;
       Unit_List_.tests;
       Unit_File.tests;
+      Unit_Lazy_with_restart.tests;
     ]
--- a/libs/commons/tests/Unit_Lazy_with_restart.ml
+++ b/libs/commons/tests/Unit_Lazy_with_restart.ml
@@ -1,0 +1,54 @@
+(*
+   Tests for our Lazy_with_restart module.
+*)
+
+let t = Testo.create
+
+exception Boom
+
+(* The whole point of the module: forcing raises the first time, then the next
+   force restarts the computation and succeeds; [is_val] stays false as long as
+   no force has succeeded. *)
+let test_restart_on_exception () =
+  let calls = ref 0 in
+  let lz =
+    Lazy_with_restart.from_fun (fun () ->
+        incr calls;
+        if !calls = 1 then raise Boom;
+        !calls)
+  in
+  Alcotest.(check bool) "not forced initially" false (Lazy_with_restart.is_val lz);
+  (try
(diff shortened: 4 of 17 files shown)


corneliuhoffman:
Traced the diagnosis through the code and it holds up. `Time_limit.set_timeout` creates a fresh `Memprof_limits.Token` per rule, and only the `limit_with_token` for *that* token turns `Limit_reached` into `Error _ -> None -> Timeout`. A `Limit_reached` memoized in a `lazy` and re-raised under a *later* rule's token has no matching handler frame left, so it escapes as exactly the rule-less fatal you describe. The "report path is already safe" claim also checks out — `Common.memoized` writes the Kcas table only on the success path.

Some things I'd like addressed, or at least written down:

### 1. The worst case goes from "fast and wrong" to "slow and right"

When the parse itself is what blows the budget (your second trigger: "a large enough file that parsing alone runs past the timeout"), every rule now re-parses from scratch and every rule times out. Before, only the first rule paid the wall-clock; the rest failed instantly off the poisoned thunk.

That's bounded when `timeout_threshold > 0` — `per_rule_boilerplate_fn` raises `File_timeout` after `threshold` timeouts, and the osemgrep default is 3. But `Core_scan_config.default.timeout_threshold = 0`, and `--timeout-threshold 0` is a documented setting ("If set to 0 will not have limit"). On that path the cost is `n_rules × timeout` for the pathological file: a 1000-rule ruleset at the default 5 s timeout turns one file into ~80 minutes where it used to be seconds.

Not an argument against the change — slow and correct beats fast and broken — but it deserves a line in the PR body, and ideally a sentence in `Lazy_with_restart.mli` saying restart is unbounded and capping retries is the caller's problem.

### 2. Was hoisting the parse out of the per-rule timeout considered?

The underlying oddity is that whichever rule happens to touch the AST first gets charged for parsing the whole file, and is therefore the one reported as `Timeout`. That misattribution exists independently of this bug. `Match_rules.check` already has the machinery (the `Profiling.ProfAll` branch pre-forces `lazy_ast_and_errors`), and `lazy_content` is in practice already forced outside any timeout by the prefilter in `is_relevant_rule_for_xtarget` / `group_rules`.

Forcing the AST unconditionally isn't free — you'd parse files where every rule would have been prefiltered away — so I'm not asking for it in this PR. But it's the complementary fix rather than a competing one, and it'd be useful for the description to say why the retry route was chosen over it.

### 3. The mli's concurrency claim promises more than the implementation delivers

> Unlike `Stdlib.Lazy.force`, concurrent forcing does not raise `Lazy.Undefined`; at worst the computation runs more than once.

The write is safe under the OCaml 5 memory model — a concurrent reader sees either `Not_forced` or a fully-initialised `Forced v`, never a torn value — so no crash, agreed. But "at worst runs more than once" undersells the consequence for `lazy_ast_and_errors` specifically: two domains racing produce **two distinct ASTs for the same target, both live**. That's a second full AST copy counting against `--max-memory`, and two sets of physically-distinct `Tok`s handed to different callers. `Stdlib.Lazy` raises `Undefined` precisely to make that situation loud.

The existing comments (in `Xtarget.ml`, and the NOTE in `is_relevant_rule_for_xtarget`) already assert one worker task per `Xtarget`, so this is theoretical today. But the mli is where the next person decides whether to reuse the module, and right now it reads as an endorsement. I'd either state plainly that it is not intended for concurrent forcing, or make `state` an `Atomic.t`. The `Common.memoized` comparison is fair for a `string`; much less so for an AST.

### 4. `Metavariable_pattern`'s `failwith` placeholder

```ocaml
Ok (Lazy_with_restart.make (fun () ->
      failwith "requesting generic AST for LRegex|LSpacegrep|LAliengrep"))
```

This is the one site where memoizing the exception was the correct behaviour — it's an unconditional "must never be forced" marker, and it's now a thunk that re-runs its own failure forever. Harmless in practice, and keeping the type uniform is worth more than the inconsistency, but it's the one conversion that isn't a strict improvement.

### 5. Smaller things

- **No tests.** `libs/commons/tests/` already exists with `Unit_*.ml` wired through `Commons_tests.ml`. Roughly fifteen lines — force raises once then the next force succeeds; `is_val` is false after a failed force; `from_val` implies `is_val` — would pin the entire contract, and this is exactly the kind of invariant that rots silently once the module has a few more users.
- **Naming.** `Stdlib.Lazy` spells the function constructor `from_fun`, not `make`. Matching it (`from_fun` / `from_val` / `force` / `is_val`) makes this a literal drop-in and the diff a pure substitution.
- **Formatting.** Several new lines run past the 80-column margin from `.ocamlformat` and look hand-edited — the `Some (String.length (Lazy_with_restart.force …))` in `Core_scan.ml`, the `let ast, _ = …` in `Match_search_mode.ml`, the `from_val (mast', [])` in `Metavariable_pattern.ml`. No CI job enforces this, so it needs a manual `dune fmt`.
- **Now-stale comment.** The `XXX: Can result in [CamlinternalLazy.Undefined] error` note sitting directly above the pre-force in `Match_rules.check` is false for the new type — `Lazy_with_restart` cannot raise it. Worth deleting here, since it's two lines above a line the diff already touches.

None of this blocks the core change: the success path is byte-for-byte the old behaviour and the exceptional path is strictly more correct. Mainly I'd like (1) recorded somewhere, and (3) either softened or hardened.

maciejpirog:
@corneliuhoffman
Re 1 and 2: Indeed, the burden of parsing is on the first task (it was like it before and has not changed in this PR). This could result in situations where the first task parses a file, but then times out on matching or tainting, and then the second task gets the file parsed and succeeds even though it might need even more time for matching or tainting. We could change this and make parsing a separate task for a file with its own timeout, but it would be a big architectural change, definitely outside of the scope of this bugfix PR.

True, that a file timeouting on parsing will have to timeout three times. My initial instinct for fixing this bug was to simply report timeout on other rules as well when the "token set" exception was reported from previous rules on the same file, but there is a chance that we would miss some results. I propose that we pursuit the architectural change in near future, but stick to the implementation in this PR for now, so that the bug is fixed.

Re 3: Yes, it would use a more explicit comment, will fix.

Re 4: I don't understand what the problem is. In this case, forcing should always throw an exception, it doesn't matter if it is the literally the same exception or not, because in this case (in contrary to the timeout exceptions) these are not local or generative.

Small notes: will fix.

corneliuhoffman (approved):
LGTM now

dimitris-m:
We can probably go back to this after 2.0 is published.

PR #786, opened 2026-08-11
Dockerfile: Fix parsing of square brackets

### LINKED ISSUE #778: Errors parsing valid Dockerfiles that contain `[` character
## Describe the bug
The opengrep Dockerfile parser fails and emits `[WARN] Syntax error` on some Dockerfile `RUN` commands that contain a `[` character, even though the file is valid per the Dockerfile spec and builds fine with `docker build`. 

## To Reproduce
Run opengrep on either Dockerfile below (`opengrep scan --verbose`):

```dockerfile
FROM alpine:3.19
RUN [ -f /etc/os-release ]
```
Error: `` `-f /etc/os-release` was unexpected ``

```dockerfile
FROM alpine:3.19
RUN grep 'Python [0-9]*\.[0-9]*'
```
Error: `` `[0` was unexpected ``

Both build successfully with `docker build`.

## Expected behavior
Shell-form commands containing `[` that aren't valid JSON should parse as shell-form, matching `docker build`'s behavior, instead of producing a parse error.

## Screenshots
N/A

## What is the priority of the bug to you?

- [x] P1: important to fix or quite annoying

## Environment
Official CLI binary, v1.26.0

## Use case
Fixing this would close a coverage gap in Dockerfile scans. `[ ... ]` test expressions are common in Dockerfile `RUN` commands (e.g. conditional file checks), but Dockerfiles using them are only partially scanned today. 

---
Likely related:
semgrep/semgrep#10306

maciejpirog:
**WARNING** Merge https://github.com/opengrep/semgrep-dockerfile/pull/5 first and then update the dockerfile submodule

Fixes #778.

## Summary

`RUN [ -f /etc/os-release ]` and `RUN grep 'Python [0-9]*\.[0-9]*'` emitted `[WARN] Syntax error`
even though both build fine with `docker build`, leaving the rest of the file unscanned.

`shell_fragment`'s main token excluded `[` from its *first* character:

```js
/[^\\\[\n#\s,=-][^\\\n<]*/
```

That exclusion existed only so the `"["` of `json_string_array` would win the longest-match race at
the start of a `RUN` body, but it applied at every token boundary. Hence the second case: `\.` ends a
token (the tail class stops at a backslash) and the next one would have to begin with `[`.

`docker build` distinguishes the exec form from the shell form by attempting to JSON-parse the
argument, which needs unbounded lookahead — not expressible in a regex token. So `[` becomes an
external token.

## Changes

**`lib/scanner.c`** — new `JSON_ARRAY_OPEN`. On `[` it looks ahead over the logical line validating a
real JSON string array (plus the overlay's `...` and `$METAVARIABLE` elements, and extras:
whitespace, line continuations, comments), consumes only the `[` via `mark_end`, and emits the token
only on success; otherwise the internal lexer takes over and `[` starts a shell fragment. This works
because the external scanner runs before the internal lexer and its token wins when it matches.
Dispatched after the `ERROR_SENTINEL` block (error recovery keeps today's behaviour) and before the
heredoc branch, returning immediately once it sees `[` so the lookahead cannot move the lexer out
from under `scan_marker`. Lookahead is capped at 64K — only a bound on the pathological case, since
validation gives up at the first character that cannot occur in an array; the limit is generous
because an `ENTRYPOINT` holding a classpath runs to several kilobytes. Stateless, so
serialize/deserialize are untouched.

**Grammars** — `[` dropped from the `shell_fragment` exclusion; `json_string_array` opens with
`alias($.json_array_open, "[")` in the base grammar and in the semgrep overlay (which redefines the
rule, so it is the one in effect). Regenerated `parser.c` / `CST.ml` / `Parse.ml` / `Boilerplate.ml`;
`EXTERNAL_TOKEN_COUNT` 5 → 6.

**`Parse_dockerfile_tree_sitter.ml`** — one constructor rename (`Pat_f8ab07f` → `Pat_7f1e397`), since
the shell-fragment regex changed. No other consumer changes: the `alias` keeps the array's opening
token typed as `Token.t (* "[" *)`.

## Behaviour

Matches docker on the ambiguous cases. `["a" "b"]` (missing comma) and `["a",]` (trailing comma) are
invalid JSON and now fall through to shell form; `[ "a" ]` is valid JSON and stays an array.

One divergence remains: a valid array followed by trailing text (`RUN ["a"] && echo done`) is shell
for docker, since the argument *as a whole* is not valid JSON, but we stop validating at the `]` and
commit to the exec form. Pinned as an expected-fail in `tests/parsing_todo/dockerfile/`.

## Tests

`./test -s dockerfile`: 170 tests, 168 pass + 2 xfail, 0 failures (was 153).

- 13 pattern pairs `tests/patterns/dockerfile/bracket-*`. These use
  `parse_and_resolve_name_fail_if_partial`, so each asserts strict parsing *and* exact match lines.
  They cover: invalid-JSON fallback (missing/trailing/leading comma, bad escape, `[42]`, `['a']`,
  `[["a"]]`); every JSON escape shape including `\u`, truncated `\u00`, non-hex `\uZZZZ`; `[]`,
  `[  ]`, brackets/commas/quotes inside strings; ellipsis and metavariable elements; comments as
  extras between array tokens; `[` around line continuations; bracket-heavy heredoc bodies and `<<-`;
  the backtick escape directive; `--mount=`/`--network=` params before the `[`; `[` in
  LABEL/ENV/ARG/WORKDIR/COPY values; a 4.8 KB classpath-sized array.
- 5 parsing fixtures: the issue's cases, arrays, `[` at EOF with no trailing newline (both forms),
  `RUN [`, `RUN [[ -f x ]]`, `RUN [ "a` unterminated, and a `SHELL ["powershell", …]` file so the
  non-bash `Other_shell_command` path is exercised.
- End to end on the issue's Dockerfiles: `--verbose` reports no partial parse, and `RUN [ -f $FILE ]`
  / `RUN grep $PAT` now match inside the previously broken commands.


dimitris-m:
Is there something supposed to exist in "tests/parsing_todo/dockerfile/"?
Because "Pinned as an expected-fail in tests/parsing_todo/dockerfile/" but nothing is there.

dimitris-m (approved):
lgtm modulo some offline comments

PR #809, opened 2026-08-28
taint: fix propagators when the 'to' precedes the 'from'

corneliuhoffman:
Within one instruction, propagator endpoints are processed in visit order. When the 'to' l-value is visited first (e.g. strcpy($DST, $SRC), where $DST is the first argument), it is queued as a pending propagation destination, but the flush of that queue lived only behind hook_propagate_to, a pro-engine hook that is never installed, so the taints were recorded and never delivered.

Flush the pending destination directly when the hook is absent: the taints are added to the destination l-value by side-effect, so subsequent uses of it are tainted.

### DIFF
--- a/src/tainting/Taint_lval_env.ml
+++ b/src/tainting/Taint_lval_env.ml
@@ -79,25 +79,6 @@ type t = {
 
 type env = t
 
-type prop_fn =
-  taints_to_propagate:taints_to_propagate ->
-  pending_propagation_dests:pending_propagation_dests ->
-  env
-
-type add_fn = IL.lval -> T.taints -> env -> env
-
-let hook_propagate_to :
-    (Var_env.var ->
-    T.taints ->
-    taints_to_propagate:taints_to_propagate ->
-    pending_propagation_dests:pending_propagation_dests ->
-    prop:prop_fn ->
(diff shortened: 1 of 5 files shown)


src/tainting/Taint_lval_env.ml
```diff
@@ -276,7 +276,24 @@ let propagate_to lang prop_var taints env =
...
+                env with
+                taints_to_propagate =
+                  VarMap.remove prop_var env.taints_to_propagate;
+                pending_propagation_dests =
+                  VarMap.remove prop_var env.pending_propagation_dests;
+              }
+            in
+            add_lval lang lval taints env
+        | None -> env)
     | Some hook ->
```
dimitris-m: We should also get rid of the hook perhaps, it's dead code

dimitris-m (approved):

dimitris-m (commented):
Let me do a quick 2nd review

src/tainting/Taint_lval_env.ml
corneliuhoffman: ammended

dimitris-m:
Let's merge it, LGTM!

PR #816, opened 2026-08-31
PHP: `from` is not a keyword

maciejpirog:
Zend test corpus coverage: 95.73% → 95.95%

## Before

The lexer treated `from` as a keyword.

## However...

In reality, `from` has a special meaning only if it directly follows the `yield` keyword. In all other contexts, `from` is a perfectly normal identifier.

## After the fix

The lexer treats `yield` + whitespace + `from` as a separate token, while otherwise `from` is not special. Incidentally, this is exactly how the PHP compiler works.


dimitris-m (approved):
LGTM, just needs a rebase

PR #822, opened 2026-09-02
PHP: Fixes in `catch`

maciejpirog:
Zend test corpus coverage: 96.01% -> 96.28%

Three defects in the `catch` grammar, and the matching they enable.

- Only the first `catch` of a `try` accepted several exception types, so `catch (A | B $e) {} catch (C | D $e) {}` failed on the second clause.
- A `catch` with no variable, as in `catch (Exception) {}` (PHP 8.0), was a syntax error.
- The types after the first were dropped on the floor, so `catch (A | B $e)` was recorded as catching only `A`. They are now kept as a union, the same node a union type annotation uses, and the tree-sitter parser agrees.

A multi-catch catches each of its types, so a rule naming any one of them now finds the clause:

```
catch (First $E) { ... }     // finds  catch (Second | First $e)
catch ($T $E) { ... }        // $T = Second | First
```

That last part is one case in `m_catch_exn` in `Generic_vs_generic`, the only change outside `languages/php/`. It is scoped to catch clauses: a union *type annotation* in TypeScript or Rust still does not match one of its members. Without it, preserving the union would silently stop `catch (\Exception $e)` rules from firing on `catch (\Exception | \Throwable $e)` — which they did before, since only the first type survived.

### DIFF
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -2971,12 +2971,61 @@ and m_catch a b =
       m_tok at bt >>= fun () ->
       m_catch_exn a1 b1 >>= fun () -> m_stmt a2 b2
 
+(* the types of a union, left to right; a lone type is a union of one *)
+and union_members ty =
+  match ty.G.t with
+  | G.TyOr (t1, _, t2) -> union_members t1 @ union_members t2
+  | _ -> [ ty ]
+
+and is_union_type ty =
+  match ty.G.t with
+  | G.TyOr _ -> true
+  | _ -> false
+
+(* a lone '$T' stands for the whole union, so it needs no per-type pass *)
+and is_metavar_type ty =
+  match ty.G.t with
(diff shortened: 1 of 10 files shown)


dimitris-m (approved):

src/matching/Generic_vs_generic.ml
```diff
@@ -2971,12 +2971,61 @@ and m_catch a b =
...
   (* dots: *)
   | G.CatchPattern (G.PatEllipsis _), _ -> return ()
   (* boilerplate *)
   | G.CatchPattern a, CatchPattern b -> m_pattern a b
+  (* a multi-catch such as PHP's 'catch (Exn1 | Exn2 $e)' catches every one of
+   * its types, so it matches a pattern naming any of them: one type, or
+   * several in any order. Trying the union whole first lets a metavariable
+   * still bind to it in full.
+   *)
+  | G.CatchParam a, B.CatchParam ({ B.ptype = Some bty; _ } as b)
```
dimitris-m, line 3002: should this not be above the previous case?

PR #824, opened 2026-09-02
PHP: More fixes - heredocs, intersection types in params, properties in const exprs

maciejpirog:
Zend test corpus coverage: 96.49% → 97.44%

## Heredoc in constant expressions

A heredoc or a nowdoc was only ever an expression, never a constant expression, so it could not initialise anything.

```php
const T = <<<EOT
plain
EOT;

class K
{
    const E = <<<EOT
in a class
EOT;

    public $prop = <<<'EOT'
a property default
EOT;

    public function m($x = <<<EOT
a parameter default
EOT
    ) {}

    public function n()
    {
        static $s = <<<EOT
a static variable
EOT;
        return $s;
    }
}

enum Backed: string
{
    case Bar = <<<BAR
an enum case
BAR;
}
```

One production covers all six positions. A heredoc that interpolates is accepted here and rejected later, as php-src does.

Not fixed: an indented closing marker, as in PHP 7.3's flexible heredoc. The indentation has to come off every line of the body, which the body has already been read as by the time the marker is seen.

## Intersection types in parameters

`f(X&Y $p)` and `f(X &$p)` share a prefix, so the parser had to choose at the `&`, before the token that tells them apart. Only parameters were affected; the same type already worked on a property and as a return type.

```php
function a(X&Y $p) {}
function b(X&Y&Z $p) {}
function c(\A\B & \C\D $p) {}

class K
{
    public function __construct(public X&Y $promoted) {}
    public function m(Countable&ArrayAccess $c): X&Y {}
}
```

php-src decides this in the lexer, by whether a `$` or a `...` follows the `&`. ocamllex cannot look ahead, so the split is done just after lexing instead, in the pass that already retokenises `>>` for generics. An intersection type then takes only the one spelling, a by-reference parameter only the other, and every other use of `&` takes either — which matters, because the target of a reference need not start with `$`: `foreach ($a as &self::$s)` begins with a name. Those forms appear nowhere in the php-src corpus, so the fixture carries them.

Not matchable: a metavariable as the second type, as in `f(A&$T $p)`, since `&$T` reads as a by-reference parameter. `f($T $p)` is unaffected.

## Properties in constant expressions

A constant expression could read a property off a class constant, which is how an enum case's value is written, but only in that one shape. PHP 8.3 allows it on any constant expression.

```php
const P = (new A)->prop;
const R = (null)?->test;
const S = __FILE__->foo;
const U = (new A)->{new B};
const W = Suit::Hearts->a->b;
```

Three special cases become two general productions, so the grammar is smaller than it was. PHP parses these and rejects the operands that turn out not to be objects when the constant is evaluated. `Suit::Hearts->value` keeps the tree it had.

## Flexible heredoc

Since PHP 7.3 the closing marker of a heredoc may be indented, and may be followed by anything rather than only by `;` and a newline. That is how a heredoc inside a method is conventionally written, so failing it took the whole file with it.

```php
class Mailer
{
    public function body(string $name): string
    {
        return <<<HTML
            <p>Hello {$name}</p>
            HTML;
    }
}

$x = trim(<<<TXT
    body
    TXT);
```

The marker's rule now takes the rest of the line, which makes it the longest match for any line starting with a name, and gives back whatever it should not have taken: the code after the marker, or the part of the body the ordinary rule would have left.

Two things about the body are decided by that marker, which the lexer reaches only after reading the body, so they are settled afterwards: the marker's indentation comes off every line, and the newline before it is not part of the string. The body's own newlines are content and are kept, which they were not before — a three-line body used to come out as `line oneline twoline three`.


languages/php/menhir/parsing_hacks_php.ml
```diff
@@ -188,6 +188,156 @@ let find_typehint toks =
...
+(* ampersands *)
+(*****************************************************************************)
+
+(* php-src decides in the lexer whether a '&' introduces a by-reference
+ * variable or joins two types, by looking at whether a '$' or a '...' comes
+ * next; intersection types accept only the latter spelling, by-reference
+ * parameters only the former, and bitwise and either. Without the split,
+ * 'f(X&Y $p)' and 'f(X &$p)' cannot be told apart until after the '&'.
+ * ocamllex cannot look ahead, so the split is done here instead.
+ * coupling: the "&" rules in Zend's zend_language_scanner.l
```
dimitris-m, line 201: zend_language_scanner.l is intentional?

dimitris-m (approved):
lgtm

Issue #138, opened 2025-03-11
Cleanup and remove unused code


dimitris-m:
See for example `libs/commons2/common2.mli`.

Many functions are not used anywhere. We should remove them to avoid fixing concurrency issues that don't really exist.

Issue #132, opened 2025-03-07
Refactor use of `set_global_memory_limit_mb` to avoid innocent data race


dimitris-m:
In function `set_global_memory_limit_mb` of file `libs/process_limits/Memory_limit.ml` we set a global reference with the global memory limit on every invocation.

This is a data race, because we did not change this decision of usign a `ref` in our fork of memprof-limits, but also we can refactor how this is used in opengrep to avoid this assignment on every invocation. 

This may require a slight re-think of how we deal with memory limits in opengrep.

Relevance once #123 is merged.

PR #827, opened 2026-09-05
Remove python

### LINKED ISSUE #584: Remove Dockerfile
Not used, misleading, not needed currently for most use cases as we have binaries one can install with a 1-liner in any docker image.

### LINKED ISSUE #627: --time --json: metavariable-pattern regex time misattributed to parse_time
## Summary

When a rule uses `metavariable-pattern` with `pattern-regex`, the regex evaluation time is reported under `parse_time` rather than `match_times` in `--time --json` output.

## Reproduction

Any rule combining `pattern-inside`, `pattern: $X`, and `metavariable-pattern` with a non-trivial `pattern-regex` on a large target file will show inflated `parse_time` and near-zero `match_times`, even though `--profile` correctly attributes the time to `Xpattern_match_regexp.matches_of_regexs`.

## Observed

```json
"match_times": [["rule-id", 0.017]],
"parse_time": 4.966,
"run_time": 4.983
```

A trivial `pattern-regex` on the same file shows `parse_time: 0.038`, confirming the file parse itself is fast.

## Impact

Misleading profiling data when diagnosing slow rules -- users are led to investigate parsing rather than regex performance.

### LINKED ISSUE #797: --exclude patterns containing a slash are not anchored, unlike the gitignore syntax --help documents
**Describe the bug**

`--exclude` is documented as gitignore syntax:

> `PATTERN` is a glob-style pattern that uses the same syntax as gitignore and semgrepignore, which is documented at https://git-scm.com/docs/gitignore#_pattern_format

But slash-containing patterns are not anchored the way gitignore anchors them. In gitignore, "a pattern containing a slash is relative to the directory level of the particular .gitignore file itself" — so `/myvendor/**` and `deep/myvendor/**` match at the project root only. opengrep treats every `--exclude` as if it were prefixed with `**/`, so both patterns skip matching directories at *every* depth.

The scan therefore reads fewer files than the pattern asks for, and the loss is silent: a file that is never scanned produces no finding, so there is nothing in the output to compare against.

**To Reproduce**

```bash
mkdir -p repro/{myvendor,deep/myvendor,a/deep/myvendor,src} && cd repro
for f in myvendor/lib.py deep/myvendor/lib.py a/deep/myvendor/lib.py src/main.py; do
  printf 'import os\n\n\ndef run(user_input):\n    os.system("ls " + user_input)\n' > "$f"
done
cat > rule.yaml <<'YAML'
rules:
  - id: os-system-call
    patterns:
      - pattern: os.system(...)
    message: os.system call
    languages: [python]
    severity: WARNING
YAML
git init -q . && git add -A && git -c user.email=t@t -c user.name=t commit -qm init

opengrep scan --config rule.yaml --force-exclude --exclude='/myvendor/**' .
opengrep scan --config rule.yaml --force-exclude --exclude='deep/myvendor/**' .
```

Findings by file (`--json`, paths relativized):

| `--exclude` | opengrep still scans | `git check-ignore` would skip |
|---|---|---|
| *(none)* | `a/deep/myvendor/lib.py`, `deep/myvendor/lib.py`, `myvendor/lib.py`, `src/main.py` | — |
| `/myvendor/**` | `src/main.py` | `myvendor/lib.py` only |
| `deep/myvendor/**` | `myvendor/lib.py`, `src/main.py` | `deep/myvendor/lib.py` only |
| `**/myvendor/**` | `src/main.py` | all three `myvendor` files |

The gitignore column is real output, taken by writing the same pattern to `.gitignore` and running `git check-ignore --no-index a/deep/myvendor/lib.py deep/myvendor/lib.py myvendor/lib.py src/main.py` (git 2.43.0).

So `/myvendor/**` and `**/myvendor/**` are treated identically by opengrep, while gitignore gives them deliberately different meanings.

**Expected behavior**

Following the gitignore pattern format that `--help` cites:

- `/myvendor/**` — skips `myvendor/lib.py` only.
- `deep/myvendor/**` — skips `deep/myvendor/lib.py` only.
- `**/myvendor/**` — skips all three (this one already behaves correctly).
- A pattern with no slash (`myvendor`, `*.py`) keeps matching at any depth, as today.

If the current semantics are intentional, the `--help` text and docs should stop citing the gitignore pattern format, since there is currently no way to express "this directory, at the root only" — prefixing everything with `**/` is a workaround for the reverse case, not for anchoring.

**Upstream context**

This is inherited from semgrep rather than introduced by opengrep, and it is open there:

- https://github.com/semgrep/semgrep/issues/10478 — "Per-rule exclude patterns should be anchored" (open since Aug 2024): "An `exclude` pattern in a rule behaves as if it was prepended with `**/`. As a result, a pattern like `/a` matches not only at the project root but also subfolders such as `b/a`. This goes against the documentation as well as what most users want." A maintainer notes `--experimental` (osemgrep) is affected too.
- https://github.com/semgrep/semgrep/issues/6835 — "Include/exclude patterns matching on start/end" (open since Jan 2023): the request for anchored matching (apply to `parsers/` but not `database/parsers/`), and it also records the documentation gap — the behaviour is described nowhere, so "I had to read semgrep sources to understand why it behaves this way".

Both issues are framed around per-rule `paths.exclude`, but the CLI flag goes through the same code: `TargetManager.preprocess_path_patterns` rewrites `foo/bar` into `**/foo/bar` and `**/foo/bar/**`, and `filter_excludes` runs the CLI `--exclude` list through it. Filing here because opengrep's own `--help` makes the gitignore claim, and because the semgrep issues have had no movement.

- [x] P1: important to fix or quite annoying

**Environment**

- Official binaries `opengrep_manylinux_x86`, versions **1.22.0** and **1.27.1** (latest release) — identical results on both.
- Linux x86_64 (WSL2). git 2.43.0 as the gitignore reference.
- Reproduces with and without `--force-exclude`, and whether the target is `.` or an absolute path.

**Use case**

We hand user-configured path filters (test/vendor/generated code) to `--exclude` so those files are never scanned, and match the *same* stored patterns with a gitignore library elsewhere in the pipeline to label findings that do come back. When the two matchers disagree about a pattern, the scanner skips files the labeller expects to see — and because no finding is produced, the disagreement is invisible in the output. Anchoring behaviour that matches the documented gitignore format would let both sides read one pattern the same way.

dimitris-m:
TL;DR Bring the ocaml cli to parity and drop python completely. Several fixes applied.

Also closes #627, #584, #132, #797, #138

TODO:
Remove flags like `--allow-local-builds` which are not used? Check cli parameters in general.


corneliuhoffman (approved):
this has gone through several rounds of offline reviews.

maciejpirog (approved):

Issue #841, opened 2026-09-12
OCaml CLI: HTTPS rule downloads fail through an HTTP proxy


intSheep:
**Describe the bug**

HTTPS rule downloads fail through an HTTP proxy on current main (`536ebce45f7f3ac159135228d88a1b63cbfbabac`). With the same proxy, the official v1.30.0 Linux x86_64 binary successfully downloads `p/ocaml`; the main build exits with code 2 and HTTP 400.

The native client sends a plain `GET` to the proxy instead of establishing a `CONNECT` tunnel for the HTTPS destination.

**To reproduce**

This local request logger makes the difference visible without contacting an external server. Save it as `proxy.py` and run `python3 -u proxy.py`:

```python
from http.server import BaseHTTPRequestHandler, HTTPServer

class Proxy(BaseHTTPRequestHandler):
    def do_GET(self):
        print(self.requestline, flush=True)
        self.send_error(400, "Use CONNECT for HTTPS")

    def do_CONNECT(self):
        print(self.requestline, flush=True)
        self.send_error(502, "Capture only; no forwarding")

    def log_message(self, *_):
        pass

HTTPServer(("127.0.0.1", 18888), Proxy).serve_forever()
```

In another terminal:

```sh
echo 'let x = 1' > target.ml
HTTPS_PROXY=http://127.0.0.1:18888 https_proxy=http://127.0.0.1:18888 \
  opengrep scan --disable-version-check \
  -c https://example.invalid/rules.yaml target.ml
```

Main sends:

```text
GET /rules.yaml HTTP/1.1
```

v1.30.0 and `curl --proxy http://127.0.0.1:18888 https://example.invalid/rules.yaml` send:

```text
CONNECT example.invalid:443 HTTP/1.1
```

The logger deliberately rejects both requests; it only records the method. With a real HTTP proxy, v1.30.0 completes the rule download while main fails.

**Expected behavior**

Establish a tunnel to the HTTPS destination, then perform TLS and fetch the rules through it.

In [`Http_helpers.call_client`](https://github.com/opengrep/opengrep/blob/536ebce45f7f3ac159135228d88a1b63cbfbabac/libs/networking/http_helpers/http_helpers.ml#L147), selecting a proxy replaces the connection URI but keeps the original request method. There is no CONNECT step. This appears to be an existing limitation of the native HTTP client exposed by the CLI switch in #827.

**Priority**

- [x] P2: regular bug that should get fixed

**Environment / use case**

[...]

PR #842, opened 2026-09-12
Kotlin: preserve property accessor bodies

### LINKED ISSUE #828: Kotlin: property getter and setter bodies are dropped from the AST
```kotlin
class A {
  val p: String
    get() = intent.getStringExtra(K)!!
  val q: String
    get() { return intent.getStringExtra(K)!! }
  fun f() = intent.getStringExtra(K)!!
}
```

```
opengrep scan -e '$I.getStringExtra($K)' -l kotlin a.kt
```

Only line 6 (the function) matches. Lines 3 and 5 never match.

`opengrep show dump-ast kotlin a.kt` shows each accessor as `OtherDef(("Getter", ()), [])` with an empty payload: the body is discarded at translation, so no pattern can match inside an accessor and taint cannot flow through one.

Checked on 1.29.0.

intSheep:
Kotlin getters and setters were parsed, but their bodies were discarded during translation to the generic AST. Calls inside them could never match search or taint rules.

Keep accessors as function definitions alongside the property, preserving their bodies, parameters, return types and modifiers. Give them property-specific names so intrafile analysis can distinguish them. Regression tests cover inline and multiline accessors, expression and block bodies, paired accessors, and tainted setter parameters.

Fixes #828.

Validation:

- `make core-test`: 6,487 successful tests, including 34 expected failures.
- The issue's example now matches all three calls. The taint fixture reports the same seven expected findings with and without `--taint-intrafile`.

Tests used SOCKS for network access with HTTP proxy variables unset because of #841.


dimitris-m:
This branch is 210 commits behind main, including the interfile merge (#779). Please rebase onto current main, adapt the Kotlin changes, and re-run the Kotlin tests.

This is not a review.

intSheep:
Rebased and force-pushed onto current `main` (`43622ac0b`). The PR is now a single commit (`f3770f296`) with no merge commit.

I also adapted the Kotlin accessor lowering for the interfile changes from #779:
- getter/setter functions inherit property and extension-receiver types, including separate/paired CST accessors;
- hidden synthetic accessor identities are excluded from source-level interfile binding/return indexes while their bodies remain available for analysis;
- added interfile regressions for typed setters, extension/semicolon accessors, and synthetic-name collisions.

Validation completed locally:
- `git diff --check`
- the new interfile fixture YAML parses successfully

I could not run the full OCaml/Kotlin suite locally because this environment does not have `dune`, `opam`, or `ocamlc`. The new `check-linear-history` and `build-test-core-x86` workflow runs are currently `action_required` and have not executed; they appear to be waiting for maintainer approval.

dimitris-m:
there is a couple of issues here, the main one being :

_I could not run the full OCaml/Kotlin suite locally because this environment does not have dune, opam, or ocamlc._ 

When one cannot verify their solution, it is best to simply open an issue.

By the way, this is in direct contradiction to the PR description:

"Validation:

make core-test: 6,487 successful tests, including 34 expected failures."

